PR #3, opened 2019-03-07
Adds first version of `validate_block`

bkchr:
Depends on: https://github.com/paritytech/substrate/pull/1926


runtime/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+// Cumulus is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+#![cfg_attr(not(feature = "std"), no_std)]
```
rphmeier: any docs?

runtime/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+use runtime_primitives::traits::Block as BlockT;
+
+#[cfg(not(feature = "std"))]
+#[doc(hidden)]
+pub use rstd::slice;
+
+#[macro_use]
+pub mod validate_block;
+
+type WitnessData = BTreeMap<Vec<u8>, Vec<u8>>;
```
rphmeier: Surely a hashmap is faster?

runtime/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// The parachain block that is created on a collator and validated by a validator.
+#[derive(Encode, Decode)]
+struct ParachainBlock<B: BlockT> {
```
rphmeier: Let's call this `ParachainBlockData` to be consistent with Polkadot terminology.
The doc would be "block data needed to validate a state transition for a Cumulus parachain. Created by collators, executed by validators". Or something.

runtime/src/lib.rs
bkchr: Yeah can switch to a hashmap, just will require some external hashmap, as std hashmap does not work on `no_std`. AFAIK.

runtime/src/validate_block/storage_functions.rs
```diff
@@ -0,0 +1,79 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! All storage functions that are replaced by `validate_block` in the Substrate runtime.
+
+use crate::WitnessData;
+use rstd::{slice, ptr, cmp};
+
+pub static mut STORAGE: Option<WitnessData> = None;
```
rphmeier: static mut is very frightening. globals should be protected by some kind of generic guard.

rphmeier: the fact that it's public is even scarier. unsafe code should be isolated at module boundaries, where invariants are protected.

runtime/src/validate_block/storage_functions.rs
```diff
@@ -0,0 +1,79 @@
...
+const STORAGE_SET_EXPECT: &str = "`STORAGE` needs to be set before calling this function.";
+
+pub unsafe fn ext_get_allocated_storage(key_data: *const u8, key_len: u32, written_out: *mut u32) -> *mut u8 {
```
rphmeier: The signature of this function is essentially `get_allocated_storage(key) -> &mut [u8]`.

It can be called twice in a row, which would give aliased mutable memory -- we should make sure the same invariants of the substrate-native implementation are preserved.

runtime/src/validate_block/storage_functions.rs
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+}
+
+pub unsafe fn ext_set_storage(key_data: *const u8, key_len: u32, value_data: *const u8, value_len: u32) {
+	let key = slice::from_raw_parts(key_data, key_len as usize);
+	let value = slice::from_raw_parts(value_data, value_len as usize);
+
+	STORAGE.as_mut().map(|s| {
+		s.insert(key.to_vec(), value.to_vec());
```
rphmeier: side-effecting `map` is a bit weird. `if let` is more idiomatic

runtime/src/validate_block/storage_functions.rs
rphmeier: this and all other functions in this module should have clear documentation about invariants and restrictions on arguments -- we generally have a "all public items should have docs" guideline but this is especially true for `unsafe` code.

runtime/src/lib.rs
rphmeier: `hashbrown` and a `no-std` wrapper implementing `Decode` I guess. But let's get to that later

runtime/src/validate_block/implementation.rs
```diff
@@ -0,0 +1,230 @@
...
+	"`STORAGE` needs to be set before calling this function.";
+const STORAGE_ROOT_LEN: usize = 32;
+
+/// Extract the hashing algorithm type from the given block type.
+type HashingOf<B> = <<B as BlockT>::Header as HeaderT>::Hashing;
+/// Extract the hash type from the given block type.
+type HashOf<B> = <B as BlockT>::Hash;
+
+/// Abstract the storage into a trait without `Block` generic.
+trait StorageT {
```
rphmeier: a bit weird to call it `StorageT`. why not just `Storage`?

runtime/src/validate_block/implementation.rs
```diff
@@ -0,0 +1,230 @@
...
+		})
+	}
+}
+
+impl<B: BlockT> StorageT for Storage<B> {
+	fn get(&self, key: &[u8]) -> Option<Vec<u8>> {
+		self.overlay.get(key).cloned().or_else(|| {
+			read_trie_value(
+				&self.witness_data,
+				&self.storage_root, key
```
rphmeier: `key` should go on new line.

runtime/src/validate_block/implementation.rs
```diff
@@ -0,0 +1,230 @@
...
+			ptr
+		},
+		None => {
+			*written_out = u32::max_value();
+			ptr::null_mut()
+		}
+	}
+}
+
+unsafe fn ext_set_storage(
```
rphmeier: i would prefer to extract these to a submodule where we can be a bit more sure about invariants

runtime/src/validate_block/implementation.rs
```diff
@@ -0,0 +1,230 @@
...
+};
+use executive::ExecuteBlock;
+
+use substrate_trie::{MemoryDB, read_trie_value, delta_trie_root};
+
+use rstd::{slice, ptr, cmp, vec::Vec, boxed::Box, mem};
+
+use hash_db::HashDB;
+
+static mut STORAGE: Option<Box<StorageT>> = None;
```
rphmeier: eep. still not a fan of the `static mut`. Why not at least a `RefCell`?

runtime/src/validate_block/implementation.rs
```diff
@@ -0,0 +1,230 @@
...
+	};
+
+	E::execute_block(block);
+}
+
+/// The storage implementation used when validating a block.
+struct Storage<B: BlockT> {
+	witness_data: MemoryDB<<HashingOf<B> as HashT>::Hasher>,
+	overlay: hashbrown::HashMap<Vec<u8>, Option<Vec<u8>>>,
+	storage_root: HashOf<B>,
```
rphmeier: `B::Hash` is easier

runtime/src/validate_block/implementation.rs
```diff
@@ -0,0 +1,230 @@
...
+impl<B: BlockT> Storage<B> {
+	/// Initialize from the given witness data and storage root.
+	///
+	/// Returns an error if given storage root was not found in the witness data.
+	fn new(
+		data: WitnessData,
+		storage_root: HashOf<B>
+	) -> Result<Self, &'static str> {
+		let mut db = MemoryDB::default();
+		data.into_iter().for_each(|i| { db.insert(&[], &i); });
```
rphmeier: what does the `&[]` mean here?

runtime/src/validate_block/implementation.rs
```diff
@@ -0,0 +1,230 @@
...
+	key_data: *const u8,
+	key_len: u32,
+	written_out: *mut u32,
+) -> *mut u8 {
+	let key = slice::from_raw_parts(key_data, key_len as usize);
+	match STORAGE.as_mut().expect(STORAGE_SET_EXPECT).get(key) {
+		Some(mut value) => {
+			*written_out = value.len() as u32;
+			let ptr = value.as_mut_ptr();
+			mem::forget(ptr);
```
rphmeier: what's the purpose of `forget`ting ptr (which is copy) and then using it right after?

runtime/src/validate_block/implementation.rs
rphmeier: TBH I see no reason why, with native-runtime optimization, this is guaranteed not to be used from multiple threads

bkchr: This will never be callable by the native side. First we don't support replacing the functions on native, second `register_validate_block!` resolves to nothing on native.

runtime/src/validate_block/implementation.rs
bkchr: I don't really know. I copied this code from substrate. I think that `&[]` is the prefix. Maybe @cheme can help us here?

runtime/src/validate_block/implementation.rs
bkchr: Because I have a struct that is called `Storage` in the same module :sweat_smile:

runtime/src/validate_block/implementation.rs
bkchr: Hmm, this is already its own module and you are sure about the invariants. This function is only usable by `validate_block`. I would not put this into another file.

runtime/src/validate_block/implementation.rs
cheme: That is the prefix for the MemoryDB, I believe here if MemoryDB is used from substrate trie crate, there is no key transformation with the prefix so it is safe to put nothing (a variant called PrefixedMemoryDB does insert the prefix before the key when inserting/querying the keyvaluedb to avoid collision in trie with unhashed value).

runtime/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+	"primitives/std",
+	"executive/std",
+	"memory-db/std",
+	"hash-db/std",
+	"trie-db/std",
+	"test-client/std",
+	"substrate-trie/std",
+]
+wasm = [
+	"hashbrown/nightly",
```
cheme: is this `wasm` feature something native or something that require some feature declaration?
(I remember that I wanted to use hashbrown but use hasmap_core in trie because it felt easier to manage with 'std' substrate feature, just thinking let's wait for hashbrown to be std and see if at some point it makes it to alloc).

bkchr: This requires a feature declaration.

runtime/src/validate_block/implementation.rs
rphmeier: hm, maybe we need this when using a chain with child tries

cheme: there is the key conflict with child trie, but this way of doing things cannot solve it (for two identical child trie there is the same prefix). I was also wondering about exposing this key transformation trait to insert the prefix with a specialized implementation (https://github.com/paritytech/trie/pull/12 last comment), but using a hashdb specific implementation would achieve the same thing.

bkchr: I assume that we then need to adapt [this](https://github.com/paritytech/substrate/blob/master/core/state-machine/src/proving_backend.rs#L104)? This is basically the functionality I use to generate the proof. I assume we would need to extend the proof with the prefix somehow?

cheme: If you use child trie I think you need https://github.com/paritytech/substrate/pull/2209 (I should split it in two has I believe the api change would not be merged or at least requires some consensus but the proof side of the pr seems ok).
But otherwhise the prefix thing should only apply to avoid reference count in db, so the Recorder of trie does not use the prefix so it should not matter (that is the good thing about the way Arkadiy solves the issue).

runtime/src/validate_block/implementation.rs
bkchr: Was incorrect and is fixed now.

PR #9, opened 2019-04-29
Pass parent hash to the validate_block function

### LINKED ISSUE #6: Change `validate_block` signature
`validate_block` needs to get the last header hash and check that the hash matches the hash in the generated header.

DarkEld3r:
Closes #6.

### DIFF
--- a/runtime/src/validate_block/implementation.rs
+++ b/runtime/src/validate_block/implementation.rs
@@ -23,6 +23,7 @@ use runtime_primitives::traits::{
 use executive::ExecuteBlock;
 
 use substrate_trie::{MemoryDB, read_trie_value, delta_trie_root};
+use codec::Decode;
 
 use rstd::{slice, ptr, cmp, vec::Vec, boxed::Box, mem};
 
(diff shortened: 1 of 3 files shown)


runtime/src/validate_block/implementation.rs
```diff
@@ -56,14 +57,14 @@ trait Storage {
...
+	mut arguments: &[u8],
 ) {
-	use codec::Decode;
-
-	let block_data = crate::ParachainBlockData::<B>::decode(&mut block_data)
+	let (parent_hash, block_data): (B::Hash, crate::ParachainBlockData::<B>) = Decode::decode(&mut arguments)
 		.expect("Could not decode parachain block.");
 	// TODO: Add `PolkadotInherent`.
 	let block = B::new(block_data.header, block_data.extrinsics);
+	assert!(parent_hash == *block.header().parent_hash(), "Invalid parent hash");
```
DarkEld3r: I'm not sure about error reporting here. Should `validate_block` signature be changed?

bkchr: `assert!` is okay here.

runtime/src/validate_block/implementation.rs
```diff
@@ -56,14 +57,14 @@ trait Storage {
 #[cfg(not(feature = "std"))]
 #[doc(hidden)]
 pub fn validate_block<B: BlockT, E: ExecuteBlock<B>>(
-	mut block_data: &[u8],
+	mut arguments: &[u8],
 ) {
-	use codec::Decode;
```
bkchr: Please bring this back and remove from the top of the file.

Did you tried to compile this for wasm and std?

runtime/src/validate_block/tests.rs
```diff
@@ -152,4 +152,25 @@ fn validate_block_with_extrinsics() {
 		witness_data_storage_root
 	);
 	call_validate_block(client.genesis_hash(), block_data).expect("Calls `validate_block`");
-}
\ No newline at end of file
+}
+
+#[test]
+#[should_panic]
```
rphmeier: if it panics internally, you can use `#[should_panic(expected = "error_substring")]`

DarkEld3r: Panic is intercepted internally, so `#[should_panic(expected = "Invalid parent hash")]` will not work.

bkchr: Yep, the panic is thrown in WASM and this just prints it.

PR #11, opened 2019-06-18
Implement Collator

bkchr:
(Just concentrate on the collator crate)

Open questions:
- [ ] What do we do with `max_duration` of `propose`?
- [ ] Do we want to return more detailed errors in `produce_candidate`?

### DIFF
--- a/collator/src/lib.rs
+++ b/collator/src/lib.rs
@@ -1,0 +1,126 @@
+// Copyright 2019 Parity Technologies (UK) Ltd.
+// This file is part of Substrate.
+
+// Substrate is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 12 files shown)


collator/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+						error!("Failed to create inherent data: {:?}", e);
+						InvalidHead
+					})
+			)
+			.and_then(|(proposer, inherent_data)| {
+				proposer.propose(
+					inherent_data,
+					Default::default(),
+					//TODO: Fix this.
+					Duration::from_secs(6),
```
rphmeier: I'm not sure what to do here, but it is worth noting that `polkadot-collator` times out collation futures that take longer than a given duration. Maybe we can pass that timeout along to `ProduceCandidate` somehow.

test/client/src/lib.rs
```diff
@@ -16,36 +16,108 @@
...
-	#![allow(missing_docs)]
-	use substrate_test_client::runtime;
-	use substrate_test_client::executor::native_executor_instance;
+	use test_client::executor::native_executor_instance;
 	native_executor_instance!(
 		pub LocalExecutor,
 		runtime::api::dispatch,
 		runtime::native_version,
-		include_bytes!("../../runtime/wasm/target/wasm32-unknown-unknown/release/cumulus_test_runtime.compact.wasm")
+		runtime::WASM_BINARY
```
rphmeier:
```suggestion
		runtime::WASM_BINARY,
```

PR #12, opened 2019-06-21
Switch to `parachain::ValidationParams`

bkchr:


runtime/Cargo.toml
```diff
@@ -5,21 +5,26 @@ authors = ["Parity Technologies <admin@parity.io>"]
...
 runtime-primitives = { package = "sr-primitives", git = "https://github.com/paritytech/substrate", default-features = false, branch = "bkchr-cumulus-branch" }
 primitives = { package = "substrate-primitives", git = "https://github.com/paritytech/substrate", default-features = false, branch = "bkchr-cumulus-branch" }
 rio = { package = "sr-io", git = "https://github.com/paritytech/substrate", default-features = false, branch = "bkchr-cumulus-branch" }
 executive = { package = "srml-executive", git = "https://github.com/paritytech/substrate", default-features = false, branch = "bkchr-cumulus-branch" }
 substrate-trie = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "bkchr-cumulus-branch" }
+
+# Polkadot dependencies
+parachain = { package = "polkadot-parachain", git = "https://github.com/paritytech/polkadot", default-features = false, branch = "bkchr-cumulus-branch" }
+
+# Other deps
```
DarkEld3r: Will it make sense to split "other dependencies" into "(other) parity dependencies" (like `codec`, `memory-db`, etc.) and "third-party dependencies" (only `hashbrown` for now)?

bkchr: No, not really. That would be to fine-grained.

PR #14, opened 2019-09-25
a simple CI for the start

TriplEight:
runs all tests frozen and collects warnings


TriplEight:
@bkchr Can be merged. Seems like Cumuls doesn't compile now.

PR #15, opened 2019-10-08
Provide `SelectChain` implementation for parachains

bkchr:


consensus/src/lib.rs
```diff
@@ -116,38 +130,112 @@ impl<B, E, Block, RA> LocalClient for Client<B, E, Block, RA> where
...
 
 		Ok(Box::new(s))
 	}
+
+	fn parachain_head_at(
+		&self,
+		at: &BlockId<PBlock>,
+		para_id: ParaId,
+	) -> ClientResult<Option<Vec<u8>>> {
+		self.runtime_api().parachain_status(at, para_id).map(|s| s.map(|s| s.head_data.0))
```
rphmeier: IMO querying storage is much more light-client friendly.
Alternatively we could scrape the block body.

rphmeier:
justification for migrating to runtime API instead of raw storage query?

bkchr:
I tried to use raw storage querys, but it seems that the parachain heads are not stored anymore under `parachains heads`. I could not find any relevant code in Polkadot.

rphmeier:
Fair enough - can revisit later.

PR #16, opened 2019-10-11
ICMP runtime module

### LINKED ISSUE #5: Cumulus Runtime
We will want a runtime component for parachains to handle incoming messages.

The module will look something like this.
```rust
trait OnMessages {
    fn on_messages(messages: Vec<polkadot::IncomingMessage>);
}

trait Trait {
    type OnMessages: OnMessages;
}

struct<T: Trait> Module<T> {
    // defines an inherent for the incoming messages to be passed. the incoming messages are passed to `T::OnMessages::on_messages`.
}
```

gavofyork:
Closes #5 

This codifies the API for handling incoming ICMP messages. It also includes a basic `Call`-dispatch system if you know that incoming messages will take the chain's `Call` format (as used in transactions).

No queuing, queue-management, error reporting, or integration with the weight system is done. We may want to consider each of these in due course.

TODO:
- [ ] Tests
- [ ] Channel saturation reporting via an inherent (to figure out what channels are available and how saturated they are)
- [ ] Outgoing message API
- [ ] Outgoing message queuing (to handle the case where messages cannot be sent due to channel saturation)
- [ ] Runtime API for creating inherents
- [ ] Weight system integration (to ensure that overly weighty blocks are not needed to be processed)
- [ ] Queuing and queue management (to manage the case where the incoming message queue is too great)

Elsewhere in the codebase, we need channels:
- [ ] Channel-saturation storage and upkeep (relay-chain runtime)
- [ ] Channel creation/removal/manipulation (relay-chain runtime)
- [ ] Channel-saturation inherent creation (cumulus scraping relay-chain state and feeding into parachain runtime) 
- [ ] Switch from ingress to egress queuing (relay-chain runtime)
- [ ] Channel-reapability provision into `set_heads` (cumulus scraping relay-chain state and feeding into set_heads)
- [ ] Egress queue reaping in `set_heads` (relay-chain runtime) 
- [ ] Incorrect reap-data in `set_heads` punishment (relay-chain runtime, plus client-based fisherman component).


JoshOrndorff:
This is superceded by #80 right? (If not, please re-open)

PR #17, opened 2019-10-18
implementation of BlockAnnounceValidator

thiolliere:
implementation of BlockAnnounceValidator that checks for justifications in data

## TODO:

* [ ] tests

### DIFF
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -1,0 +1,82 @@
+use substrate_client::error::{Error as ClientError};
+use sr_primitives::traits::{Block as BlockT};
+use substrate_consensus_common::block_validation::{Validation, BlockAnnounceValidator};
+
+use polkadot_primitives::{
+	Hash as PHash, parachain::{CandidateReceipt, ValidatorIndex, ValidatorSignature, ValidatorId},
+};
+use polkadot_statement_table::Statement;
+use polkadot_validation::check_statement;
+
+use codec::{Decode, Encode};
+
+use std::marker::PhantomData;
+
+/// Justification that a parachain block is the parachain block candidate of one of the relay chain
+/// validator.
+#[derive(Encode, Decode)]
+pub struct BlockCandidateJustification {
+	/// Receipt of the parachain block candidate of the signer.
+	candidate_receipt: CandidateReceipt,
+	/// Signer of `signature`.
+	signer: ValidatorIndex,
+	/// Signature of the Candidate statement with `candidate_receipt`.
+	signature: ValidatorSignature,
+	/// The parent block of which the candidate must be include.
+	relay_chain_parent_hash: PHash,
+}
+
+/// Validate that data is a valid justification form a relay-chain validator that the block is a
+/// valid parachain-block candidate.
+pub struct JustifiedBlockAnnounceValidator<B> {
+	authorities: Vec<ValidatorId>,
+	phantom: PhantomData<B>,
+}
+
+impl<B: BlockT> JustifiedBlockAnnounceValidator<B> {
+	pub fn new(authorities: Vec<ValidatorId>) -> Self {
+		Self {
+			authorities,
+			phantom: Default::default(),
+		}
+	}
+}
+
+impl<B: BlockT> BlockAnnounceValidator<B> for JustifiedBlockAnnounceValidator<B> {
+	fn validate(&mut self, header: &B::Header, mut data: &[u8])
+		-> Result<Validation, Box<dyn std::error::Error + Send>>
+	{
+		let justification = BlockCandidateJustification::decode(&mut data)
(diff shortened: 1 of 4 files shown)


consensus/src/lib.rs
```diff
@@ -239,3 +248,72 @@ impl<Block, PC, SC> SelectChainT<Block> for SelectChain<Block, PC, SC> where
...
+			phantom: Default::default(),
+		}
+	}
+}
+
+impl<B: BlockT> BlockAnnounceValidator<B> for JustifiedBlockAnnounceValidator<B> {
+	fn validate(&mut self, header: &B::Header, mut data: &[u8])
+		-> Result<Validation, Box<dyn std::error::Error + Send>>
+	{
+		let justification = BlockCandidateJustification::decode(&mut data)
```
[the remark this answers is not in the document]
thiolliere: in which situation ? to validate our own block candidate ?

consensus/src/lib.rs
```diff
@@ -239,3 +248,72 @@ impl<Block, PC, SC> SelectChainT<Block> for SelectChain<Block, PC, SC> where
 		}
 	}
 }
+
+#[derive(Encode, Decode)]
+/// Justification that a parachain block is the parachain block candidate of one of the relay chain
+/// validator.
+pub struct BlockCandidateJustification {
```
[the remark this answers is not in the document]
thiolliere: do you expect to have only this struct in network yet or also logic to validate a block candidate ?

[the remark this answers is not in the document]
thiolliere: all moved

network/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+			phantom: Default::default(),
+		}
+	}
+}
+
+impl<B: BlockT> BlockAnnounceValidator<B> for JustifiedBlockAnnounceValidator<B> {
+	fn validate(&mut self, header: &B::Header, mut data: &[u8])
+		-> Result<Validation, Box<dyn std::error::Error + Send>>
+	{
+		let justification = BlockCandidateJustification::decode(&mut data)
```
bkchr: This still fails when we don't have any justification.

network/src/lib.rs
```diff
@@ -0,0 +1,102 @@
...
+use polkadot_validation::check_statement;
+
+use codec::{Decode, Encode};
+
+use std::marker::PhantomData;
+
+/// Justification that a parachain block is the parachain block candidate of one of the relay chain
+/// validator.
+#[derive(Encode, Decode)]
+pub struct BlockCandidateJustification {
```
bkchr: Why not use the polkadot candidate message directly?

Do you want to save some spaces? If yes, could you please provide a `From` implementation.

network/src/lib.rs
thiolliere: if there is no justification why is block valid ? in which situation does this happens, if it happens when you import your own block then we should check that it is indeed your block so you can trust it.

Is there another point in the code that check that the justification exist and I'm not aware of ? because if none I don't understand why this is valid to validate a block without justification.

network/src/lib.rs
thiolliere: or maybe you mean using a field with Statement instead of CandidateReceipt. in this case ok yes I'll do it, It is better indeed.

network/src/lib.rs
bkchr: We will not persist the justifications, so when you are doing an initial sync, you will get the blocks without any justification.

network/src/lib.rs
bkchr: I just thought that the whole justification should just be the statement, as the statement should contain all of these information?

thiolliere: `polkadot-network::GossipStatement` contains all informations. we can make this being the justification.

thiolliere: I mean we can reuse this entirely and we just need to check header match as well.

https://github.com/paritytech/polkadot/blob/master/network/src/gossip.rs#L426-L455

But I feel like polkadot could change a bit and require more than this to check the message, and this would be unfortunate.

I can make this polkadot stuff more public and reuse it here. I thought it is better to not rely on the whole stuff as polkadot could change MessageValidationData and if it happens cumulus can't have some of them then we are stuck back to this current state.

network/src/lib.rs
thiolliere: I'll have to understand where unjustified new block import will be refused but at least this is done now

network/src/lib.rs
bkchr: But polkadot can also not change these structs as will, as they are specified as well.

bkchr:
Do you still want to add tests?

thiolliere:
OK I was very slow on this one, test can be added in a second PR as there is currently very few unit test in cumulus anyway.

bkchr:
Okay

PR #33, opened 2020-01-14
Make everything compile and fix tests

bkchr:


test/parachain/src/service.rs
```diff
@@ -49,43 +48,53 @@ construct_simple_protocol! {
...
+				client,
+				inherent_data_providers.clone(),
+			)?;
+
+			Ok(import_queue)
+		})?;
 
 		(builder, inherent_data_providers)
-	}}
+		}};
```
cecton: This is not properly indented

Issue #24, opened 2019-11-29
Collator node thrown incorrect warnings on startup


JoshOrndorff:
When starting the collator node, it reliably prints errors like

```
Unable to bind server to 127.0.0.1:9955. Trying random port.
Not a valid bootnode address: /ip4/127.0.0.1/tcp/30333/QmbQMG8VWKumRyXg47UixQiosUMP2nWRsmWRojqDqJTG9t
```

These messages do not correspond to actual underlying errors. UIs still connect to the websocket port correctly, and the collator still peers with the bootnode at the supposedly invalid address. Possibly related to #19

bkchr:
As written in #34 we should provide some sensible defaults for the relay chain node.

cecton:
This error still exist:
```
Error on libp2p listener ListenerId(1): Address already in use (os error 98)
```

zhangtianai:
```
2020-09-04 22:43:43 📪 Libp2p listener () closed: Address already in use (os error 48)
2020-09-04 22:43:43 📪 Libp2p listener () closed: Address already in use (os error 48)
```

PR #42, opened 2020-01-16
Splitting relaychain and parachain command-line arguments

### LINKED ISSUE #34: Pass CLI arguments to parachain/relay chain node
When running the collator, I want to pass CLI arguments that are related to the parachain and to the relay chain. As both are substrate projects, we need to make it possible to pass the same arguments twice to the same binary.

We want the following syntax:

`collator --name parachain -- --name relaychain`

The first arguments before the `--` are the parachain parameters and everything after `--` is a relay chain parameter.

`--help` should make it clear how we expect parameters to be passed.

Maybe we could also include some reasonable defaults for ports, so that they don't clash: https://github.com/paritytech/cumulus/issues/24

Like putting the relay chain ports at `default + 1` (if they are not overwritten by a cli arg)

cecton:
Fixes #34

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+[*]
+indent_style=tab
+indent_size=tab
+tab_width=4
+end_of_line=lf
+charset=utf-8
+trim_trailing_whitespace=true
+max_line_length=100
+insert_final_newline=true
+
+[*.yml]
+indent_style=space
+indent_size=2
+tab_width=8
+end_of_line=lf
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,9 +8,3 @@ charset=utf-8
 trim_trailing_whitespace=true
 max_line_length=100
 insert_final_newline=true
-
-[*.yml]
-indent_style=space
-indent_size=2
-tab_width=8
-end_of_line=lf
--- a/test/parachain/src/cli.rs
+++ b/test/parachain/src/cli.rs
@@ -64,7 +64,7 @@ struct ExportGenesisStateCommand {
 }
 
 /// Parse command line arguments into service configuration.
-pub fn run<I, T, E>(args: I, exit: E, version: VersionInfo) -> error::Result<()>
+pub fn run<I, T, E>(args_parachain: I, args_relaychain: I, exit: E, version: VersionInfo) -> error::Result<()>
 where
 	I: IntoIterator<Item = T>,
 	T: Into<std::ffi::OsString> + Clone,
@@ -74,7 +74,7 @@ where
 	match parse_and_prepare::<SubCommands, NoCustom, _>(
 		&version,
 		"cumulus-test-parachain-collator",
-		args,
+		args_parachain,
 	) {
 		ParseAndPrepare::Run(cmd) => cmd.run(
 			load_spec,
--- a/test/parachain/src/cli.rs
+++ b/test/parachain/src/cli.rs
@@ -71,6 +71,13 @@ where
 	E: IntoExit + Send + 'static,
 {
 	type Config<T> = Configuration<(), T>;
+
+	let config_relaychain = parse_and_prepare::<NoCustom, NoCustom, _>(
+		&version,
+		"cumulus-test-parachain-collator",
+		args_relaychain,
+	).into_configuration::<(), _, _, _>(load_spec);
+
 	match parse_and_prepare::<SubCommands, NoCustom, _>(
 		&version,
 		"cumulus-test-parachain-collator",
--- a/test/parachain/src/cli.rs
+++ b/test/parachain/src/cli.rs
@@ -97,7 +97,6 @@ where
 				// TODO
 				config.network.listen_addresses = Vec::new();
 
-				// TODO
 				let mut polkadot_config = parse_and_prepare::<NoCustom, NoCustom, _>(
 					&version,
 					"cumulus-test-parachain-collator",
--- a/test/parachain/src/cli.rs
+++ b/test/parachain/src/cli.rs
@@ -105,8 +105,11 @@ where
 					load_spec_polkadot,
 					config.in_chain_config_dir("polkadot"),
 				)
-					.map_err(|e| e.to_string())?
-					.expect("not a run command?");
+				.map_err(|e| e.to_string())?
+				.expect(
+					"can only fail when this is a CustomCommand. Running parse_and_prepare with \
+					NoCustom can never return a CustomCommand; therefore this will never fail; qed"
+				);
 				polkadot_config.network.boot_nodes = config.network.boot_nodes.clone();
 
 				if let Some(ref config_dir) = polkadot_config.config_dir {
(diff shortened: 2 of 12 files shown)


test/parachain/src/cli.rs
```diff
@@ -95,22 +98,18 @@ where
...
 
 				// TODO
-				let mut polkadot_config =
-					polkadot_collator::Configuration::default_with_spec_and_base_path(
-						polkadot_service::ChainSpec::from_json_bytes(
-							&include_bytes!("../res/polkadot_chainspec.json")[..],
-						)?,
-						config.in_chain_config_dir("polkadot"),
-					);
+				let mut polkadot_config = parse_and_prepare::<NoCustom, NoCustom, _>(
```
rphmeier: can the TODO above be removed? it's a bit weird to have a TODO where it's unspecified what should be done

test/parachain/src/cli.rs
```diff
@@ -95,22 +98,18 @@ where
...
+					&version,
+					"cumulus-test-parachain-collator",
+					args_relaychain,
+				).into_configuration::<CustomConfigurationPolkadot, _, _, _>(
+					load_spec_polkadot,
+					config.in_chain_config_dir("polkadot"),
+				)
+					.map_err(|e| e.to_string())?
+					.expect("not a run command?");
```
rphmeier: in Substrate and Polkadot codebases, we have a stricter policy on panickers than that - please provide a logical proof (premise -> conclusion) that this does not fail.

test/parachain/src/main.rs
```diff
@@ -30,17 +31,28 @@ pub use sc_cli::{error, IntoExit, VersionInfo};
 
 /// The parachain id of this parachain.
 pub const PARA_ID: ParaId = ParaId::new(100);
+const EXECUTABLE_NAME: &'static str = "cumulus-test-parachain-collator";
+const DESCRIPTION: &'static str =
+	"Cumulus test parachain collator\n\nThe command-line arguments provided first will be \
+	passed to the parachain node, while the arguments provided after -- will be passed \
+	to the relaychain node.\n\n\
+	cumulus-test-parachain-collator --parachain-args -- --relaychain-args";
```
rphmeier:
```suggestion
	cumulus-test-parachain-collator [parachain-args] -- [relaychain-args]";
```

test/parachain/src/cli.rs
cecton: I don't think I can proof anything here. The user can send other commands than RunCmd. But I think we expect only RunCmd here, right? @bkchr

rphmeier: If you can't prove anything, then you can get rid of the `expect` and exit gracefully.

bkchr: Yeah, just do `ok_or_else()?`

test/parachain/src/cli.rs
```diff
@@ -94,23 +97,18 @@ where
...
-					);
+				let mut polkadot_config = parse_and_prepare::<NoCustom, NoCustom, _>(
+					&version,
+					"cumulus-test-parachain-collator",
+					args_relaychain,
+				).into_configuration::<CustomConfigurationPolkadot, _, _, _>(
+					load_spec_polkadot,
+					config.in_chain_config_dir("polkadot"),
+				)
+					.map_err(|e| e.to_string())?
```
bkchr: Please remove on level of indentation here and below.

test/parachain/src/main.rs
```diff
@@ -30,17 +31,28 @@ pub use sc_cli::{error, IntoExit, VersionInfo};
 
 /// The parachain id of this parachain.
 pub const PARA_ID: ParaId = ParaId::new(100);
+const EXECUTABLE_NAME: &'static str = "cumulus-test-parachain-collator";
+const DESCRIPTION: &'static str =
```
bkchr: Please create an issue for this. We need to improve this.

test/parachain/src/cli.rs
cecton: I just checked because I changed the code and I forgot exactly what was the issue. But I left that "?" expressly for someone to check that at the review.

This is an unwrap on an Option. This option is always Some unless the command was a custom command. In the current context, it is impossible that it is a custom command because the `parse_and_prepare` is ran with NoCustom. So this won't happen.

Nevertheless... the subcommand can be anything: it can be a RunCmd (what we expect) but also an import-blocks, export-blocks and everything else that is supposed. This makes no sense and I'm afraid at the comment there isn't much I can't do to prevent this.

rphmeier: > This is an unwrap on an Option. This option is always Some unless the command was a custom command. In the current context, it is impossible that it is a custom command because the parse_and_prepare is ran with NoCustom. So this won't happen.

The point is to put this logic _in the code_ so that if it triggers later we can see exactly what assumption was broken. I am not saying to never use `expect`, only to use it with our guidelines.

Here is an example :

`expect("can only fail when this is not a RunCmd. Running parse_and_prepare with NoCustom can only return a RunCmd; therefore this is a RunCmd; qed")`

cecton: @rphmeier I adapted the message consequently but I think you misunderstood what I said.

Whatever I do, the user CAN use import-blocks, export-blocks and the other subcommands in the relay chain arguments. There is nothing I can do to prevent that unless I do quite big modifications on substrate's cli.

I believe the current behavior is acceptable for now but I plan to refactor substrate's cli and I will fix this in a near future.

(The current behavior is: whatever subcommand you use in the relay chain arguments, it will run the collator)

.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+indent_style=tab
+indent_size=tab
+tab_width=4
+end_of_line=lf
+charset=utf-8
+trim_trailing_whitespace=true
+max_line_length=100
+insert_final_newline=true
+
+[*.yml]
```
bkchr: We don't have yamls in here :D

test/parachain/src/cli.rs
rphmeier: > Whatever I do, the user CAN use import-blocks, export-blocks and the other subcommands in the relay chain arguments. There is nothing I can do to prevent that unless I do quite big modifications on substrate's cli.

I followed you this far in the previous comment, and I agree that it would be a large refactoring. Can you clarify: the `expect` call you added cannot be triggered by user input?

.editorconfig
cecton: I shamefully copied the file from substrate :grin:

test/parachain/src/cli.rs
cecton: No it's impossible. `into_configuration` returns None only in case of a custom command but parse_and_prepare (the second one, not the first one) is ran without custom command (NoCustom)

cecton: I have already started the refactoring and it will makes some things (like this) much easier: https://github.com/paritytech/substrate/pull/4692

cecton:
@bkchr CI is green. It's ready for merging but it's using my own variation of the branches where I cherry-picked the `into_configuration`. It also doesn't solve #24

PR #53, opened 2020-02-04
Update CLI

### LINKED ISSUE #24: Collator node thrown incorrect warnings on startup

cecton:
Fixes #24

Default ports for polkadot chain.

### DIFF
--- a/test/parachain/src/service.rs
+++ b/test/parachain/src/service.rs
@@ -89,7 +89,7 @@ macro_rules! new_full_start {
 ///
 /// This function blocks until done.
 pub fn run_collator<E: sc_service::ChainSpecExtension>(
-	parachain_config: sc_service::Configuration<GenesisConfig, E>,
+	mut parachain_config: sc_service::Configuration<GenesisConfig, E>,
 	key: Arc<CollatorPair>,
 	polkadot_config: polkadot_collator::Configuration,
 ) -> sc_cli::error::Result<()> {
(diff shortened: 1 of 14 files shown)


test/parachain/tests/running_the_node_and_interrupt.rs
```diff
@@ -0,0 +1,58 @@
...
+		}
+		eprintln!("Took to long to exit. Killing...");
+		let _ = child.kill();
+		child.wait().unwrap();
+
+		None
+	}
+
+	fn run_command_and_kill(signal: Signal) {
+		let mut cmd = Command::new(cargo_bin("cumulus-test-parachain-collator")).spawn().unwrap();
```
bkchr:
```suggestion
		let mut cmd = Command::new(cargo_bin("cumulus-test-parachain-collator")).args(&["--dev", "-d", "interrupt_test"]).spawn().unwrap();
```

bkchr: Please make sure that you delete the `interrupt_test` dir before running the command.

test/parachain/src/service.rs
```diff
@@ -76,40 +82,87 @@ macro_rules! new_full_start {
...
-	cumulus_collator::run_collator(
-		setup_parachain,
-		crate::PARA_ID,
-		key,
-		polkadot_config,
-	)
+) -> sc_cli::error::Result<()> {
+	let para_id = crate::PARA_ID;
+
+	if polkadot_config.expect_chain_spec().is_kusama() {
```
bkchr: No, this check should not be here. This needs to be done in Polkadot!

cecton: I'm working on this but it is complicated because the services can be full or light and the signature is different, etc...

MEANWHILE: this was intended :grin: I know you don't like it but the code is also on polkadot and doesn't change. The only reason I duplicate it here is to be able to clone the `task_executor` upon launch.

bkchr: I know that we have the same code in Polkadot, but that is already some generic nightmare there. Downstream should not be required to do the same :)

cecton: I made the issue #56 to not lose track of this

Issue #59, opened 2020-02-11
Runtime upgrade


bkchr:
Substrate supports to upgrade the runtime from within the Blockchain. We need to implement some extended functionality to make the upgrade fit into the Polkadot model of upgrading Parachain runtimes, described [here](https://github.com/paritytech/polkadot/issues/825).

- `validate_block` needs to check that `:code` is only altered when allowed (see the Polkadot issue for the requirements).
- Provide a pallet in Cumulus that makes the process of registering an upgrade/checking if an upgrade is doable easy to use. 
- It would probably be nice to augment `set_code` etc to fail on applying, if the requirements are not allowed (instead of failing in `validate_block`). Users can for example send a proposal to upgrade the runtime, this should fail to dispatch if it does not follow the correct path.

PR #70, opened 2020-03-06
runtime upgrade

coriolinus:
Closes https://github.com/paritytech/cumulus/issues/59.


pallet-parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,557 @@
...
+		// this storage value is set by cumulus during block validation
+		storage::unhashed::get(VALIDATION_FUNCTION_PARAMS)
+			// this storage value is set each block by the ProvideInherent
+			// implementation, ensuring that the data is also available
+			// during block production
+			.unwrap_or_else(|| VFPs::get())
+			// in the event that something has gone seriously wrong and neither
+			// of those values are present, I believe that VFPs::get falls back
+			// to the Default::default() value, which should be harmless in that
+			// it prevents users from doing any upgrades during the invalid block.
```
coriolinus, line 104: Assumption about what happens in the event that the VFPs are not initialized for whatever reason. The most likely causes of this error are that this pallet is used outside a cumulus parachain, or is not included in the runtime's inherent providers.

Note that this doesn't panic. An alternate approach here might be to do `.or_else(|| VFPs::try_get().ok()).expect("VFPs were not set; is the runtime configured properly?")`, which would panic in that case. I decided not to, on the theory that panics are bad and should be avoided, but given that this is caused by misconfiguration, it might be desirable to do that instead.

pallet-parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,557 @@
...
+			.and_then(|r| r.ok_or_else(|| "Validation Function Params inherent data not found".into()))
+			.expect("Gets and decodes vfp inherent data");
+
+		VFPs::set(vfp);
+
+		// not sure what precisely the semantics of a None return here are, as
+		// it's not documented, but if I wrote a function like this, then I'd
+		// interpret a None result as "please don't actually create an inherent
+		// for me this block"
+		None
```
coriolinus, line 159: Assumption about `create_inherent` semantics for which I could not find documentation.

runtime/src/lib.rs
```diff
@@ -74,3 +74,50 @@ impl<B: BlockT> ParachainBlockData<B> {
 		&self.extrinsics
 	}
 }
+
+
+pub mod validation_function_params {
```
coriolinus, line 79: Locating this module directly under `runtime` is kind of an accident; I was just looking for a place where its imports would require the fewest new dependencies. if there's a better place to put it, I'm happy to move it.

pallet-parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,559 @@
...
+
+// The pallet's dispatchable functions.
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		// Initializing events
+		// this is needed only if you are using events in your pallet
+		fn deposit_event() = default;
+
+		fn on_initialize(_n: T::BlockNumber) {
+			if let Ok((apply_block, validation_function)) = PendingValidationFunction::try_get() {
```
bkchr, line 49: Please take a look at the `frame-scheduler` for doing this stuff.

pallet-parachain-upgrade/Cargo.toml
```diff
@@ -0,0 +1,108 @@
...
+package = 'parity-scale-codec'
+version = '1.0.0'
+
+[dependencies.cumulus-runtime]
+default-features = false
+path = "../runtime"
+
+[dependencies.sp-core]
+default-features = false
+git = 'https://github.com/paritytech/substrate.git'
```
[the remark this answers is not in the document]
JoshOrndorff: This is an artifact of starting from the pallet template. Substrate release node template script always rewrites the dependencies like this.

pallet-parachain-upgrade/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+This template pallet does not have any genesis configuration.
+
+## Reference Docs
+
+You can view the reference docs for this pallet by running:
+
+```
+cargo doc --open
+```
```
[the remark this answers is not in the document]
JoshOrndorff: Also an artifact from the pallet template

pallet-parachain-upgrade/Cargo.toml
```diff
@@ -0,0 +1,108 @@
...
+std = [
+    'serde',
+    'codec/std',
+    'frame-support/std',
+    'pallet-balances/std',
+	'cumulus-runtime/std',
+    'sp-core/std',
+    'sp-runtime/std',
+    'sp-io/std',
+    'system/std',
```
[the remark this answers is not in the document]
cecton: According to the .editorconfig it should be tabs and the existing toml files are using tabs

pallet-parachain-upgrade/README.md
```diff
@@ -0,0 +1,69 @@
+# parachain-upgrade-pallet
```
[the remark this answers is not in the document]
cecton: Crate-level documentation are on the top of the file:
```
//! some important doc here about the whole crate
//! blah blah

mod good_module
...
```

pallet-parachain-upgrade/Cargo.toml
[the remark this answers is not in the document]
bkchr: Fuck, I meant tabs xD

pallet-parachain-upgrade/Cargo.toml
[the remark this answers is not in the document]
bkchr: Yeah I know^^

runtime/src/lib.rs
coriolinus: Good idea; done in https://github.com/paritytech/cumulus/pull/70/commits/91f12d9ecf055a02099f9f8ca2b457355a6f5c06.

pallet-parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,559 @@
...
+	type Call = Call<T>;
+	type Error = ProvideInherentError;
+	const INHERENT_IDENTIFIER: InherentIdentifier = INHERENT_IDENTIFIER;
+
+	fn create_inherent(data: &InherentData) -> Option<Self::Call> {
+		let vfp = data.get_data(&INHERENT_IDENTIFIER)
+			.and_then(|r| r.ok_or_else(|| "Validation Function Params inherent data not found".into()))
+			.expect("Gets and decodes vfp inherent data");
+
+		VFPs::set(vfp);
```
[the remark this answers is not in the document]
coriolinus: OK: 2e9fa3c

pallet-parachain-upgrade/src/lib.rs
coriolinus: I see it, and it's cool, but it's not obvious to me what action you intend me to take from this suggestion. Do you mean that I should build on that pallet for implementing this functionality? That kind of feels like overkill, to me.

bkchr: Why is that an overkill?

bkchr: This pallet exists exactly for this use case.

coriolinus: - Storage: implementing the delayed update ourselves requires storing a tuple of `(u32, Vec<u8>)`. At the minimum, storing it in the scheduler requires storing a map of `u32 => Option<Vec<dyn Fn>>`, which likely takes more storage space.
- Performance: on every block, the schedule pallet has to load the appropriate vector from the storage map and iterate through it. I believe that this is slower than just checking for existence, which is the only check I make when doing it myself. 
- Implementation. The code which exists in this pallet has some tests and appears to work. Rewriting it in terms of the scheduler would take additional time, digging through resources, etc. While I'm willing to bet that there are examples to find of other pallets using the schedule pallet, working from example is slower than working from documentation, which is scarce.

I might be wrong about any or all of those points; I'd be glad to learn so, if that's the case. However, there's one point more important than those: the schedule pallet operates in terms of the chain's native block number: the parachain block number. That doesn't work for our use case: we check the planned upgrade against the relay chain block number, and ignore the parachain block number. Looking at its code, I can't see a way to make the schedule pallet do what we want in that respect without significant modification.

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,588 @@
...
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		// Initializing events
+		// this is needed only if you are using events in your pallet
+		fn deposit_event() = default;
+
+		fn on_initialize(_n: T::BlockNumber) -> Weight {
+			if let Ok((apply_block, validation_function)) = PendingValidationFunction::try_get() {
+				let relay_chain_height = Self::validation_function_params().relay_chain_height;
+				if relay_chain_height >= apply_block {
```
bkchr, line 66: @rphmeier When we later have delayed `PoVBlock` inclusion, the `ValidationParams` given to `validate_block` will always be build based on the block, the `PoVBlock` was build for? Everything else wouldn't make any sense?

primitives/src/lib.rs
```diff
@@ -51,3 +51,50 @@ pub trait UpwardMessageSender {
...
+
+	/// Current validation function parameters.
+	pub const VALIDATION_FUNCTION_PARAMS: &'static [u8] = b":validation_function_params";
+
+	/// Code upgarde (set as appropriate by a pallet).
+	pub const NEW_VALIDATION_CODE: &'static [u8] = b":new_validation_code";
+
+	/// Validation Function Parameters
+	///
+	/// This struct is the subset of [`ValidationParams`](../../polkadot_parachain/primitives/struct.ValidationParams.html)
```
bkchr, line 69:
```suggestion
	/// This struct is the subset of [`ValidationParams`](polkadot_parachain::ValidationParams)
```

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,588 @@
...
+// The pallet's dispatchable functions.
+decl_module! {
+	pub struct Module<T: Trait> for enum Call where origin: T::Origin {
+		// Initializing events
+		// this is needed only if you are using events in your pallet
+		fn deposit_event() = default;
+
+		fn on_initialize(_n: T::BlockNumber) -> Weight {
+			if let Ok((apply_block, validation_function)) = PendingValidationFunction::try_get() {
+				let relay_chain_height = Self::validation_function_params().relay_chain_height;
```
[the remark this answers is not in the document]
coriolinus: If there's a simple solution to this, I don't know what it is. 

The most straightforward solution I can think of right now is to add a new hook `on_inherents_complete` to `decl_module`; it would run after inherents are complete but before any non-inherents are processed. I'm pretty sure I can figure out how to do that, but it would be a substantial change to Substrate, and would mean that this PR can definitely not be released until after Substrate 2.0. Still, we could then just get rid of `on_initialize` and use `on_inherents_complete` instead.

Does that make sense, or is there a better idea that I'm not seeing?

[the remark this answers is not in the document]
rphmeier: Actually, Substrate is weird about this and inherents can be interleaved with regular transactions as I understand it. Maybe this data could instead go in a `PreRuntime` digest (see BABE module), although it would need to go in the header then, which could be wasteful for light clients down the road.

parachain-upgrade/src/lib.rs
rphmeier: The `ValidationParams` should always correspond to the state of the `relay_parent` in the `AbridgedCandidateReceipt`, so when we have what I've been calling "contextual execution", where we can include an `AbridgedCandidateReceipt` from the context of a earlier `relay_parent`, the `ValidationParams` will still match that `relay_parent`.

If that's what you mean by delayed `PoVBlock` inclusion, then yes.

bkchr: Yep:) "contextual execution" sounds much better ;)

parachain-upgrade/src/lib.rs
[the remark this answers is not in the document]
coriolinus: Would it be both possible and desirable to force all inherents to be executed before any regular transactions? I'm trying to weigh the technical benefits of my approach vs yours.

Also, I'm going to look into the weight of the `ValidationFunctionParameters`; my instinct is that it comes to less than 128 bytes, which feels negligible, but having a measurement allows us to measure precisely what impact it will have for light clients.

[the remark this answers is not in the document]
coriolinus: In discussion, @bkchr asserted that all inherents are already executed before any transactions, which simplifies things a bit.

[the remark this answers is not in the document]
coriolinus: For what it's worth, `ValidationFunctionParams` occupies 102 bits, assuming an 8-bit discriminant for an `Option`, not counting alignment. I don't have the knowledge to say whether that's heavy compared to the rest of the `PreRuntime` digests.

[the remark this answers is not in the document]
rphmeier: > all inherents are already executed before any transactions, which simplifies things a bit.

This is definitely what our block producing code does, but is it enforced by the runtime somewhere?

If we have this property then we can just make this inherent mandatory, although it will be weird not to have it available in `on_initialize`. I can imagine other modules wanting hooks into `OnValidationParamsIncluded` as a replacement.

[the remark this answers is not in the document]
coriolinus: https://github.com/paritytech/cumulus/pull/70/commits/3eb1618d2cb63b2aa186fa2694da382d3f678a3c should resolve this problem.

parachain-upgrade/src/lib.rs
```diff
@@ -178,9 +166,21 @@ impl<T: Trait> ProvideInherent for Module<T> {
 	const INHERENT_IDENTIFIER: InherentIdentifier = INHERENT_IDENTIFIER;
 
 	fn create_inherent(data: &InherentData) -> Option<Self::Call> {
-		let vfp = data.get_data(&INHERENT_IDENTIFIER)
-			.and_then(|r| r.ok_or_else(|| "Validation Function Params inherent data not found".into()))
-			.expect("Gets and decodes vfp inherent data");
+		let vfp: ValidationFunctionParams = data.get_data(&INHERENT_IDENTIFIER)
+			.expect("Gets and decodes vfp inherent data")
+			.expect("Validation Function Params inherent data not found");
```
rphmeier, line 171: These types of assertions are not allowed in any of our codebases. The approach with panickers is "Prove or remove" - the message of an `expect` statement should be a proof that the expect will not occur. This lets us know which assumptions have been violated when the program encounteres a panic.

There are rare situations in the runtime where we intentionally want to panic during transaction processing or `on_finalize` because we want the block to become invalid. However, this is not one of those situations.

If the inherent is mandatory, I'd recommend returning `None` here but panicking in the `on_finalize` implementation if a `DidUpdate: bool` in storage is not `true`. See what is done in the `frame-timestamp` or `polkadot-runtime-common/parachains` modules - this `DidUpdate` value should never hit disk, so it is an efficient check.

parachain-upgrade/src/lib.rs
```diff
@@ -178,9 +166,21 @@ impl<T: Trait> ProvideInherent for Module<T> {
...
+
+
+		// initialization logic: we know that this runs exactly once every block,
+		// which means we can put the initialization logic here to remove the
+		// sequencing problem.
+		if let Ok((apply_block, validation_function)) = PendingValidationFunction::try_get() {
```
rphmeier, line 177: Can you elaborate a bit more on the context that you expect this initialization code to run in?

My understanding of the `create_inherent` function is that it's run only by the block author at the time of authoring the block. Furthermore, it is run in the context of a Runtime API instance, where this context is not the same as the one we create the block in the block.

As regular importing nodes of the chain will not run this logic, it seems the wrong place to put initialization behavior. Likewise, the changes made during the execution of this function on the authoring node will be immediately discarded and not used in the creation of the block.

parachain-upgrade/src/lib.rs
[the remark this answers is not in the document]
rphmeier: I have left a comment on that snippet; I don't really understand the approach.

parachain-upgrade/src/lib.rs
coriolinus: If that's the case, then this fix will not work. It was suggested [here](https://github.com/paritytech/substrate/issues/5757#issuecomment-618903497); I do not have enough knowledge of the system as is to say whether or not it does what it intends to in all cases. 

> the changes made during the execution of this function on the authoring node will be immediately discarded and not used in the creation of the block.

If that's the case, then an appropriate fix would be to move the pending validation function check into `Call::set_validation_function_parameters`; doing so should ensure those changes persist as part of the created inherent.

coriolinus: > an appropriate fix would be to move the pending validation function check...

This turns out not to be true; when I make that change, tests fail. Returning the call apparently does not execute the call.

parachain-upgrade/src/lib.rs
coriolinus: Good spot, thanks! db26357 does as you say; there was already an `on_finalize` assertion.

parachain-upgrade/src/lib.rs
coriolinus: A new attempt in e9cf4f8 appears to work to address this.

coriolinus:
The CI failure looks like this:

```
    Compiling polkadot-parachain v0.7.30 (https://github.com/paritytech/polkadot?branch=cumulus-branch#d240ef74)
error[E0433]: failed to resolve: could not find `wasm` in `proc_macro_runtime_interface`
  --> /ci-cache/cumulus/cargo/test-linux-stable/git/checkouts/polkadot-4038f27d5e4ea2e8/d240ef7/parachain/src/wasm_api.rs:30:1
   |
30 | #[runtime_interface]
   | ^^^^^^^^^^^^^^^^^^^^ could not find `wasm` in `proc_macro_runtime_interface`
   |
   = note: this error originates in an attribute macro (in Nightly builds, run with -Z macro-backtrace for more info)
error[E0425]: cannot find function `to_substrate_wasm_fn_return_value` in crate `sp_core`
  --> /ci-cache/cumulus/cargo/test-linux-stable/git/checkouts/polkadot-4038f27d5e4ea2e8/d240ef7/parachain/src/wasm_api.rs:61:11
   |
61 |     sp_core::to_substrate_wasm_fn_return_value(&result)
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `sp_core`
error: aborting due to 2 previous errors
```

As this PR doesn't directly touch either polkadot or substrate, I'm not sure what to do about that.

coriolinus:
Whatever is going on with the build failures, it's not related to the enabled feature set:

```sh
$ git diff f741cf0..008d3ea '**/Cargo.toml'
$ git diff db26357..008d3ea '**/Cargo.toml'
$
```

Given that there are also no `Cargo.toml` changes in db26357, which built successfully, this means that 

1. If featureset changes were a problem, they have been reverted to a known-good state.
2. No non-featureset changes should be able to cause build failures in dependencies.
3. Therefore, something is wrong in CI which is independent of the goodness of this PR.

I still don't know how to fix the CI problem, but I believe I can now confidently state that it _is_ a CI problem.

cecton:
```
error[E0425]: cannot find function `to_substrate_wasm_fn_return_value` in crate `sp_core`
  --> /ci-cache/cumulus/cargo/test-linux-stable/git/checkouts/polkadot-4038f27d5e4ea2e8/d240ef7/parachain/src/wasm_api.rs:61:11
   |
61 |     sp_core::to_substrate_wasm_fn_return_value(&result)
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `sp_core`
```
Looks like you leaked std somehow. Did you manage to find the issue?

coriolinus:
> @cecton 
> Did you manage to find the issue?

Yes, merging master yesterday fixed it. 

@rphmeier 
I like those ideas; I'll put them in this afternoon.

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,633 @@
...
+		storage::unhashed::get(VALIDATION_FUNCTION_PARAMS)
+			.expect("validation function params must be set")
+	}
+
+	/// Get validation function parameters.
+	///
+	/// This will return `None` if this module's inherent has not yet run.
+	/// If it returns `Some(_)`, the validation function params are current for this block.
+	pub fn get_validation_function_params() -> Option<ValidationFunctionParams> {
+		if DidUpdateVFPs::try_get().unwrap_or(false) {
```
bkchr, line 162:
```suggestion
		if DidUpdateVFPs::get() {
```

`get()` will fallback to `Default::default()` if the value does not exists and `Default` for bool is `false`.

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,633 @@
...
+
+		fn on_finalize() {
+			assert!(DidUpdateVFPs::take(), "VFPs must be updated once per block");
+		}
+	}
+}
+
+impl<T: Trait> Module<T> {
+	/// Get validation function parameters.
+	fn validation_function_params() -> ValidationFunctionParams {
```
bkchr, line 150: This should be merged with `get_validation_function_params()`

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,633 @@
...
+use frame_support::{
+	decl_error, decl_event, decl_module, decl_storage, ensure, storage, traits::Get,
+	weights::DispatchClass,
+};
+use parachain::primitives::RelayChainBlockNumber;
+use sp_core::storage::well_known_keys;
+use sp_inherents::{ProvideInherent, InherentData, InherentIdentifier};
+use sp_version::RuntimeVersion;
+
+pub type ValidationFunction = Vec<u8>;
```
bkchr, line 47: Docs.

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,633 @@
...
+	/// Set this to `()` if not needed.
+	type OnValidationFunctionParams: OnValidationFunctionParams;
+}
+
+// This pallet's storage items.
+decl_storage! {
+	trait Store for Module<T: Trait> as ParachainUpgrade {
+		// we need to store the new validation function for the span between
+		// setting it and applying it.
+		PendingValidationFunction get(fn new_validation_function): (RelayChainBlockNumber, ValidationFunction);
```
bkchr, line 69:
```suggestion
		PendingValidationFunction get(fn new_validation_function): Option<(RelayChainBlockNumber, ValidationFunction)>;
```

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,633 @@
...
+		// weight data just stolen from Timestamp::set; may be inappropriate
+		#[weight = (0, DispatchClass::Mandatory)]
+		fn set_validation_function_parameters(origin, vfp: ValidationFunctionParams) {
+			ensure_none(origin)?;
+			assert!(!DidUpdateVFPs::exists(), "VFPs must be updated only once in the block");
+
+			// initialization logic: we know that this runs exactly once every block,
+			// which means we can put the initialization logic here to remove the
+			// sequencing problem.
+			if let Ok((apply_block, validation_function)) = PendingValidationFunction::try_get() {
```
bkchr, line 129:
```suggestion
			if let Some((apply_block, validation_function)) = PendingValidationFunction::get() {
```

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,633 @@
...
+				if vfp.relay_chain_height >= apply_block {
+					PendingValidationFunction::kill();
+					Self::put_parachain_code(&validation_function);
+					Self::deposit_event(Event::ValidationFunctionApplied(vfp.relay_chain_height));
+				}
+			}
```
bkchr, line 135: Formatting?

parachain-upgrade/src/lib.rs
bkchr: And than you don't need to use an `expect` here.

parachain-upgrade/src/lib.rs
coriolinus: I'm aware; I chose this approach to make it more explicit. I'll change it in a minute.

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,633 @@
...
+
+impl<T: Trait> ProvideInherent for Module<T> {
+	type Call = Call<T>;
+	type Error = sp_inherents::MakeFatalError<()>;
+	const INHERENT_IDENTIFIER: InherentIdentifier = INHERENT_IDENTIFIER;
+
+	fn create_inherent(data: &InherentData) -> Option<Self::Call> {
+		// If the inherent is not present, this returns None early. This in turn will
+		// cause the on_finalize assertion to fail.
+		let vfp: ValidationFunctionParams = data.get_data(&INHERENT_IDENTIFIER).ok().flatten()?;
```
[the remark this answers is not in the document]
coriolinus: I will change that, but I'd like to talk with you at some point about what the expected semantics are in `create_inherent`, then, because they clearly aren't what I imagined.

parachain-upgrade/src/lib.rs
```diff
@@ -0,0 +1,588 @@
...
+			}
+
+			// TOOD: figure out a better value than this WAG
+			100_000
+		}
+
+		// TODO: figure out a bettwe weight than this WAG
+		#[weight = SimpleDispatchInfo::FixedOperational(1_000_000)]
+		pub fn set_code(origin, validation_function: ValidationFunction) {
+			// TODO: in the future, we can't rely on a superuser existing
```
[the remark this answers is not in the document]
coriolinus: https://github.com/paritytech/cumulus/issues/87

parachain-upgrade/src/lib.rs
coriolinus: I'm not sure what formatting you mean here.

primitives/src/validation_function_params.rs
```diff
@@ -0,0 +1,51 @@
+//! Validation Function Parameters govern the ability of parachains to upgrade their validation functions.
```
bkchr, line 1: License header missing.

coriolinus: I think I need a VS Code extension that nags me if the license header is missing; I just don't think of that at first.

PR #77, opened 2020-04-15
Update XCMP draft to master

JoshOrndorff:
This basically merges master into #16. Two questions about this work.

1. @gavofyork The nomenclature that was already in #16 uses "icmp" to refer to messages from parachains as well as from the relay chain. That's different than the definition of xcmp we discussed recently. Would you like me to update the code to match our recent discussions, or do a more trivial s/icmp/xcmp/g?

2. This compiles when I run `cargo check -p cumulus-runtime` but not when I just run `cargo check` in the root. How can I solve that?


bkchr:
Closing it in favor of: https://github.com/paritytech/cumulus/pull/80

PR #78, opened 2020-04-15
Wrapped DOT Parachain

JoshOrndorff:
This PR adds the ability to host DOTs to the test parachain using the APIs introduced in #80


JoshOrndorff:
I've got something that can almost be played with, but I can't figure out how to create Alice's AccountId in the runtime.

```rust
impl<T: Trait> DownwardMessageHandler for Module<T> {
	fn handle_downward_message(msg: &()) {
		// This will be fleshed out once non-trivial messages are supported.
		// For now, every incoming message represents an inbound transfer of 10 DOTs for Alice.

		// TODO decode these parameters from the message rather than setting them manually.
		//GRRRRRR How the heck do I make Alice's accountID!?
		let dest = sr25519::Pair::from_string(&"//Alice", None)
			.expect("static values are valid; qed")
			.public()
			.into_account();
		let value = T::Balance::from(10);

		// Mint new tokens in the parachain
		Balances::<T>::insert(&dest, Balances::<T>::get(&dest) + value);

		// Emit event for successful inbound token transfer
		Self::deposit_event(RawEvent::TransferIn(dest, value));
	}
}
```

JoshOrndorff:
Closing this in favor of #118

PR #80, opened 2020-04-17
Introduce message broker for receiving and sending relay chain messages

bkchr:
This provides a new crate `cumulus-message-broker` for receiving downward messages and sending upward messages. The messages are currently just `()` and will be replaced in a later pr. This also only introduces the runtime side, providing the inherent data will come later as well.

### DIFF
--- a/message-broker/src/lib.rs
+++ b/message-broker/src/lib.rs
@@ -18,14 +18,16 @@
 //!
 //! This pallet provides support for handling downward and upward messages.
 
-use sp_inherents::{ProvideInherent, InherentData, InherentIdentifier, MakeFatalError};
-
-/// Something that should be called when a downward message is received.
-#[impl_trait_for_tuples::impl_for_tuples(30)]
-pub trait DownwardMessageHandler {
-	/// Handle the given downward message.
-	fn handle_downward_message(msg: &());
-}
+use cumulus_primitives::{
+	inherents::{DownwardMessagesType, DOWNWARD_MESSAGES_IDENTIFIER},
+	well_known_keys, DownwardMessageHandler, UpwardMessageSender,
+};
+use frame_support::{
+	decl_module, storage,
+	weights::{SimpleDispatchInfo, WeighData, Weight},
+};
+use frame_system::ensure_none;
+use sp_inherents::{InherentData, InherentIdentifier, MakeFatalError, ProvideInherent};
 
 /// Configuration trait of this pallet.
 pub trait Trait: frame_system::Trait {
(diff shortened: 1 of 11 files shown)


primitives/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Cumulus related primitive types and traits.
```
JoshOrndorff, line 17:
```suggestion
//! Cumulus-related primitive types and traits.
```

primitives/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+/// Inherents identifiers and types related to
```
JoshOrndorff, line 21:
```suggestion
/// Identifiers and types related to Cumulus Inherents
```

message-broker/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+			SimpleDispatchInfo::default().weigh_data(())
+		}
+	}
+}
+
+impl<T: Trait> UpwardMessageSender for Module<T> {
+	fn send_upward_message(_: &()) -> Result<(), ()> {
+		Ok(())
+	}
+}
```
JoshOrndorff, line 63: Why does the broker pallet need to implement this trait? What kinds of messages will it be sending? Maybe the idea is that it will aggregate messages from other pallets and send them?

Will other pallets that want to send upward messages (eg. WrappedDOT) implement `UpwardMessageSender` themselves? Or will they depend on this broker pallet to send messages on their behalf?

I realize this is a stub implementation, but what is the plan for how these messages will actually reach the relay chain? I guess messages will be appended to the storage at `UPWARD_MESSAGES` and then somhow they will be included in the relay chain block?

bkchr: Yeah it will aggregate it, I may need to rename something. If you have some suggestions, I'm open. 

In your WrappedDot example you would implement `DownwardMessageHandler` for your module and add a `type UpwardMessageSender: UpwardMessageSender;` to your trait. This upward message sender can be used to send these messages from your pallet and will be initialized in the runtime.

> I realize this is a stub implementation, but what is the plan for how these messages will actually reach the relay chain? I guess messages will be appended to the storage at UPWARD_MESSAGES and then somhow they will be included in the relay chain block?

Exactly. I will extract them on the validation side and return them to Polkadot to handle them ;)

PR #83, opened 2020-04-22
Keep unpinned blocks for a while before announcing

cecton:
Related to #7


test/parachain/src/service.rs
```diff
@@ -72,10 +67,12 @@ macro_rules! new_full_start {
 ///
 /// This function blocks until done.
 pub fn run_collator(
-	parachain_config: Configuration,
+	mut parachain_config: Configuration,
 	key: Arc<CollatorPair>,
 	polkadot_config: polkadot_collator::Configuration,
 ) -> sc_service::error::Result<impl AbstractService> {
+	parachain_config.announce_block = false;
```
bkchr, line 74: This should be moved into some function provided by Cumulus. As people will need to replicate this all the time and we may need to "tweak" the config even more in the future.

test/parachain/src/service.rs
```diff
@@ -15,17 +15,12 @@
...
 use sc_service::{AbstractService, Configuration};
 use sc_finality_grandpa::{FinalityProofProvider as GrandpaFinalityProofProvider, StorageAndProofProvider};
-
 use polkadot_primitives::parachain::CollatorPair;
-
 use cumulus_collator::CollatorBuilder;
-
-use futures::FutureExt;
-
+use futures::prelude::*;
```
bkchr, line 23: I'm no huge fan of blindly including preludes. (But it is okay :P )

network/src/lib.rs
```diff
@@ -141,3 +148,105 @@ where
...
+						target: "cumulus-network",
+						"previous task that waits for announce block has been canceled",
+					);
+				}
+			}
+		}).into()) {
+			error!(
+				target: "cumulus-network",
+				"Could not spawn a new task to wait for the announce block: {:?}",
+				err
```
bkchr, line 222:
```suggestion
				err,
```

network/src/lib.rs
```diff
@@ -141,3 +148,105 @@ where
 		Ok(Validation::Success)
 	}
 }
+
+pub struct WaitToAnnounce<Block: BlockT> {
```
bkchr, line 152: Some docs would be nice.

network/src/lib.rs
```diff
@@ -141,3 +148,105 @@ where
...
+	spawner: Arc<dyn Spawn + Send + Sync>,
+	network: Arc<NetworkService<Block, <Block as BlockT>::Hash>>,
```
bkchr, line 154:
```suggestion
	network: Arc<NetworkService<Block, Block::Hash>>,
```

test/parachain/src/service.rs
cecton: :sloth: will fix it :)

network/src/lib.rs
cecton: Yes! I was waiting for you to validate the implementation first to be sure

cecton:
hmm 🤔 that's annoying, I need to create a fake NetworkService to fix the test

collator/src/lib.rs
```diff
@@ -433,13 +463,15 @@ mod tests {
 		let id = ParaId::from(100);
 		let _ = env_logger::try_init();
 		let spawner = futures::executor::ThreadPool::new().unwrap();
+		let announce_block = |_, _| ();
```
cecton, line 466: :rofl:

cecton:
I replaced the NetworkService parameter with a closure that just does `announce_block`. It removed the dependencies on sc_network.

network/src/lib.rs
cecton: done

test/parachain/src/service.rs
cecton: done

test/parachain/src/service.rs
cecton: Should I create a crate `service` for this?

cecton: Ok as discussed I placed it in the `collator` crate

PR #90, opened 2020-05-06
Align trie iteration to substrate.

cheme:
I was checking how proof would be checked with cumulus, and see that clear_prefix is not aligned with substrate, this PR fixes that.
(corresponding substrate code https://github.com/paritytech/substrate/blob/a2512e837003db3d3267764283edf31cf95a206c/primitives/state-machine/src/trie_backend_essence.rs#L239 ).

### DIFF
--- a/runtime/src/validate_block/implementation.rs
+++ b/runtime/src/validate_block/implementation.rs
@@ -215,17 +215,8 @@ impl<B: BlockT> Storage for WitnessStorage<B> {
 			Err(_) => panic!(),
 		};
 
-		let iterator = match TrieDBIterator::new_prefixed(&trie, prefix)
-		{
-			Ok(r) => r,
-			Err(_) => panic!(),
-		};
-
-		for x in iterator {
+		for x in TrieDBIterator::new_prefixed(&trie, prefix).expect("Creates trie iterator") {
 			let (key, _) = x.expect("Iterating trie iterator");
-
-			debug_assert!(!key.starts_with(prefix));
-
 			self.overlay.insert(key, None);
 		}
 	}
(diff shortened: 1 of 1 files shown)


runtime/src/validate_block/implementation.rs
```diff
@@ -215,15 +215,16 @@ impl<B: BlockT> Storage for WitnessStorage<B> {
 			Err(_) => panic!(),
 		};
 
-		let mut iter = trie.iter().expect("Creates trie iterator");
-		iter.seek(prefix).expect("Seek trie iterator");
+		let iterator = match TrieDBIterator::new_prefixed(&trie, prefix)
```
bkchr, line 218: Why can we not use expect here?

runtime/src/validate_block/implementation.rs
```diff
@@ -215,15 +215,16 @@ impl<B: BlockT> Storage for WitnessStorage<B> {
...
+		{
```
bkchr, line 219: This should be moved one line above.

runtime/src/validate_block/implementation.rs
```diff
@@ -215,15 +215,16 @@ impl<B: BlockT> Storage for WitnessStorage<B> {
...
+		};
 
-		for x in iter {
+		for x in iterator {
 			let (key, _) = x.expect("Iterating trie iterator");
 
-			if !key.starts_with(prefix) {
-				break;
-			}
+			debug_assert!(!key.starts_with(prefix));
```
bkchr, line 227: This code will probably never compiled with debug asserts enabled.

cheme: That's sad, I am not sure how it differs from assert, but I am just removing it.

runtime/src/validate_block/implementation.rs
cheme: Dumb copy paste :)

runtime/src/validate_block/implementation.rs
bkchr: `debug_assert!` is only activated on debug builds or when you set `debug-asserts`. However, I assume that we would hit the bug before on the node.

PR #91, opened 2020-05-07
Integration test

cecton:


test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,146 @@
...
+	let mut polkadot_bob = Command::new(cargo_bin("cumulus-test-parachain-collator"))
+		.stdout(Stdio::piped())
+		.stderr(Stdio::piped())
+		.args(POLKADOT_ARGS)
+		.arg("--base-path")
+		.arg(polkadot_bob_dir.path())
+		.arg("--bob")
+		.spawn()
+		.unwrap();
+	let polkadot_bob_clean_up = ProcessCleanUp(&mut polkadot_bob);
```
coriolinus, line 115: Should Bob's node have some arguments to set a different port? Should we `wait_for_tcp` for Bob's port?

cecton: Not for my test based on the manual example but sure you can do whatever you want for your use case.

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,411 @@
...
+// Substrate is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
+
+// TODO: this is necessary for the jsonrpsee macro used
+#![allow(unused_variables, dead_code)]
```
cecton, line 18: I made a ticket for this [there](https://github.com/paritytech/jsonrpsee/issues/106).

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,411 @@
...
+
+	// check output
+	cumulus_child.terminate();
+	assert!(
+		cumulus_child
+			.read_stderr_to_end()
+			.unwrap()
+			.contains("best: #2"),
+		"no parachain blocks seems to have been produced",
+	);
```
coriolinus, line 396: In the event that this test fails, it will be very useful to have the log output.

```suggestion
      let cumulus_log = cumulus_child.read_stderr_to_end().unwrap();
      println!("{}", cumulus_log);
	assert!(
		cumulus_log.contains("best: #2"),
		"no parachain blocks seems to have been produced",
	);
```

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,411 @@
...
+				path.pop();
+			}
+			path
+		})
+		.unwrap()
+}
+
+struct ChildHelper<'a> {
+	name: String,
+	child: &'a mut Child,
```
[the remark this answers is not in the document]
cecton: Yes it started differently, not really as a helper... I will check that out

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,411 @@
...
+			.await
+			.unwrap()
+	})
+	.unwrap()
+	.number
+	.saturated_into::<u64>();
+
+	let genesis_block =
+		async_std::task::block_on(async { Chain::block_hash(&mut client, 0).await.unwrap() })
+			.unwrap();
```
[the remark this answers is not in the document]
cecton: I will check that out

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,411 @@
...
+	assert!(
+		cumulus_child
+			.read_stderr_to_end()
+			.unwrap()
+			.contains("best: #2"),
+		"no parachain blocks seems to have been produced",
+	);
+}
+
+fn find_local_node_identity(instance: &mut ChildHelper) -> String {
```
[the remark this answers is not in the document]
cecton: :ok_hand:

test/parachain/tests/integration_test.rs
cecton: It does already! Small summary: this whole helper struct has a String for stdout and stderr, every time you call a function to get the logs it first add to that string and then return a string slice to that string so nothing is never lost. When the helper is dropped, it always prints out stdout and stderr but if the test succeeds cargo doesn't show what has been printed.

coriolinus: Is the `Drop` implementation guaranteed to fire before the failed assertion aborts the test? It may be worth writing a small secondary test which is `#[ignore]`d, just to demonstrate that that output does in fact get captured and emitted.

My impression was that `assert` macro failures are panics under the hood, and may not necessarily unwind all `Drop`s.

cecton: assert does panic. panics call all the Drops. I don't see the value to adding such a test. The day this test fail someone will see the logs and if they don't fix it they will fix it. A test that needs to be run by human is guaranteed to be never run imo

coriolinus: OK, fair enough.

cecton: Wait, you said:
> may not necessarily unwind all Drops.

I read that it does unwind. Do you know for sure I should not rely on that behavior??

test/parachain/tests/integration_test.rs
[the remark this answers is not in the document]
cecton: converted the whole test to async

test/parachain/tests/integration_test.rs
coriolinus: I am not certain that you can not rely on that behavior; this isn't an area I've got a lot of experience in. I'm just aware of a bunch of chatter in the community that [you can't rely on `Drop` being executed in all cases](https://doc.rust-lang.org/nightly/nomicon/leaking.html).

Still, in this case at least, it looks like it does work properly: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=53f6cea04710cc1ed2fad0d34a18a3ee

test/parachain/tests/integration_test.rs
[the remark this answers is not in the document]
cecton: I think I will keep that one as it as I think this is not meant to wrap and hide the child process. It might be useful for the tester to still be able to access the child. I did rename the variable to make it clear it's not the child but the helper

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,404 @@
...
+			.as_mut()
+			.ok_or_else(|| io::Error::new(io::ErrorKind::Other, "stdout not captured"))?
+			.read_to_string(&mut output)?;
+		self.stdout.push_str(output.as_str());
+
+		Ok(&self.stdout)
+	}
+
+	fn read_stderr_to_end(&mut self) -> io::Result<&str> {
+		let mut output = String::new();
```
bkchr, line 136: Do you see the similarities of this function to the one above? :D

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,404 @@
...
+		.arg(format!(
+			"--bootnodes=/ip4/127.0.0.1/tcp/50666/p2p/{}",
+			polkadot_bob_id
+		))
+		.spawn()
+		.unwrap();
+	let mut cumulus_helper = ChildHelper::new("cumulus", &mut cumulus);
+
+	// wait for blocks to be generated
+	thread::sleep(Duration::from_secs(60));
```
bkchr, line 393: Never sleep in tests! ;) 

Why can we not check that the best blocks progresses using the rpc call?

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,404 @@
...
+
+		let _ = self.child.wait();
+	}
+}
+
+fn tcp_port_is_open<A: net::ToSocketAddrs>(address: A) -> bool {
+	match net::TcpStream::connect(&address) {
+		Ok(_) => true,
+		Err(_) => false,
+	}
```
bkchr, line 199:
```suggestion
	net::TcpStream::connect(&address).is_ok()
```

Rust docs is your friend.

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,404 @@
...
+fn wait_for_tcp<A: net::ToSocketAddrs>(address: A) -> io::Result<()> {
+	let mut tries = 10;
+
+	loop {
+		tries -= 1;
+
+		match net::TcpStream::connect(&address) {
+			Ok(_) => break Ok(()),
+			Err(err) if tries == 0 => break Err(err),
+			_ => thread::sleep(Duration::from_secs(1)),
```
bkchr, line 211: This breaks faster as you can say "break".

If the CI is under high load, waiting 10 seconds may not be enough.

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,404 @@
...
+		.spawn()
+		.unwrap();
+	let polkadot_bob_helper = ChildHelper::new("bob", &mut polkadot_bob);
+	wait_for_tcp("127.0.0.1:9934").unwrap();
+
+	// wait a bit for some relay chains blocks to be generated
+	thread::sleep(Duration::from_secs(10));
+
+	// export genesis state
+	let cmd = Command::new(cargo_bin("cumulus-test-parachain-collator"))
```
bkchr, line 263: You can just call the code that generate the genesis state, instead of reading it from stdout.

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,404 @@
...
+		.arg("--bob")
+		.arg("--unsafe-rpc-expose")
+		.arg("--rpc-port=9934")
+		.spawn()
+		.unwrap();
+	let polkadot_bob_helper = ChildHelper::new("bob", &mut polkadot_bob);
+	wait_for_tcp("127.0.0.1:9934").unwrap();
+
+	// wait a bit for some relay chains blocks to be generated
+	thread::sleep(Duration::from_secs(10));
```
bkchr, line 260: Why?

test/parachain/tests/integration_test.rs
cecton: > I think we should just do everything with unbounded waiting and put an overall kill after like 10min into the code.

So I just make it wait indefinitely and put a overall timer?

test/parachain/tests/integration_test.rs
cecton: What do you want exactly? A macro?

test/parachain/tests/integration_test.rs
cecton: I think it's not useful anymore because we wait for bob to be running

test/parachain/tests/integration_test.rs
cecton: The code is in command.rs, the crate is not a lib, is there a way to import a binary like that? Or should I move this somewhere else?

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,404 @@
...
+use codec::Encode;
+use polkadot_primitives::parachain::{Info, Scheduling};
+use polkadot_primitives::Hash as PHash;
+use polkadot_runtime::{Header, OnlyStakingAndClaims, Runtime, SignedExtra, SignedPayload};
+use polkadot_runtime_common::{parachains, registrar, BlockHashCount};
+use serde_json::Value;
+use sp_arithmetic::traits::SaturatedConversion;
+use sp_runtime::generic;
+use sp_version::RuntimeVersion;
+use std::io::Read;
```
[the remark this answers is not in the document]
cecton: done

test/parachain/tests/integration_test.rs
cecton: done

test/parachain/tests/integration_test.rs
cecton: done

test/parachain/tests/integration_test.rs
cecton: done

test/parachain/tests/integration_test.rs
coriolinus: I'd imagine a function like this:

```rust
fn read_to_end<'a, r: Read>(reader: &mut Option<R>, store: &'a mut String) -> io::Result<&'a str> {
    let mut output = String::new():

    reader
        .as_mut()
        .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "not captured"))?
        .read_to_string(&mut output)?;
    store.push_str(output.as_str())

    Ok(&output)
}

fn read_stdout_to_end(&mut self) -> io::Result<&str> {
    Self::read_to_end(&mut self.child.stdout, &mut self.stdout)
}
```

cecton: Actually I don't think I need those functions anymore since I don't need to keep track of what has been read already. I will remove both.

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,399 @@
...
+async fn wait_for_tcp<A: net::ToSocketAddrs + std::fmt::Display>(address: A) {
+	loop {
+		match net::TcpStream::connect(&address) {
+			Ok(_) => break,
+			Err(err) => {
+				eprintln!("Waiting for {} to be up ({})...", address, err);
+				sleep(Duration::from_secs(2)).await;
+			}
+		}
+	}
```
coriolinus, line 184: Simplification

```suggestion
	while let Err(err) = net::TcpStream::connect(&address) {
		eprintln!("Waiting for {} to be up ({})...", address, err);
		sleep(Duration::from_secs(2)).await;
	}
```

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,399 @@
...
+			.stderr(Stdio::piped())
+			.args(POLKADOT_ARGS)
+			.arg("--base-path")
+			.arg(polkadot_alice_dir.path())
+			.arg("--alice")
+			.arg("--unsafe-rpc-expose")
+			.spawn()
+			.unwrap();
+		let polkadot_alice_helper = ChildHelper::new("alice", &mut polkadot_alice);
+		wait_for_tcp("127.0.0.1:9933").await;
```
coriolinus, line 209: Test might run slightly faster if we start both Alice and Bob's nodes, then join on both `wait_for_tcp` calls.

cecton: Really not worth the effort imo

cecton: (also not really good for readability)

coriolinus: Sure, agree it's not likely a huge savings.

test/parachain/tests/integration_test.rs
cecton: I don't think it's resolved as the code is still duplicated. But I think it's fine as it

cecton:
> I like the way you wait for four observed parachain blocks.

Thanks :)

> I don't know why async sleep is OK on CI if normal sleep is not. Assuming that it is in fact OK, then I think this should be approvable.

I don't think that's what @bkchr meant. I think he meant that the resources of the CI can be overloaded and therefore waiting for x seconds for a condition to become true is unreliable. This is really true of course. You can definitely sleep in a CI but you should preferably not expect for a condition to become true after any amount of time (even a reasonable one).

Though at some point you still need to make some reasonable assumption thus the 10 min for the overall test.

> [edit] On my machine, given a pre-built release test target, `cargo test --release -- --ignored integration_test` passes in 1m40s real time. This seems plausible to me.

Indeed! But if the conditions are really bad, the test is allowed to run for a total of 10 min which is overly excessive for a local test but reasonable for CI. Maybe this is something we should parametrize. Thanks for the idea!

bkchr:
> This is really true of course. You can definitely sleep in a CI but you should preferably not expect for a condition to become true after any amount of time (even a reasonable one).

Exactly!

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,390 @@
...
+use futures::{future::FutureExt, pin_mut, select};
+use polkadot_primitives::parachain::{Info, Scheduling};
+use polkadot_primitives::Hash as PHash;
+use polkadot_runtime::{Header, OnlyStakingAndClaims, Runtime, SignedExtra, SignedPayload};
+use polkadot_runtime_common::{parachains, registrar, BlockHashCount};
+use serde_json::Value;
+use sp_arithmetic::traits::SaturatedConversion;
+use sp_runtime::generic;
+use sp_version::RuntimeVersion;
+use std::collections::HashSet;
```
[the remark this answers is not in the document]
cecton: My next personal project will be a tool that re-organize and sort all the imports XD

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,390 @@
...
+			Ok(None) => {}
+			Err(err) => {
+				eprintln!("could not wait for child process to finish: {}", err);
+				let _ = self.child.kill();
+				let _ = self.child.wait();
+				return;
+			}
+		}
+
+		#[cfg(unix)]
```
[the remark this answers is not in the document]
cecton: The default kill method does really send SIGKILL to the process. This one will attempt to terminate gracefully if it can... I'm not sure if that's absolutely necessary but it's less surprising: if I want to restart a node using the existing config dir or if I want to check what's inside after exit...

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,390 @@
...
+			.stderr(Stdio::piped())
+			.args(POLKADOT_ARGS)
+			.arg("--base-path")
+			.arg(polkadot_alice_dir.path())
+			.arg("--alice")
+			.arg("--unsafe-rpc-expose")
+			.spawn()
+			.unwrap();
+		let polkadot_alice_helper = ChildHelper::new("alice", &mut polkadot_alice);
+		wait_for_tcp("127.0.0.1:9933").await;
```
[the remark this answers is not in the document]
cecton: Seems really overkill imo but ok I'll do it

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,390 @@
...
+
+async fn wait_for_tcp<A: net::ToSocketAddrs + std::fmt::Display>(address: A) {
+	while let Err(err) = net::TcpStream::connect(&address) {
+		eprintln!("Waiting for {} to be up ({})...", address, err);
+		sleep(Duration::from_secs(2)).await;
+	}
+}
+
+#[async_std::test]
+#[ignore]
```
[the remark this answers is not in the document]
cecton: It's complicated.

Originally I wanted to make the test only run with `--release`. Unfortunately there isn't a flag for that and [the common way to do this is to check the flag debug_assertions instead](https://stackoverflow.com/questions/39204908/how-to-check-release-debug-builds-using-cfg-in-rust#39205417).

But... the CI explicitly enable the `debug_assertions` with `--release` during the tests. So that doesn't work.

The closest way I found to solve this issue is [here](https://www.reddit.com/r/rust/comments/3i1nki/how_to_skip_expensive_tests_with_cargo_test/)

Do you have another idea?

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,390 @@
...
+	assert!(
+		!net::TcpStream::connect("127.0.0.1:9934").is_ok(),
+		"tcp port is already open 127.0.0.1:9934, this test cannot be run",
+	);
+
+	let t1 = sleep(Duration::from_secs(60 * 10)).fuse();
+	let t2 = async {
+		// start alice
+		let polkadot_alice_dir = tempdir().unwrap();
+		let mut polkadot_alice = Command::new(cargo_bin("cumulus-test-parachain-collator"))
```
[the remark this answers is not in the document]
cecton: done

test/parachain/tests/integration_test.rs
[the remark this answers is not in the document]
cecton: Actually it doesn't change much :joy: done!

But don't merge because the test is flaky all of a sudden... maybe it's related to this change.

test/parachain/tests/integration_test.rs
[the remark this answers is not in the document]
cecton: done

test/parachain/tests/integration_test.rs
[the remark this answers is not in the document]
coriolinus: Interestingly, as of 19b7942ee, running the test in debug mode fails in only 11m30s. Presumably, 90s was spent in setup, and the other 10 minutes was spent attempting to gather blocks.

The build mode of the test suite shouldn't affect the runtime of the chains themselves, which are run via the `cargo_run` helper; I imagine that that always builds in release mode regardless. 

I think this data may support running this test unconditionally, instead of gating on release.

test/parachain/tests/integration_test.rs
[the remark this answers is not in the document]
cecton: I think it comes from the p2p port. It depends on which order it is opened. So maybe I should set a port manually too :sweat_smile:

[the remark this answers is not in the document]
cecton: done

test/parachain/tests/integration_test.rs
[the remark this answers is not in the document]
cecton: Nah this made my life harder. I didn't know I had to run in release and I didn't realize it was a problem (I don't think this is obvious at all). I prefer to make it very explicit that this test needs to run in release, even adding a note or a safeguard if I could.

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,403 @@
...
+	path::PathBuf,
+	process::{Child, Command, Stdio},
+	thread,
+	time::Duration,
+};
+use substrate_test_runtime_client::AccountKeyring::Alice;
+use tempfile::tempdir;
+
+static POLKADOT_ARGS: &[&str] = &["polkadot", "--chain=res/polkadot_chainspec.json"];
+static INTEGRATION_TEST_ALLOWED_TIME: Option<&str> = option_env!("INTEGRATION_TEST_ALLOWED_TIME");
```
cecton, line 46: I added this so we can change the timeout locally easily

test/parachain/tests/integration_test.rs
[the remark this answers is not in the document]
bkchr: You mean `rustfmt`?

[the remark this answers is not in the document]
cecton: rustfmt doesnt group imports like that apparently.

No I meant "done" as I moved the HashSet in the std import below.

test/parachain/tests/integration_test.rs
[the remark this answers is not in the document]
bkchr: If it can not end it gracefully, there is a bug.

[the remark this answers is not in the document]
bkchr: I think we should remove this extra code here.

PR #92, opened 2020-05-07
Add storage_append to runtime

cecton:


cecton:
It's one of the method of the runtime. If the method is not declared, it falls back to the default implementation that just panics. Literally a runtime exception XD

coriolinus:
@bkchr Is there any reason not to merge this immediately?

runtime/src/validate_block/implementation.rs
```diff
@@ -220,6 +233,24 @@ impl<B: BlockT> Storage for WitnessStorage<B> {
 			self.overlay.insert(key, None);
 		}
 	}
+
+	fn storage_append(&mut self, key: &[u8], value: Vec<u8>) {
+		let key_vec = key.to_vec();
+		let mut value_vec = Vec::with_capacity(1);
+		value_vec.push(EncodeOpaqueValue(value));
```
bkchr, line 240:
```suggestion
		let value_vec = vec![EncodeOpaqueValue(value)];
```

runtime/src/validate_block/implementation.rs
```diff
@@ -220,6 +233,24 @@ impl<B: BlockT> Storage for WitnessStorage<B> {
...
+
+		if let Some(Some(item)) = self.overlay.remove(&key_vec) {
```
bkchr, line 242: You want to use the `entry` api here.

bkchr:
> @bkchr Is there any reason not to merge this immediately?

Sometimes even I take some time to review prs :D

runtime/src/validate_block/implementation.rs
cecton: I don't think I want this time. I'm doing a remove to extract the value to transform it and re-insert it in the hashmap. I don't need a mutable reference here

runtime/src/validate_block/implementation.rs
cecton: Not available in nostd. Or am I missing something?

bkchr: `sp_std::vec!`

runtime/src/validate_block/implementation.rs
bkchr: You want it ;) `sp_std::mem::take` is your friend.

runtime/src/validate_block/implementation.rs
cecton: ok fixed

runtime/src/validate_block/implementation.rs
```diff
@@ -220,6 +233,20 @@ impl<B: BlockT> Storage for WitnessStorage<B> {
...
+
+		if let Some(item) = sp_std::mem::take(current_value) {
+			*current_value = Some(match Vec::<EncodeOpaqueValue>::append_or_new(item, &value_vec) {
+				Ok(item) => item,
+				Err(_) => value_vec.encode(),
+			});
+		} else {
+			*current_value = Some(value_vec.encode());
+		}
+	}
```
cecton, line 249: Implementation with the entry api instead of remove/insert. Is it good? :sweat_smile: I guess since it doesn't remove and then re-insert, this saves an allocation

runtime/src/validate_block/implementation.rs
cecton: Probably fixed in here: https://github.com/paritytech/cumulus/pull/92/files#r422010526

runtime/src/validate_block/implementation.rs
bkchr, line 249:
```suggestion
		let item = current_value.take().unwrap_or_default();
		*current_value = Some(match Vec::<EncodeOpaqueValue>::append_or_new(item, &value_vec) {
			Ok(item) => item,
			Err(_) => value_vec.encode(),
		});
	}
```

PR #96, opened 2020-05-13
Use JustifiedBlockAnnounceValidator for parachain block announce validator

cecton:
Related to #7


cecton:
I forgot to turn it to Draft

test/parachain/tests/integration_test.rs
```diff
@@ -0,0 +1,454 @@
...
+			if previous_blocks.insert(current_block_hash) {
+				eprintln!("new parachain block: {}", current_block_hash);
+
+				if previous_blocks.len() == number_of_blocks {
+					break;
+				}
+			}
+
+			sleep(Duration::from_secs(2)).await;
+		}
```
cecton, line 442: I also added a test in the integration test but it is not super useful, it will only detect something if the implementation is broken but not if the implementation returns always Success like the default implementation of BlockAnnounceValidator

test/parachain/tests/integration_test.rs
```diff
@@ -329,21 +329,72 @@ async fn integration_test() {
...
+
+		// run cumulus dave
+		let cumulus_dave_dir = tempdir().unwrap();
+		let mut cumulus_dave = Command::new(cargo_bin("cumulus-test-parachain-collator"))
```
bkchr, line 365: What does this test?

test/parachain/tests/integration_test.rs
```diff
@@ -329,21 +329,72 @@ async fn integration_test() {
...
-		let transport_client_cumulus =
+		let transport_client_cumulus_charlie =
 			jsonrpsee::transport::http::HttpTransportClient::new("http://127.0.0.1:27017");
-		let mut client_cumulus = jsonrpsee::raw::RawClient::new(transport_client_cumulus);
+		let mut client_cumulus_charlie = jsonrpsee::raw::RawClient::new(transport_client_cumulus_charlie);
 
 		// wait for parachain blocks to be produced
 		let number_of_blocks = 4;
 		let mut previous_blocks = HashSet::with_capacity(number_of_blocks);
 		loop {
```
bkchr, line 346: I expect that these tests are split and not that we just grow this one test. Or that we make the tests based on each other (by reusing the db or similar).

However, we should at least start to refactor stuff into functions. You just copied the same code.

network/src/lib.rs
```diff
@@ -148,6 +148,38 @@ where
...
+}
+
+impl<B: BlockT> BlockAnnounceValidator<B> for DelayedBlockAnnounceValidator<B>
+{
+	fn validate(
+		&mut self,
+		header: &B::Header,
+		data: &[u8],
+	) -> Result<Validation, Box<dyn std::error::Error + Send>> {
+		match *self.0.lock().unwrap() {
```
bkchr, line 176: No plain `unwrap`, use `parking_lot` if you don't want a result when locking.

network/src/lib.rs
```diff
@@ -148,6 +148,38 @@ where
...
+			Some(ref mut validator) => validator.validate(header, data),
```
bkchr, line 177: You can just use `as_mut().expect("`BlockAnnounceValidator` is set before validating the first announcement; qed")` to remove this match.

network/src/lib.rs
```diff
@@ -148,6 +148,38 @@ where
...
+}
+
+impl<B: BlockT> Clone for DelayedBlockAnnounceValidator<B> {
+	fn clone(&self) -> DelayedBlockAnnounceValidator<B> {
+		DelayedBlockAnnounceValidator(self.0.clone())
+	}
+}
+
+impl<B: BlockT> BlockAnnounceValidator<B> for DelayedBlockAnnounceValidator<B>
+{
```
bkchr, line 170:
```suggestion
impl<B: BlockT> BlockAnnounceValidator<B> for DelayedBlockAnnounceValidator<B> {
```

network/src/lib.rs
```diff
@@ -148,6 +148,38 @@ where
 	}
 }
 
+pub struct DelayedBlockAnnounceValidator<B: BlockT>(Arc<Mutex<Option<Box<dyn BlockAnnounceValidator<B> + Send>>>>);
```
bkchr, line 151: Docs and especially why we do this! + an issue that we need to fix this.

test/parachain/tests/integration_test.rs
cecton: As I said in https://github.com/paritytech/cumulus/pull/96#discussion_r426663407

I also added a test in the integration test but it is not super useful, it will only detect something if the implementation is broken but not if the implementation returns always Success like the default implementation of BlockAnnounceValidator.

So for example if the implementation is broken and no block at all are validated, then we would see on Dave that no block are being imported and the test would fail.

test/parachain/tests/integration_test.rs
cecton: I just refactored. Should I go further or is it good for now?

network/src/lib.rs
cecton: done

bkchr: You should create an issue for this. And this should contain this info.

cecton: Ok! Sorry, created #104

cecton: done

PR #102, opened 2020-05-19
Consistency fix for repositories in Cargo.toml

cecton:
Currently there are inconsistencies: in parachain-upgrade it's using the .git at the end of the url and everywhere else it is not.

Apparently the correct git HTTPS url is with ".git" so I added it instead of removing it.

This PR is just for consistency sake because it makes my regex more complicated when it's not consistent.

### DIFF
--- a/collator/Cargo.toml
+++ b/collator/Cargo.toml
@@ -6,21 +6,21 @@ edition = "2018"
 
 [dependencies]
 # Substrate dependencies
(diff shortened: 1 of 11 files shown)


cecton:
> Normally, if someone does these kind of changes, they switch everything to the dominant part. Aka, please just change the upgrade pallet.

That's what I intended to do until I realized:

image

Not that I mind though...

bkchr:
Yeah I know what github offers you, but we also don't use the ".git" extension everywhere, plus this makes it even more complicated to write a regex.

PR #103, opened 2020-05-19
Ensure relay chain parent is the relay chain head 

cecton:
Related to #7


collator/src/lib.rs
```diff
@@ -276,11 +282,13 @@ where
 				parachain::HeadData(head_data.encode()),
 			);
 
-			wait_to_announce.lock().wait_to_announce(
-				hash,
-				relay_chain_parent,
-				encoded_header,
-			);
+			if relay_chain_parent == best_hash {
```
bkchr, line 285: This is not required. This code is only called with the best hash.

network/src/lib.rs
```diff
@@ -106,6 +109,23 @@ where
 			},
 		} = gossip_statement;
 
+		// Check that the relay chain parent of the block is the relay chain head
+		let best_number = self.polkadot_client.info().best_number;
+
+		match self.polkadot_client.number(relay_chain_leaf) {
```
bkchr, line 115:
```suggestion
		match self.polkadot_client.number(relay_chain_leaf)? {
```

Doesn't this work?

network/src/lib.rs
```diff
@@ -106,6 +109,23 @@ where
...
+				return Err(Box::new(ClientError::Backend(format!(
+					"could not find block number for {}: {}",
+					relay_chain_leaf, err,
+				))));
+			},
+			Ok(Some(x)) if x == best_number => {},
+			Ok(None) => {
+				return Err(Box::new(ClientError::UnknownBlock(relay_chain_leaf.to_string())));
+			},
+			Ok(Some(_)) => return Ok(Validation::Failure),
```
bkchr, line 126: Having a trace log would be nice.

network/src/tests.rs
```diff
@@ -0,0 +1,456 @@
...
+		"validation should fail if the relay chain leaf does not exist"
+	);
+	assert!(matches!(
+		*res.unwrap().downcast::<ClientError>().unwrap(),
+		ClientError::UnknownBlock(_)
+	));
+}
+
+#[test]
+fn check_relay_parent_does_not_fail() {
```
bkchr, line 161: It does not fail but the implementation checks for failing?

network/src/tests.rs
```diff
@@ -0,0 +1,456 @@
...
+	) -> std::result::Result<sc_client_api::blockchain::BlockStatus, sp_blockchain::Error> {
+		Ok(sc_client_api::blockchain::BlockStatus::Unknown)
+	}
+
+	fn number(
+		&self,
+		hash: PHash,
+	) -> std::result::Result<Option<NumberFor<PBlock>>, sp_blockchain::Error> {
+		Ok(match hash.as_fixed_bytes() {
+			&[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] => {
```
bkchr, line 436: We should use `if hash == H256::from_low_u64_be(1)` etc here. That is much shorter.

network/src/lib.rs
cecton: hmmm no :disappointed: 

```
   --> network/src/lib.rs:115:54
    |
115 |         match self.polkadot_client.number(relay_chain_leaf)? {
    |                                                            ^ the trait `std::convert::From<sp_blockchain::error::Error>` is not implemented for `std::boxed::Box<dyn std::error::Error + std::marker::Send>`
    |
```

network/src/tests.rs
cecton: yeah makes no sense, will rename that

network/src/tests.rs
cecton: done

network/src/lib.rs

PR #105, opened 2020-05-27
Reject blocks without justification which don't have the best number

cecton:
Related to #7

### DIFF
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -24,7 +24,7 @@ mod tests;
 use sp_api::ProvideRuntimeApi;
 use sp_blockchain::{Error as ClientError, HeaderBackend};
 use sp_consensus::block_validation::{BlockAnnounceValidator, Validation};
-use sp_runtime::{generic::BlockId, traits::Block as BlockT};
+use sp_runtime::{generic::BlockId, traits::{Block as BlockT, Header as HeaderT}};
 
 use polkadot_collator::Network as CollatorNetwork;
 use polkadot_network::legacy::gossip::{GossipMessage, GossipStatement};
@@ -51,24 +51,27 @@ use parking_lot::Mutex;
 /// the justification.
 ///
 /// Note: if no justification is provided the annouce is considered valid.
-pub struct JustifiedBlockAnnounceValidator<B, P> {
+pub struct JustifiedBlockAnnounceValidator<B, P, C> {
 	phantom: PhantomData<B>,
 	polkadot_client: Arc<P>,
+	parachain_client: Arc<C>,
 }
 
-impl<B, P> JustifiedBlockAnnounceValidator<B, P> {
-	pub fn new(polkadot_client: Arc<P>) -> Self {
+impl<B, P, C> JustifiedBlockAnnounceValidator<B, P, C> {
+	pub fn new(polkadot_client: Arc<P>, parachain_client: Arc<C>) -> Self {
 		Self {
 			phantom: Default::default(),
 			polkadot_client,
+			parachain_client,
 		}
 	}
 }
 
-impl<B: BlockT, P> BlockAnnounceValidator<B> for JustifiedBlockAnnounceValidator<B, P>
+impl<B: BlockT, P, C> BlockAnnounceValidator<B> for JustifiedBlockAnnounceValidator<B, P, C>
 where
 	P: ProvideRuntimeApi<PBlock> + HeaderBackend<PBlock>,
 	P::Api: ParachainHost<PBlock>,
+	C: HeaderBackend<B>,
 {
 	fn validate(
 		&mut self,
@@ -78,6 +81,19 @@ where
 		// If no data is provided the announce is valid.
 		if data.is_empty() {
 			return Ok(Validation::Success);
+		}
+
+		// Check if block is one higher than best
+		let best_number = self.parachain_client.info().best_number;
+		let block_number = header.number();
+
+		if !(block_number == best_number + 1 || block_number == best_number) {
+			trace!(
+				target: "cumulus-network",
+				"validation failed because the block number is not the best block number or one higher",
+			);
+
+			return Ok(Validation::Failure);
 		}
 
 		// Check data is a gossip message.
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -83,7 +83,7 @@ where
 		let polkadot_info = self.polkadot_client.info();
 
 		if data.is_empty() {
-			// Check if block is one higher than best
+			// Check if block is equal or higher than best (this requires a justification)
 			let runtime_api_block_id = BlockId::Hash(polkadot_info.best_hash);
 			let block_number = header.number();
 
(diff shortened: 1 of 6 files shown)


network/src/lib.rs
```diff
@@ -77,7 +77,12 @@ where
 	) -> Result<Validation, Box<dyn std::error::Error + Send>> {
 		// If no data is provided the announce is valid.
```
JoshOrndorff, line 78: Comment needs updating

cecton: fixed but the whole code changed

network/src/lib.rs
```diff
@@ -51,32 +51,48 @@ use parking_lot::Mutex;
...
 		mut data: &[u8],
 	) -> Result<Validation, Box<dyn std::error::Error + Send>> {
-		// If no data is provided the announce is valid.
+		// If no data is provided the announce is probably valid
 		if data.is_empty() {
+			// Check if block is one higher than best
+			let best_number = self.parachain_client.info().best_number;
+			let block_number: <<B as BlockT>::Header as HeaderT>::Number = *header.number();
+
+			if !(block_number == best_number + One::one() || block_number == best_number) {
```
bkchr, line 87: Okay, I thought again about this. And I made a mistake. 

We should extract the last parachain header that was stored on the relay chain. We get the number of this Parachain header an make sure that:
`block_number == parachain_header_number + 1`

This way we make sure that we support forks on the same height

bkchr: If this condition holds, we need to make sure that there is a justification.

network/src/lib.rs
```diff
@@ -45,6 +45,12 @@ use log::{error, trace};
 use std::{marker::PhantomData, sync::Arc};
 use parking_lot::Mutex;
 
+/// The head data of the parachain, stored in the relay chain.
```
bkchr, line 48: Please move that into `primitives` and consolidate with the one in `collator`

network/src/lib.rs
```diff
@@ -75,8 +83,34 @@ where
...
 		mut data: &[u8],
 	) -> Result<Validation, Box<dyn std::error::Error + Send>> {
-		// If no data is provided the announce is valid.
+		let runtime_api = self.polkadot_client.runtime_api();
+		let polkadot_info = self.polkadot_client.info();
+
 		if data.is_empty() {
+			// Check if block is one higher than best
+			let runtime_api_block_id = BlockId::Hash(polkadot_info.best_hash);
+			let block_number: <<B as BlockT>::Header as HeaderT>::Number = *header.number();
```
bkchr, line 92:
```suggestion
			let block_number = header.number().clone();
```

network/src/lib.rs
```diff
@@ -75,8 +83,34 @@ where
...
+
+			let local_validation_data = runtime_api
+				.local_validation_data(&runtime_api_block_id, self.para_id)
+				.map_err(|e| Box::new(ClientError::Msg(format!("{:?}", e))) as Box<_>)?
+				.ok_or_else(|| {
+					Box::new(ClientError::Msg("no local validation data".to_string())) as Box<_>
```
bkchr, line 98:
```suggestion
					Box::new(ClientError::Msg("Could not find parachain head in relay chain".into())) as Box<_>
```

network/src/lib.rs
```diff
@@ -75,8 +83,34 @@ where
...
+				})?;
+			let parent_head = HeadData::<B>::decode(&mut &local_validation_data.parent_head.0[..])
+				.map_err(|e| Box::new(ClientError::Msg(format!("{:?}", e))) as Box<_>)?;
```
bkchr, line 101:
```suggestion
				.map_err(|e| Box::new(ClientError::Msg(format!("Failed to decode parachain head: {:?}", e))) as Box<_>)?;
```

network/src/lib.rs
```diff
@@ -75,8 +83,34 @@ where
...
+			let known_best_number = *parent_head.header.number();
+
+			if !(block_number < known_best_number) {
```
bkchr, line 104:
```suggestion
			let res = if block_number >= known_best_number {
```

Negations always makes stuff much more complicated.

network/src/lib.rs
```diff
@@ -75,8 +83,34 @@ where
 		header: &B::Header,
 		mut data: &[u8],
 	) -> Result<Validation, Box<dyn std::error::Error + Send>> {
-		// If no data is provided the announce is valid.
+		let runtime_api = self.polkadot_client.runtime_api();
+		let polkadot_info = self.polkadot_client.info();
+
 		if data.is_empty() {
+			// Check if block is one higher than best
```
bkchr, line 90: You need a justification if this is a new best block.

network/src/lib.rs
```diff
@@ -75,8 +83,34 @@ where
...
+				trace!(
+					target: "cumulus-network",
+					"validation failed because the block number is not at least the best block \
+					number known",
+				);
+
+				return Ok(Validation::Failure);
+			}
+
 			return Ok(Validation::Success);
```
bkchr, line 114:
```suggestion
				Validation::Failure
			} else {
				Validation::Success
			};

			return Ok(res);
```

network/src/lib.rs

network/src/lib.rs

network/src/lib.rs

network/src/lib.rs
cecton: eww no, the rest of the function follows the pattern:

```
if <condition> {
    return ... // short-circuit
}
```

network/src/lib.rs
cecton: like this? https://github.com/paritytech/cumulus/pull/105/commits/2f9131c563d74a36b7e34f989f334e0eac8288d1

network/src/lib.rs
bkchr: Yeah, but here a `else` makes sense.

primitives/src/lib.rs
```diff
@@ -20,6 +20,9 @@
 
 pub mod validation_function_params;
 
+use codec::{Decode, Encode};
+use sp_runtime::traits::{Block as BlockT};
```
bkchr, line 24:
```suggestion
use sp_runtime::traits::Block as BlockT;
```

network/src/lib.rs
```diff
@@ -75,9 +79,36 @@ where
...
+					target: "cumulus-network",
+					"validation failed because a justification is needed if this is a new best \
+					block",
+				);
+
+				Validation::Failure
+			} else {
+				Validation::Success
+			});
+
```
bkchr, line 111:
```suggestion
```

network/src/lib.rs
```diff
@@ -75,9 +79,36 @@ where
...
+					Box::new(ClientError::Msg("Could not find parachain head in relay chain".into())) as Box<_>
+				})?;
+			let parent_head = HeadData::<B>::decode(&mut &local_validation_data.parent_head.0[..])
+				.map_err(|e| Box::new(ClientError::Msg(format!("Failed to decode parachain head: {:?}", e))) as Box<_>)?;
+			let known_best_number = parent_head.header.number();
+
+			return Ok(if block_number >= known_best_number {
+				trace!(
+					target: "cumulus-network",
+					"validation failed because a justification is needed if this is a new best \
```
bkchr, line 103:
```suggestion
					"validation failed because a justification is needed if the block at the top of the chain."
```

network/src/lib.rs
```diff
@@ -75,9 +79,36 @@ where
...
+					block",
```
bkchr, line 104:
```suggestion
```

PR #112, opened 2020-06-15
Update polkadot & substrate

cecton:
This will bring the `--tmp` which is useful for demo

### DIFF
--- a/test/parachain/src/command.rs
+++ b/test/parachain/src/command.rs
@@ -253,7 +253,8 @@ impl CliConfiguration for PolkadotCli {
 		Ok(self
 			.shared_params()
 			.base_path()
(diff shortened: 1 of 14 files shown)


test/parachain/src/command.rs
```diff
@@ -210,10 +209,11 @@ pub fn run() -> Result<()> {
 					.chain(cli.relaychain_args.iter()),
 			);
 
-			polkadot_cli.base_path = cli.run.base_path()?.map(|x| x.join("polkadot"));
-
 			runner.run_full_node(
 				|config| {
+					polkadot_cli.base_path =
+						config.base_path.as_ref().map(|x| x.path().join("polkadot"));
+
```
cecton, line 216: The base_path had to move here because the temporary directory is created when the configuration is created.

PR #138, opened 2020-07-06
Reorganize/Move Docker scripts

cecton:
I want to do this but I can test as Docker has some issues on my machine...


cecton:
But it's not super important either so I guess it could stay the way it is now

bkchr:
Did you checked this?

cecton:
No and I can't fix my Docker easily... I will do it at some point. I believe the changes I did are correct but it would be better to actually test

cecton:
Ok I will rebase and merge as-it. If it breaks someone will hate me but I don't think many people use it except you atm...

Issue #155, opened 2020-07-23
Fix `storage_append`


bkchr:
`storage_append` needs to check the witness data if a value already exists, currently we only check the overlay.

PR #158, opened 2020-07-26
Fix storage_append

bkchr:
Fixes: https://github.com/paritytech/cumulus/issues/155


bkchr:
> This is the fix for the parachain wasm execution you mentioned before?

Sadly not :( I found this while trying to find the other issue. I'm still convinced that the other issue is a compiler mis-compilation :(

PR #161, opened 2020-07-27
Collate only if `--validator` is given

### LINKED ISSUE #152: Only collate when `--validator` is given as CLI argument
We should only collate when the user passed `--validator` to the node as CLI argument.

cecton:
Fixes #152

### DIFF
--- a/test/parachain/src/service.rs
+++ b/test/parachain/src/service.rs
@@ -88,6 +88,7 @@ pub fn run_collator(
 	key: Arc<CollatorPair>,
 	mut polkadot_config: polkadot_collator::Configuration,
 	id: polkadot_primitives::parachain::Id,
+	validator: bool,
 ) -> sc_service::error::Result<ServiceComponents<
 	parachain_runtime::opaque::Block,
 	TFullBackend<parachain_runtime::opaque::Block>,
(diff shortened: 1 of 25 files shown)


.rustfmt.toml
```diff
@@ -1 +1,2 @@
 hard_tabs=true
+merge_imports=true
```
cecton, line 2: Ah so rustfmt does have something to merge imports? cool!

(I didn't add that line. Well.. I can't remember I did it)

bkchr: I added this in the rococo branch.

Why does this contains rococo changes?

cecton: Because we also need to get rococo changes on cumulus branch right?

rphmeier:
Nice, so this enables regular full nodes?

cecton:
Yes!! ^_^ if I can make the ci pass...

cecton:
Actually it also does light node if asked to do so. I will remove that because I think Basti asked me to do only full node for now

cecton:
Right!! Thanks

test/parachain/src/service.rs
```diff
@@ -87,84 +105,126 @@ pub fn run_collator(
...
-		client,
-		announce_block,
-		block_announce_validator,
-	);
-
-	polkadot_config.informant_output_format = OutputFormat {
-		enable_color: true,
-		prefix: format!("[{}] ", Color::Blue.bold().paint("Relaychain")),
+	let finality_proof_provider =
+		GrandpaFinalityProofProvider::new_for_service(backend.clone(), client.clone());
```
bkchr, line 130: This can be removed

test/parachain/src/service.rs
```diff
@@ -87,84 +105,126 @@ pub fn run_collator(
...
 
-	Ok(service)
+	let _rpc_handlers = sc_service::spawn_tasks(sc_service::SpawnTasksParams {
+			on_demand: None,
+			remote_blockchain: None,
+			rpc_extensions_builder: Box::new(|_| ()),
+			client: client.clone(),
+			transaction_pool: transaction_pool.clone(),
+			task_manager: &mut task_manager,
+			telemetry_connection_sinks: sc_service::TelemetryConnectionSinks::default(),
```
bkchr, line 160:
```suggestion
			telemetry_connection_sinks: Default::default(),
```

cecton: done

test/parachain/src/service.rs
cecton: done

PR #173, opened 2020-08-04
Rename cumulus-test-parachain-collator to rococo-collator

danforbes:
Ran `find . -type f -name "*" -print0 | xargs -0 sed -i -e "s/cumulus-test-parachain-collator/rococo-collator/g"`


danforbes:
I `cargo check`ed and `test`ed this let me know if there's anything else I need to do to make sure this is ready for primetime.

PR #177, opened 2020-08-05
Don't validate a block announcement when syncing

### LINKED ISSUE #150: Don't validate a block announcement when syncing
When we sync the Parachain/Relay chain, we should not validate block announcements, because it leads to the following output:

```
2020-07-16 21:21:48 💔 Block announcement validation errored: UnknownBlock: 0x06b0…78b4
2020-07-16 21:21:48 💔 Block announcement validation errored: UnknownBlock: 0x06b0…78b4
```

This happens because the relay chain block is not yet known to the node.

cecton:
Fixes #150

### DIFF
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -23,7 +23,7 @@ mod tests;
 
 use sp_api::ProvideRuntimeApi;
 use sp_blockchain::{Error as ClientError, HeaderBackend};
-use sp_consensus::block_validation::{BlockAnnounceValidator, Validation};
+use sp_consensus::{block_validation::{BlockAnnounceValidator, Validation}, SyncOracle};
 use sp_core::traits::SpawnNamed;
 use sp_runtime::{
 	generic::BlockId,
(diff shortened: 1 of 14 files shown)


network/src/lib.rs
```diff
@@ -114,6 +124,10 @@ where
 			});
 		}
 
+		if (self.is_major_syncing)() {
+			return Ok(Validation::Success { is_new_best: true });
+		}
```
bkchr, line 129:
```suggestion
		if (self.is_major_syncing)() {
			return Ok(Validation::Success { is_new_best: false });
		}
```

And please move it above the other if and remove the new added inner check.

network/src/lib.rs
```diff
@@ -55,14 +55,20 @@ pub struct JustifiedBlockAnnounceValidator<B, P> {
 	phantom: PhantomData<B>,
 	polkadot_client: Arc<P>,
 	para_id: ParaId,
+	is_major_syncing: Box<dyn FnMut() -> bool + Send + Sync>,
```
bkchr, line 58: Why is that only a FnMut? Why not use the `SyncOracle`?

And please name it:

`polkadot_sync_oracle`.

cecton: :thinking: fair point, you're totally right

cecton: Can you check again? 6dec62e

cecton: I think it is possible to not use Box if I expect the polkadot's "Service" (the network) instead of a SyncOracle

bkchr: LGTM

PR #179, opened 2020-08-05
Move `pub storage ParachainId` parameter type to the pallet

JoshOrndorff:
This PR moves the declaration of `ParachainId` into the `message_example` pallet. This is a stylistic improvement because the struct is defined in the same crate where it is modified. It is also a practical improvement because the pallet can now be moved to its own crate more easily.

### DIFF
--- a/test/parachain/runtime/src/lib.rs
+++ b/test/parachain/runtime/src/lib.rs
@@ -36,6 +36,7 @@ use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 
 mod message_example;
+use message_example::ParachainId;
 
 // A few exports that help ease life for downstream crates.
 pub use frame_support::{
@@ -241,10 +242,6 @@ impl pallet_sudo::Trait for Runtime {
 impl cumulus_parachain_upgrade::Trait for Runtime {
 	type Event = Event;
 	type OnValidationFunctionParams = ();
-}
-
-parameter_types! {
-	pub storage ParachainId: cumulus_primitives::ParaId = 100.into();
 }
 
 impl cumulus_message_broker::Trait for Runtime {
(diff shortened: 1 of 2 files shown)


bkchr:
This should not be moved outside into its own pallet. All this stuff is just to get it running.

bkchr:
So yeah and the parameter type also belongs into the runtime.

riusricardo:
> This should not be moved outside into its own pallet. All this stuff is just to get it running. 

Why? It is already a pallet and can be moved into a crate. Unless you are talking about not moving it because it restricts the customization hack.

> So yeah and the parameter type also belongs into the runtime. 

Makes sense to keep the parameter at runtime level because it doesn't belong to one specific pallet. The issue we got into, is that it can't be customized if it is moved to a pallet crate.

bkchr:
> Why? It is already a pallet and can be moved into a crate. Unless you are talking about not moving it because it restricts the customization hack.

Hey it can be moved into a pallet, but I don't want to see this hack in a pallet :D

JoshOrndorff:
@bkchr Maybe you can help me understand the architecture a little better. The context for this change is that @riusricardo and I are working on a collator template that anyone can use to build a parachain similar to the node template.

So imagine I'm developing a parachain that has three of my own custom pallets. Each of them want to receive downward messages. This is slightly different than the `rococo-collator` which only has a single pallet that will receive downward messages. In this scenario do you envision one of the pallets being "special" in the sense that it contains [this code](https://github.com/paritytech/cumulus/blob/3f3beea7f0190145332c7033fef3eaa28db7f65d/test/parachain/runtime/src/message_example.rs#L61-L68) to initialize the `ParachainId` in storage?

I don't think it is great design to have a single pallet responsible for that. I also now understand why it shouldn't be the job of an individual pallet. What do you think about moving that `decl_storage!` block into the runtime's `lib.rs`? Is that a better solution?

JoshOrndorff:
Maybe a runtime interface is a better solution for getting the parachain id into the runtime?

PR #185, opened 2020-08-07
Graceful shutdown relay chain

### LINKED ISSUE #164: The relay chain doesn't have a clean shutdown
~~When ran without --validator, a full node is started,~~ _For the relay chain,_ a new task manager is created but `graceful_shutdown` is never called on that task manager.

We could implement something in substrate that would allow adding a sub-task manager to a task manager and calls `graceful_shutdown` on every sub-task manager.

_[Update: this issue also exists for --validator]_

cecton:
Fixes #164

### DIFF
--- a/test/parachain/src/service.rs
+++ b/test/parachain/src/service.rs
@@ -17,7 +17,6 @@
 use ansi_term::Color;
 use cumulus_collator::{prepare_collator_config, CollatorBuilder};
 use cumulus_network::{DelayedBlockAnnounceValidator, JustifiedBlockAnnounceValidator};
-use futures::{future::ready, FutureExt};
 use polkadot_primitives::v0::{CollatorPair, Block as PBlock, Id as ParaId};
 use sc_executor::native_executor_instance;
 pub use sc_executor::NativeExecutor;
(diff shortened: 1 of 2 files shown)


test/parachain/src/service.rs
```diff
@@ -191,21 +190,17 @@ pub fn run_collator(
...
-		// Make sure the polkadot task manager survives as long as the service.
-		let polkadot_future = polkadot_future.then(move |_| {
-			let _ = polkadpt_task_manager;
-			ready(())
-		});
-
 		task_manager
 			.spawn_essential_handle()
-			.spawn("polkadot", polkadot_future);
+			.spawn("polkadot-collator", polkadot_future);
```
bkchr, line 198:
```suggestion
			.spawn("polkadot", polkadot_future);
```

This is the future of the Polkadot service.

bkchr:
bot merge

test/parachain/src/service.rs
cecton: hmmm no I think it's the collator future: https://github.com/paritytech/polkadot/blob/master/collator/src/lib.rs#L263

It's a big loop that runs "collation work" when there are new notifications.

What I would call the "polkadot service" is the task manager that is passed as child right after this.

cecton: I guess my vocabulary is still a big wrong :sweat:

PR #186, opened 2020-08-07
Proof of concept: Add a new runtime that uses pallet_contracts

athei:
This adds pallet contracts to a new runtime called `contracts-runtime`. The pallet is copied over rather than added as dependency because some changes needed to be made in order to make it compatible to the current parachain storage implementation. 

There are two features missing when the parachain validation function is executed which are both used by the upstream contracts pallet:
1. Storage Transactions
2. Child Tries

The first issue is addressed by using an older (pre rc4) of the pallet that did not make use of storage transaction. On top of that older version some backports of newer changes are applied in order to make it compatible to ink!2.1 and the current polkadot js UI. The latter issue is fixed by using a prefix instead of a child trie for contract storage.

## Caveats

There are some caveats but that should not matter for a proof of concept that is only for testing purposes. Review accordingly.

### Restoration is non-functional

This is because of missing child trie support. Trying to restore a tombstone will always yield an error:
https://github.com/paritytech/cumulus/blob/f0748deb22d4d70426006ab2ec7f58e2d3b20c7e/test/parachain/runtime/pallet_contracts/src/lib.rs#L708-L717

### More allocation

The prefix creation is implemented in a naive way and induces additional allocations:
https://github.com/paritytech/cumulus/blob/f0748deb22d4d70426006ab2ec7f58e2d3b20c7e/test/parachain/runtime/pallet_contracts/src/lib.rs#L225-L233

### Runtime is not used by the client, yet

The current client is not equipped for being a multi runtime client. Some further refactoring is needed in order to actually **use** the new runtime with the `rococo-collator`. The two references to contract specific stuff in the genesis and rpc configuration are commented out with a `TODO` in front of them.


athei:
I will re-organize the directory layout before merging. Also the contracts pallet will be put into its own runtime rather than being added to the existing one.

PR #189, opened 2020-08-09
Rework default values used by the RelayChainCli

bkchr:
This reworks the default values used by the RelayChainCli for stuff like
the listen port etc.

This also renames all the contracts related stuff to `cumulus-*` to
support `.cargo/config` overrides.

### DIFF
--- a/rococo-parachains/src/command.rs
+++ b/rococo-parachains/src/command.rs
@@ -24,8 +24,8 @@ use log::info;
 use parachain_runtime::Block;
 use polkadot_parachain::primitives::AccountIdConversion;
 use sc_cli::{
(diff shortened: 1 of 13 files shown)


rococo-parachains/src/command.rs
```diff
@@ -269,7 +269,25 @@ pub fn run() -> Result<()> {
 	}
 }
 
-impl CliConfiguration for RelayChainCli {
+impl DefaultConfigurationValues for RelayChainCli {
```
JoshOrndorff, line 272:
```suggestion
// Relay chain default values are one higher than parachain and Substrate default values.
impl DefaultConfigurationValues for RelayChainCli {
```

PR #201, opened 2020-08-13
Fix integration test again

cecton:


rococo-parachains/src/service.rs
```diff
@@ -235,9 +236,42 @@ where
...
+
+			(future, task_manager)
+		} else {
+			polkadot_collator::start_collator(builder, id, key, polkadot_config)?
+		};
+
+		task_manager
+			.spawn_essential_handle()
+			.spawn("polkadot", polkadot_future);
+*/
```
cecton, line 270:
```suggestion
```

cecton:
CI is red but but... it works on my machine :sob: 

image

cecton:
@bkchr @ddorgan @TriplEight FINALLY FIXED IT!! :partying_face: 

The difference between CI and locally was only because I ran the test from a terminal so the STDIN is kept opened. In the CI, the STDIN is closed.

For someone who made [a library that makes pipelines with processes](https://github.com/cecton/destream) it is infuriating that I didn't see that earlier :disappointed:

service/Cargo.toml
```diff
@@ -25,3 +25,7 @@ sp-blockchain = { git = "https://github.com/paritytech/substrate", branch = "roc
 polkadot-collator = { git = "https://github.com/paritytech/polkadot", branch = "rococo-branch" }
 polkadot-primitives = { git = "https://github.com/paritytech/polkadot", branch = "rococo-branch" }
 polkadot-service = { git = "https://github.com/paritytech/polkadot", branch = "rococo-branch" }
+# TODO should be dev dependency
+polkadot-test-service = { git = "https://github.com/paritytech/polkadot", branch = "rococo-branch" }
```
cecton, line 29: :disappointed:

service/src/lib.rs
```diff
@@ -99,8 +101,30 @@ where
 		block_announce_validator,
 	);
 
-	let (polkadot_future, polkadot_task_manager) =
-		polkadot_collator::start_collator(builder, para_id, collator_key, polkadot_config)?;
+	let (polkadot_future, polkadot_task_manager) = if test {
+		let (task_manager, client, handles, _network, _rpc_handlers) = polkadot_test_service::polkadot_test_new_full(
```
bkchr, line 105: This should not be done here in this crate.

In the future I also don't want to use the rococo-collator for testing. We need a collator that is just used for testing. We really need to rework all this.

So, please move this functionality for now into the `rococo-collator`.

bkchr (approved):
One nitpick, otherwise it looks good.

cecton:
bot merge

PR #204, opened 2020-08-20
feature flag relaychain runtimes

### LINKED ISSUE #175: Building the Collator builds all the Polkadot runtimes.
When building the test collator with `cargo build --release -p rococo-validator`, all polkadot runtimes are built.  For example it builds `kusama-runtime`, `westend-runtime`, `polkadot-test-runtime`, etc.

This is harmless in terms of correctness, but it adds a lot of build time.

xlc:
Closes #175

The compiling time of building 5 runtimes (4 relaychain + 1 parachain, maybe twice wasm & native?) is literally killing my life.

Also relay chain runtime std feature is currently not enabled even in std because cargo. The dependent package can however depends on runtime directly and enable it.

### DIFF
--- a/upward-message/Cargo.toml
+++ b/upward-message/Cargo.toml
@@ -13,30 +13,31 @@ polkadot-core-primitives = { git = "https://github.com/paritytech/polkadot", bra
 polkadot-parachain = { git = "https://github.com/paritytech/polkadot", branch = "rococo-branch", default-features = false }
 
 # All these should be optional dependenices, but given the perfect Cargo, it is not possible.
-polkadot-runtime = { git = "https://github.com/paritytech/polkadot", branch = "rococo-branch", default-features = false }
-kusama-runtime = { git = "https://github.com/paritytech/polkadot", branch = "rococo-branch", default-features = false }
-westend-runtime = { git = "https://github.com/paritytech/polkadot", branch = "rococo-branch", default-features = false }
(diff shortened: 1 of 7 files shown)


xlc:
@bkchr does it currently builds wasm file as well? Is there a way to disable wasm building?

upward-message/Cargo.toml
```diff
@@ -13,30 +13,27 @@ polkadot-core-primitives = { git = "https://github.com/paritytech/polkadot", bra
...
 std = [
 	"sp-std/std",
-	"polkadot-runtime/std",
-	"kusama-runtime/std",
-	"westend-runtime/std",
-	"rococo-runtime/std",
 	"polkadot-core-primitives/std",
 	"polkadot-parachain/std",
 ]
-# Should be enabled by when compiling the runtime for WASM.
```
bkchr, line 32: Please add back.

xlc:
The feature handling with cargo makes thing bit more complicated than what I was expecting... Does it make sense for me to continue on this PR or just close it because it may not actually solve my problem?

bkchr:
> The feature handling with cargo makes thing bit more complicated than what I was expecting... Does it make sense for me to continue on this PR or just close it because it may not actually solve my problem?

As you like, but to solve this correctly we will also need to make the runtimes optional to the polkadot-service.

xlc:
Ok. Since this doesn't look like a low hanging fruit anymore, I will close this.

PR #205, opened 2020-08-24
Fail when storage is not in proof

cheme:
This PR adds a new function to storage in order to check for change of the current block processing without querying the trie backend.
This way, this PR allows trie backend to fail on incomplete proofs.

### DIFF
--- a/runtime/src/validate_block/implementation.rs
+++ b/runtime/src/validate_block/implementation.rs
@@ -261,20 +261,14 @@ impl<B: BlockT> Storage for WitnessStorage<B> {
 				.overlay
 				.get(key)
 				.cloned()
-				.or_else(|| {
+				.unwrap_or_else(|| {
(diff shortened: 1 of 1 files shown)


runtime/src/validate_block/implementation.rs
```diff
@@ -253,6 +267,11 @@ impl<B: BlockT> Storage for WitnessStorage<B> {
 						&self.storage_root,
 						key,
 					)
+					.map_err(|e| match *e {
+						TrieError::DecoderError(..) => panic!("Invalid encoding in proof"),
+						TrieError::IncompleteDatabase(..) => panic!("Incomplete proof"),
+						_ => panic!("Invalid proof"),
+					})
 					.ok()
```
bkchr, line 275:
```suggestion
					.expect("Reading key from trie.")
```

bkchr (approved):

PR #214, opened 2020-09-11
Use state machine externalities in validation runtime.

cheme:
@bkchr, I am not sure how to push this pr, it is using specific polkadot branch (no change from rococo branch except
using the substrate one) and specific substrate branch (rococo branch with cherry-picking the no_std pr).

So first step would be to merge substrate rococo branch with mine this PR shall not need the toml repo changes and then could be mergeable.

Only relevant changes are in runtime/src/validate_block/implementation.rs  and runtime/Cargo.toml. 

Also this time (it passes with previous version), I did not ran all test, probably need #204 for that.

I also did write the boilerplate for the other ext functions, but at this point ran no test on it.


bkchr:
Sorry for taking to so long.

After https://github.com/paritytech/cumulus/pull/224 is merged, cumulus is compatible with substrate master and we will be able to merge your pr :)

cheme:
Np, I did realize (from riot) that the test issues were not just a matter of merging #204, I will update when #224 lands.

bkchr:
@cheme Please update to master

cheme:
Updated, also not too sure about my comment on #217 (issue description seems a bit outdated, in my merge we just check its values on update, but maybe the update occurs every time so it is a bit similar).
Also maybe the check_validation_data method could be call afterward (after execute_block if there is a change in the overlay), that will be a bit less optimal, but remove some code.

runtime/src/validate_block/implementation.rs
```diff
@@ -369,38 +382,151 @@ fn host_storage_read(key: &[u8], value_out: &mut [u8], value_offset: u32) -> Opt
...
+		.expect("Runing with a correct environment")
+}
+
+fn host_default_child_storage_exists(
+	storage_key: &[u8],
+	key: &[u8],
+) -> bool {
+	let child_info = ChildInfo::new_default(storage_key);
+	with_externalities(|ext| ext.exists_child_storage(&child_info, key))
+		.expect("Runing with a correct environment")
```
bkchr, line 505: Can we please put these `with_externalities` calls into a function that just panics directly. (so just a wrapper around it)

And please use a better expect message :D

bkchr (approved):
One nitpick and please merge master one more time.

Ty for the work! :)

bkchr:
> Updated, also not too sure about my comment on #217 (issue description seems a bit outdated, in my merge we just check its values on update, but maybe the update occurs every time so it is a bit similar).
> Also maybe the check_validation_data method could be call afterward (after execute_block if there is a change in the overlay), that will be a bit less optimal, but remove some code.

We can tackle this code duplication later as part of #217.

runtime/src/validate_block/implementation.rs
cheme: Done, not sure my new message is a lot better (the two new ones are copied from substrate).

PR #220, opened 2020-09-16
Uniformize tests

cecton:
WIP

- [x] copy rococo parachain runtime to make test runtime
- [x] move and adapt integration test that uses rococo parachain runtime
- [x] move and adapt the other tests that use substrate-test-runtime


test/service/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+		+ 'static,
+	RuntimeApi::RuntimeApi: sp_transaction_pool::runtime_api::TaggedTransactionQueue<Block>
+		+ sp_api::Metadata<Block>
+		+ sp_session::SessionKeys<Block>
+		+ sp_api::ApiExt<
+			Block,
+			Error = sp_blockchain::Error,
+			StateBackend = sc_client_api::StateBackendFor<TFullBackend<Block>, Block>,
+		> + sp_offchain::OffchainWorkerApi<Block>
+		+ sp_block_builder::BlockBuilder<Block>,
```
bkchr, line 95: Why is that soo generic?

test/service/tests/integration_test.rs
```diff
@@ -0,0 +1,96 @@
...
+		Bob,
+		|| {},
+		vec![alice.addr.clone()],
+	);
+
+	// ensure alice and bob can produce blocks
+	join!(alice.wait_for_blocks(2), bob.wait_for_blocks(2));
+
+	// call function to register the parachain
+	let function = polkadot_test_runtime::Call::Sudo(pallet_sudo::Call::sudo(Box::new(
```
bkchr, line 44: That should clearly be a default feature of the testing framework. I don't want to repeat this code over and over.

test/service/src/genesis.rs
```diff
@@ -0,0 +1,58 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+use codec::Encode;
+use cumulus_primitives::ParaId;
+use cumulus_test_runtime::Block;
+use polkadot_primitives::v0::HeadData;
+use sp_runtime::traits::{Block as BlockT, Hash as HashT, Header as HeaderT, Zero};
+
+pub fn generate_genesis_state(
```
bkchr, line 23: Put this function into cumulus-primitives. There is no reason copy this.

test/service/src/chain_spec.rs
```diff
@@ -0,0 +1,131 @@
+use cumulus_primitives::ParaId;
```
bkchr, line 1: No license header.

test/runtime/src/message_example.rs
```diff
@@ -0,0 +1,179 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
bkchr, line 1: Please remove this. We don't need it

test/primitives/src/lib.rs
```diff
@@ -0,0 +1,59 @@
+// Copyright 2020 Parity Technologies (UK) Ltd.
```
bkchr, line 1: WHy do we need this?

runtime/src/validate_block/tests.rs
```diff
@@ -74,53 +74,38 @@ fn call_validate_block(
...
-	extrinsics: Vec<<Block as BlockT>::Extrinsic>,
-) -> (Block, sp_trie::StorageProof) {
+fn build_block_with_proof(client: &Client) -> (Block, sp_trie::StorageProof) {
 	let block_id = BlockId::Hash(client.info().best_hash);
 	let mut builder = client
 		.new_block_at(&block_id, Default::default(), true)
 		.expect("Initializes new block");
 
-	extrinsics
+	let mut inherent_data = sp_consensus::InherentData::new();
```
bkchr, line 90: It would be nice to have a trait that extends `BlockBuilder` to not always have to repeat this code.

runtime/src/validate_block/tests.rs
```diff
@@ -135,23 +120,10 @@ fn build_block_with_proof(
...
-
-	let block_data = ParachainBlockData::new(header.clone(), extrinsics, witness_data);
-
-	let res_header = call_validate_block(parent_head, block_data).expect("Calls `validate_block`");
-	assert_eq!(header, res_header);
-}
-
-#[test]
-fn validate_block_with_extrinsics() {
+fn validate_block_no_extrinsics() {
```
bkchr, line 123: That is not true.

And please also include some trivial balance transfers. That should also be part of the test framework.

bkchr (commented):
Goes already into a good direction, however I want MOARE:

- It should be similar to the polkadot test service/runtime
- Having functions for doing transfers, easy setup, no that complicated inherent setup
- Easy, easy easy :P 

I just want to be able to happily write my tests and don't want to care about all the other shit, when I just want to test some small thing.

test/service/src/lib.rs
cecton: OK
fire

test/service/tests/integration_test.rs
cecton: https://github.com/paritytech/polkadot/pull/1733

test/service/src/chain_spec.rs
cecton: fixed

test/service/src/genesis.rs
cecton: fixed in ea792034e8801b67f918301230901688771d3f50

test/service/src/lib.rs
cecton: Mostly fixed in d2e8715e8add209c4cae9ea30d80a40ebcee97a7

test/runtime/src/message_example.rs
cecton: done in edac257

test/primitives/src/lib.rs
cecton: We need Block, Hash, AccountId, Signature of the test runtime... Maybe I can move it to the test-runtime crate directly?

runtime/src/validate_block/tests.rs
cecton: I tried but I'm struggling with the generics https://github.com/paritytech/cumulus/pull/220/commits/ea5b9e3b31410ec18ad9626b53835ad5d0d0bb60

Logs: https://gitlab.parity.io/parity/cumulus/-/jobs/653559

I think I'm close though!

test/service/tests/integration_test.rs
cecton: done in ce34da6

runtime/src/validate_block/tests.rs
cecton: I give up on that one.

test/primitives/src/lib.rs
cecton: done in 78338b3

runtime/src/validate_block/tests.rs
cecton: done in f6581e7

runtime/src/validate_block/tests.rs
```diff
@@ -74,53 +75,28 @@ fn call_validate_block(
...
 	let block_id = BlockId::Hash(client.info().best_hash);
 	let mut builder = client
 		.new_block_at(&block_id, Default::default(), true)
 		.expect("Initializes new block");
 
-	extrinsics
+	builder
+		.cumulus_inherents(client)
+		.into_iter()
+		.for_each(|e| builder.push(e).expect("Pushes an extrinsic"));
```
bkchr, line 97:
```suggestion
		.for_each(|e| builder.push(e).expect("Pushes an inherent"));
```

test/client/src/block_builder.rs
```diff
@@ -0,0 +1,63 @@
...
+		let timestamp = runtime::MinimumPeriod::get();
+
+		inherent_data
+			.put_data(sp_timestamp::INHERENT_IDENTIFIER, &timestamp)
+			.expect("Put timestamp failed");
+		inherent_data
+			.put_data(
+				cumulus_primitives::inherents::VALIDATION_FUNCTION_PARAMS_IDENTIFIER,
+				&cumulus_primitives::validation_function_params::ValidationFunctionParams::default(
+				),
```
bkchr, line 50:
```suggestion
				&cumulus_primitives::validation_function_params::ValidationFunctionParams::default(),
```

test/client/src/block_builder.rs
```diff
@@ -0,0 +1,63 @@
...
+	A::Api: BlockBuilderApi<Block, Error = Error>
+		+ ApiExt<Block, StateBackend = backend::StateBackendFor<B, Block>>,
+	B: backend::Backend<Block>,
+{
+	fn cumulus_inherents(&mut self, client: &Client) -> Vec<runtime::UncheckedExtrinsic> {
+		let mut inherent_data = sp_consensus::InherentData::new();
+		let timestamp = runtime::MinimumPeriod::get();
+
+		inherent_data
+			.put_data(sp_timestamp::INHERENT_IDENTIFIER, &timestamp)
```
bkchr, line 44: The moment you want to build a second block, this will fail. 

For the test runtime we could just introduce a new runtime api that can fetch the latest timestamp. Than you would fetch the latest timestamp and add the minimum period.

test/client/src/block_builder.rs
```diff
@@ -0,0 +1,63 @@
...
+use sc_block_builder::{BlockBuilder, BlockBuilderApi};
+use sc_client_api::backend;
+use sp_api::{ApiExt, BlockT, ProvideRuntimeApi};
+use sp_blockchain::Error;
+use sp_core::ExecutionContext;
+use sp_runtime::generic::BlockId;
+
+/// Push the inherents to a block so you don't have to.
+pub trait PushInherents {
+	/// Push the inherents to a block so you don't have to.
```
bkchr, line 27: That is not correct.

You just generate the inherents in this function. But, you could also change it in the way as you describe it in the comment.

test/client/src/block_builder.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::Client;
+use sc_block_builder::{BlockBuilder, BlockBuilderApi};
+use sc_client_api::backend;
+use sp_api::{ApiExt, BlockT, ProvideRuntimeApi};
+use sp_blockchain::Error;
+use sp_core::ExecutionContext;
+use sp_runtime::generic::BlockId;
+
+/// Push the inherents to a block so you don't have to.
+pub trait PushInherents {
```
bkchr, line 26: Also not true and for "one function" traits I often like to have the trait and the function the same name (while the function is in snake_case)

test/client/src/lib.rs
```diff
@@ -128,20 +136,59 @@ impl DefaultTestClientBuilderExt for TestClientBuilder {
...
-			AccountKeyring::Bob.into(),
-			AccountKeyring::Charlie.into(),
-		],
-		1000,
-		Default::default(),
-		Default::default(),
+	cumulus_test_service::local_testnet_genesis(changes_trie_config)
+}
+
+fn additional_storage_with_genesis(genesis_block: &Block) -> BTreeMap<Vec<u8>, Vec<u8>> {
```
bkchr, line 142: I don't think that we need this entire function?

test/client/src/lib.rs
```diff
@@ -128,20 +136,59 @@ impl DefaultTestClientBuilderExt for TestClientBuilder {
...
+		),
+	);
+	let signature = raw_payload.using_encoded(|e| origin.sign(e));
+
+	UncheckedExtrinsic::new_signed(
+		function.clone(),
+		origin.public().into(),
+		Signature::Sr25519(signature.clone()),
+		extra.clone(),
 	)
```
bkchr, line 193: Could you put this into a generic function? I like the `transfer` function and it could call this new function. However, we can later use this for other transactions.

test/service/tests/integration_test.rs
```diff
@@ -0,0 +1,90 @@
...
+
+use cumulus_primitives::ParaId;
+use cumulus_test_service::initial_head_data;
+use futures::join;
+use sc_service::TaskExecutor;
+use substrate_test_runtime_client::AccountKeyring::*;
+
+#[substrate_test_utils::test]
+#[ignore]
+async fn integration_test(task_executor: TaskExecutor) {
```
bkchr, line 25: Maybe a more expressive name what this is doing? :D

test/service/src/chain_spec.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/// Get the chain spec for a specific parachain ID.
+pub fn get_chain_spec(id: ParaId) -> ChainSpec {
+	ChainSpec::from_genesis(
+		"Local Testnet",
+		"local_testnet",
+		ChainType::Local,
+		move || {
+			testnet_genesis(
```
bkchr, line 71: Why isn't that here just calling `local_testnet_genesis`?

test/runtime/src/lib.rs
```diff
@@ -1,27 +1,354 @@
...
+		NodeBlock = NodeBlock,
+		UncheckedExtrinsic = UncheckedExtrinsic,
+	{
+		System: frame_system::{Module, Call, Storage, Config, Event<T>},
+		Timestamp: pallet_timestamp::{Module, Call, Storage, Inherent},
+		Balances: pallet_balances::{Module, Call, Storage, Config<T>, Event<T>},
+		Sudo: pallet_sudo::{Module, Call, Storage, Config<T>, Event<T>},
+		RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Module, Call, Storage},
+		ParachainUpgrade: cumulus_parachain_upgrade::{Module, Call, Storage, Inherent, Event},
+		MessageBroker: cumulus_message_broker::{Module, Call, Inherent, Event<T>},
```
bkchr, line 226: Can be removed

test/runtime/src/lib.rs
```diff
@@ -1,27 +1,354 @@
...
+pub const DAYS: BlockNumber = HOURS * 24;
+
+// 1 in 4 blocks (on average, not counting collisions) will be primary babe blocks.
+pub const PRIMARY_PROBABILITY: (u64, u64) = (1, 4);
+
+#[derive(codec::Encode, codec::Decode)]
+pub enum XCMPMessage<XAccountId, XBalance> {
+	/// Transfer tokens to the given account from the Parachain account.
+	TransferToken(XAccountId, XBalance),
+}
```
bkchr, line 86: Can be removed

bkchr (approved):
Looks good in general, some last things that need to be done

Nice :+1:

test/client/src/block_builder.rs
cecton: Totally agree!

test/client/src/block_builder.rs
cecton: Yes but I couldn't :sweat: I might give it another try (3rd) but something is complicated with the generics.

If I can make it work. I will call it PushInherents / push_inherents and keep it as a trait.

If I can't: I will remove the trait, just keep the function (or make the trait on the Client maybe) because I don't use `&mut self`

test/client/src/lib.rs
cecton: Yes we do in line ~95

test/client/src/block_builder.rs
cecton: Done in 8fe7122a2fec7f9d1bd1d63ed98d59831f87b33d

test/client/src/lib.rs
cecton: Done in d00c4db83a9d8ea55955d045f13e1c5f4a2d8082

test/service/tests/integration_test.rs
cecton: Done in 56722dbcaf8c23c1bba7954292ca30b76a2c1a4e

"test_collating_and_non_collator_mode_catching_up"

test/client/src/block_builder.rs
cecton: Done in 5d89280

cecton:
bot merge

PR #222, opened 2020-10-01
Add --collator flag as alternative to --validator 

### LINKED ISSUE #193: Add a --collator flag
The term "validator" is reserved for relay chain validators in the polkadot ecosystem, so it is confusing to pass the `--validator` flag to a collator node.

It is easy enough to add a `--collator` flag which can be documented and used in tutorials. Whether we can or should remove the `--validator` flag is open for discussion.

cecton:
Closes #193

### DIFF
--- a/rococo-parachains/src/cli.rs
+++ b/rococo-parachains/src/cli.rs
@@ -93,8 +93,10 @@ pub struct Cli {
 	#[structopt(flatten)]
 	pub run: RunCmd,
 
-	/// Run a collator node.
-	#[structopt(long)]
+	/// Run node as collator.
+	///
+	/// Note that this is the same as running with `--validator`.
+	#[structopt(long, conflicts_with = "validator")]
 	pub collator: bool,
 
 	/// Relaychain arguments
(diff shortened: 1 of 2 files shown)


rococo-parachains/src/cli.rs
```diff
@@ -93,6 +93,10 @@ pub struct Cli {
 	#[structopt(flatten)]
 	pub run: RunCmd,
 
+	/// Run a collator node.
```
bkchr, line 96:
```suggestion
	/// Run node as collator.
	///
	/// Note that this is the same as running with `--validator`.
```

rococo-parachains/src/cli.rs
```diff
@@ -93,6 +93,10 @@ pub struct Cli {
...
+	#[structopt(long)]
```
bkchr, line 97:
```suggestion
	#[structopt(long, conflicts_with = "validator")]
```

bkchr (approved):

cecton:
bot merge

PR #224, opened 2020-10-13
Update Cumulus for Parachains V1

bkchr:


collator/src/lib.rs
```diff
@@ -215,197 +210,204 @@ where
...
+		let block_data = BlockData(block.encode());
+		let header = block.into_header();
+		let head_data = HeadData(header.encode());
+
+		let state = match self.backend.state_at(BlockId::Hash(block_hash)) {
+			Ok(state) => state,
 			Err(e) => {
-				error!(target: "cumulus-collator", "Could not decode the head data: {:?}", e);
-				return Box::pin(future::ready(None));
+				error!(target: "cumulus-collator", "Failed to get state of the freshly build block: {:?}", e);
```
rphmeier, line 226:
```suggestion
				error!(target: "cumulus-collator", "Failed to get state of the freshly built block: {:?}", e);
```

collator/src/lib.rs
```diff
@@ -215,197 +210,204 @@ where
...
-				None => {
-					error!(
-						target: "cumulus-collator",
-						"Proposer did not return the requested proof.",
-					);
+			let new_validation_code = sp_io::storage::get(well_known_keys::NEW_VALIDATION_CODE);
 
+			Some(Collation {
+				//TODO: What should we do here?
+				fees: 0,
```
rphmeier, line 246: The fees member can just be removed. We don't have anything that fees are charged for anymore, now that we use channels for messages.

collator/src/lib.rs
```diff
@@ -215,197 +210,204 @@ where
...
+			.ok()?;
 
-			if let Err(err) = block_import
-				.lock()
-				.import_block(block_import_params, Default::default())
-			{
+		let inherent_data = self.inherent_data(
+			&validation_data,
+			// TODO get the downward messages
+			Vec::new(),
```
rphmeier, line 291: what was the blocker here so far? Is it not passed in to the collation pipeline at the moment?

collator/src/lib.rs
```diff
@@ -215,197 +210,204 @@ where
...
+		let Proposal {
+			block,
+			storage_changes,
+			proof,
+		} = proposer
+			.propose(
+				inherent_data,
+				Default::default(),
+				//TODO: Fix this.
+				Duration::from_millis(500),
```
rphmeier, line 303: what would this fix look like?

collator/src/lib.rs
pepyakin: Oh, should it? I assumed the collator would call the runtime API for fetching those

parachain-upgrade/src/lib.rs
```diff
@@ -103,52 +106,60 @@ decl_module! {
...
 					Self::put_parachain_code(&validation_function);
-					Self::deposit_event(Event::ValidationFunctionApplied(vfp.relay_chain_height));
+					Self::deposit_event(Event::ValidationFunctionApplied(vfp.persisted.block_number));
 				}
 			}
 
-			storage::unhashed::put(VALIDATION_FUNCTION_PARAMS, &vfp);
-			DidUpdateVFPs::put(true);
-			<T::OnValidationFunctionParams as OnValidationFunctionParams>::on_validation_function_params(vfp);
+			storage::unhashed::put(VALIDATION_DATA, &vfp);
```
rphmeier, line 124: Can you explain the new relationship between `ValidationData` and `ValidationParams` and for which situations each is relevant?

collator/src/lib.rs
bkchr: This comes from the Polkadot side.

rphmeier (approved):
The code changes look fine to me, but can you explain in a bit more detail the expected inputs to the validation function and how it interfaces with Polkadot?

collator/src/lib.rs
bkchr: Before it was passed to the collation. I think it would be nice to follow this pattern, as we pass `ValidationData` as well.

collator/src/lib.rs
bkchr: This comment is old. However, I think that I will be able to improve this with the switch to the slot interface. In the end I just need a reasonable max time for the block production. However, we will probably stick to 500ms for block production? This gives us 1.5 seconds for propagating it to the parachain validators and validation on their end.

parachain-upgrade/src/lib.rs
bkchr: `ValidationData` is passed using an inherent into the runtime, while `ValidationParams` are passed when validating the parachain.

We use the `ValidationParams` to validate the `ValidationData` while validating the PovBlock.

bkchr:
> The code changes look fine to me, but can you explain in a bit more detail the expected inputs to the validation function and how it interfaces with Polkadot?

Hmm? We take the `ValidationParams` as given by Polkadot. How should I take more data? Or do you mean something else?

And with interfacing with Polkadot, do you mean the client side?

collator/src/lib.rs
pepyakin: Yeah, I also think it would be convenient. But I also think it would potentially be less flexible, because after all it's the collator who knows which messages it should pull, whereas the collation-generation doesn't necessary know about the specifics of the network.

collator/src/lib.rs
pepyakin: Yeah, we can remove it there. It's plumbed everywhere but essentially it starts it's life here. 

https://github.com/paritytech/polkadot/blob/4e5df2988ce13851faf56de4e2f85b68a5601b8c/node/core/candidate-validation/src/lib.rs#L494-L494

so it's just a constant

collator/src/lib.rs
bkchr: Wouldn't it need to fetch all downward messages anyway?

> whereas the collation-generation doesn't necessary know about the specifics of the network.

Can you explain what you mean?

pepyakin: Potentially, it might be many, esp. if you include all the HRMP queues. But not only that, imagine a parachain that doesn't need HRMP, then it can just avoid requesting HRMP data at all. I think that it may be fair for XCMP as well. A conclusion I draw from this, is that the collator knows about the target chain (= "the network") better than the collation-generation can do (unless, we complicate the interface way more). That's why I say that the pull model is more flexible, because the collator has more information on when and how to query this data.

bkchr: Okay fine :)

PR #231, opened 2020-11-09
Integrate DMP into collation

pepyakin:


pepyakin:
Per private conversation, we should move from message-passing to direct calls into runtime.

PR #233, opened 2020-11-10
Integrate DMP into collation, take 2

pepyakin:
Subsumes #231

### DIFF
--- a/collator/src/lib.rs
+++ b/collator/src/lib.rs
@@ -68,6 +68,7 @@ pub struct Collator<Block: BlockT, PF, BI, BS, Backend> {
 	block_status: Arc<BS>,
 	wait_to_announce: Arc<Mutex<WaitToAnnounce<Block>>>,
 	backend: Arc<Backend>,
+	retrieve_dmq_contents: Arc<dyn Fn(PHash) -> Option<DownwardMessagesType> + Send + Sync>,
 }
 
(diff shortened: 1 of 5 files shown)


collator/src/lib.rs
```diff
@@ -68,6 +68,7 @@ pub struct Collator<Block: BlockT, PF, BI, BS, Backend> {
 	block_status: Arc<BS>,
 	wait_to_announce: Arc<Mutex<WaitToAnnounce<Block>>>,
 	backend: Arc<Backend>,
+	retrieve_dmq_contents: Arc<dyn Fn(PHash) -> Option<DownwardMessagesType> + Send + Sync>,
```
bkchr, line 71: Why isn't that just a function of the `Collator`?

What is the benefit of having this as `Arc`?

bkchr (approved):

collator/src/lib.rs
pepyakin: The answer for both is simplicity in terms of implementation: I figured that would be a way cleaner than plumbing all the associated generics for embedding a polkadot client into the `Collator` struct, and the cost seems to be negligible. 

`Arc` was used with a similar reasoning. An alternative is to use `Box<... + Clone>` and that stems from the fact that we need to implement `Clone` for `Collator`.

PR #207, opened 2020-08-29
XCM v0

gavofyork:
Goes along with paritytech/polkadot#1657

PR #238, opened 2020-11-17
Add XCM to Message Broker

shawntabrizi:
This PR adds the XCM pipeline from XCM v0 into the Cumulus Message Broker.

See: https://github.com/paritytech/cumulus/pull/207/


pepyakin (commented):
I don't think this is the correct approach. I'd pretty much prefer to keep the XCM stuff out of the message-broker. 

The XCM logic should be hooked up to the message broker via a `DownwardMessageHandlers` implementation.

PR #251, opened 2020-12-01
Update polkadot to the latest master

pepyakin:
This includes 

- renaming uses of `::Trait` to `::Config`
- renaming `trait Trait` declarations to `Config`
- fixing the fallout after https://github.com/paritytech/substrate/pull/7472 by introducing `BlockAnnounceError`, a box wrapper around a String.
- updating the 'static bound. This is needed for the further refactorings, namely moving the client into the `struct Collator`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -303,7 +303,7 @@ dependencies = [
  "futures-io",
  "rustls",
  "webpki",
- "webpki-roots",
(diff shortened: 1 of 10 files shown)


bkchr (approved):

bkchr:
Ty

PR #252, opened 2020-12-01
Update parachain registration script to new extrinsic

rphmeier:
Tested locally.

Usage: `{url} {wasm_file} {genesis_head} {para_id} {types_file}`

In practice: `scripts/register_parachain.sh ws://127.0.0.1:9944 "//Alice" ~/adder_genesis.wasm $(cat ~/adder_genesis.head) 100 scripts/temp_parachain_types.json`

I removed the balance setting thing as that's a separate concern and it didn't work for some reason.

### DIFF
--- a/scripts/register_parachain.sh
+++ b/scripts/register_parachain.sh
@@ -2,7 +2,7 @@
 
 usage() {
     echo Usage:
-    echo "$0 <url> <seed> <wasm> <genesis> <parachain-id> <tokens> <account>"
+    echo "$0 <url> <seed> <wasm> <genesis> <parachain-id> <types>"
     exit 1
 }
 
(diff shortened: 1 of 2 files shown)


scripts/temp_parachain_types.json
```diff
@@ -0,0 +1,64 @@
+{
```
bkchr, line 1: @jacogr can you upstream these types please.

bkchr (approved):

scripts/temp_parachain_types.json
jacogr: https://github.com/polkadot-js/api/pull/2895

jacogr: And the apps UI API bump - https://github.com/polkadot-js/apps/pull/4136

scripts/register_parachain.sh
```diff
@@ -11,46 +11,40 @@ seed=$2
...
     --sudo \
     --seed "${seed?}" \
-    tx.balances.setBalance \
-        "${account?}" \
-        $((tokens * 10 ** 12)) \
-        0
+    --types "${types?}" \
+    tx.parasSudoWrapper.sudoScheduleParaInitialize \
+        "${parachain_id?}" \
+        "{ \"genesisHead\":\"${genesis?}\", \"validationCode\":\"${wasm?}\", \"parachain\": true }" \
```
JoshOrndorff, line 50: Doesn't this put the entire ~1MB wasm blob on the CLI? Was your shell able to handle that?

rphmeier: Shell should be able to handle 1MB easily. Many terminal emulators might not. Just send output of the script to `/dev/null` if it's a problem.

menduong: I try 
./register.sh ws://127.0.0.1:9944 "//Alice" ~/genesis.wasm $(cat ~/genesis-state) 14 temp_parachain_types.json > /dev/null 2>&1
but nothing happened in my network

Issue #235, opened 2020-11-11
Authorize incoming downward and horizontal messages


pepyakin:
After https://github.com/paritytech/cumulus/pull/233 was landed, the collator submits messages and PVF just processes them without ever verifying them.

That relies on honesty of the collator: it can actually submit any messages it wants. To solve this. we should check if the messages supplied by the collator hash into the DMQ MQC beacon provided via the ground truth.

PR #258, opened 2020-12-03
Integrate HRMP

pepyakin:
This is a first stab at this and by no means final. As with UMP integration, we still lack a mechanism to pass the required data (such as, if there is a message we are intending to send a message to)

This also fixes a bug with sending an upward message where the same message is kept sending.

I recommend viewing commits independently


message-broker/src/lib.rs
```diff
@@ -75,16 +76,32 @@ decl_module! {
 				&dmp_count,
 			);
 
+			let mut hrmp_watermark = None::<relay_chain::BlockNumber>;
```
shawntabrizi, line 89: Is this type annotation needed?

message-broker/src/lib.rs
```diff
@@ -116,6 +124,50 @@ decl_module! {
 				);
 				*up = up.split_off(num);
 			});
+
+			// Sending HRMP messages is a little bit more involved. On top of the number of messages
+			// per block limit, there is also a constraint that it's possible to send only a single
+			// message to a given recipient per candidate.
+			let mut non_empty_hrmp_channels = <Self as Store>::NonEmptyHrmpChannels::get();
```
shawntabrizi, line 148: maybe a dumb question, but could all of this logic happen at the beginning of the next block?

shawntabrizi (approved):

message-broker/src/lib.rs
pepyakin: Not anymore, good catch!

collator/src/lib.rs
```diff
@@ -162,6 +162,31 @@ where
 			.ok()
 	}
 
+	/// Returns channels contents for each inbound HRMP channel addressed to the parachain we are
+	/// collating for.
+	///
+	/// Empty channels are also included.
+	fn retrieve_inbound_hrmp_channels_contents(&self, relay_parent: PHash)
```
rphmeier, line 172: nit: colloquially it reads more 'correctly' to say `hrmp_channel_contents`, even though it is reading many channels. You might change it to `all_hrmp_channel_contents` or similar if conveying the "all" aspect is important.

pepyakin: oh thanks! I did want to emphasize that it's multiple channels but I also didn't like how `hrmp_channels_contents`. `all_hrmp_channel_contents` indeed sounds better. Changed!

primitives/src/lib.rs
```diff
@@ -27,17 +27,35 @@ pub use polkadot_primitives::v1::{
...
+	use super::{InboundDownwardMessage, InboundHrmpMessage, ParaId};
+
+	/// Inherent identifier for message ingestion inherent.
+	pub const MESSAGE_INGESTION_IDENTIFIER: InherentIdentifier = *b"msgingst";
+	/// The data passed via a message ingestion inherent. Consists of a bundle of
+	/// DMP and HRMP messages.
+	#[derive(codec::Encode, codec::Decode, sp_core::RuntimeDebug, Clone, PartialEq)]
+	pub struct MessageIngestionType {
+		/// Downward messages in the order they were sent.
+		pub dmp: Vec<InboundDownwardMessage>,
```
bkchr, line 53: What does this `p` stand for?

And honestly I would like to have `downward_messages` and `horizontal_messages`.

I always need to think 5 minutes on what these abbreviations mean. :(

runtime/src/validate_block/implementation.rs
```diff
@@ -165,15 +166,28 @@ pub fn validate_block<B: BlockT, E: ExecuteBlock<B>>(params: ValidationParams) -
 		.and_then(|v| Decode::decode(&mut &v[..]).ok())
 		.expect("`ValidationData` is required to be placed into the storage!");
 
+	let horizontal_messages = match overlay.storage(HRMP_OUTBOUND_MESSAGES).flatten() {
+		Some(encoded) => Vec::<OutboundHrmpMessage>::decode(&mut &encoded[..])
+			.expect("Outbound HRMP messages vec is not correclty encoded in the storage!"),
```
bkchr, line 171:
```suggestion
			.expect("Outbound HRMP messages vec is not correctly encoded in the storage!"),
```

primitives/src/lib.rs
```diff
@@ -58,6 +76,18 @@ pub mod well_known_keys {
...
 	pub const NEW_VALIDATION_CODE: &'static [u8] = b":cumulus_new_validation_code:";
 
+	/// The storage key with which the runtime passes outbound HRMP messages it wants to send to the
+	/// PVF.
+	///
+	/// The value is stored as SCALE encoded `Vec<OutboundHrmpMessage>`
+	pub const HRMP_OUTBOUND_MESSAGES: &'static [u8] = b":cumulus_hrmp_outbound_messages:";
+
+	/// The storage key for communicating the HRMP watermark from the runtime to the PVF. Cleared by
+	/// the runtime each block and set after  message ingestion, but only if there were messages.
```
bkchr, line 86:
```suggestion
	/// the runtime each block and set after message inclusion, but only if there were messages.
```

collator/src/lib.rs
```diff
@@ -164,6 +165,31 @@ where
...
+	/// Empty channels are also included.
+	fn retrieve_all_inbound_hrmp_channel_contents(&self, relay_parent: PHash)
+		-> Option<BTreeMap<ParaId, Vec<InboundHrmpMessage>>>
+	{
+		self
+			.polkadot_client
+			.runtime_api()
+			.inbound_hrmp_channels_contents_with_context(
+				&BlockId::hash(relay_parent),
+				sp_core::ExecutionContext::Importing,
```
bkchr, line 180: Why a special context here?

collator/src/lib.rs
```diff
@@ -296,15 +331,50 @@ where
...
+				Some(Err(e)) => {
+					error!(
+						target: "cumulus-collator",
+						"Failed to decode the HRMP watermark: {:?}",
+						e
+					);
+					return None
+				}
+				None => {
+					// If the runtime didn't set `HRMP_WATERMARK`, then it means no messages were
```
bkchr, line 362: Should we, as a honest collator do the check here as well?

message-broker/src/lib.rs
```diff
@@ -27,57 +27,112 @@ use frame_support::{
...
+			let mut hrmp_watermark = None;
+			for (sender, channel_contents) in hrmp {
+				for hrmp_message in channel_contents {
+					hrmp_watermark = Some(hrmp_message.sent_at);
+
+					T::HrmpMessageHandlers::handle_hrmp_message(sender, hrmp_message);
+				}
+			}
+
+			// If we read at least one message, then advance watermark to that location.
```
bkchr, line 98:
```suggestion
			// If we processed at least one message, then advance watermark to that location.
```

message-broker/src/lib.rs
```diff
@@ -27,57 +27,112 @@ use frame_support::{
...
 			storage::unhashed::put(
 				well_known_keys::PROCESSED_DOWNWARD_MESSAGES,
-				&messages_len,
+				&dmp_count,
 			);
+
+			let mut hrmp_watermark = None;
+			for (sender, channel_contents) in hrmp {
+				for hrmp_message in channel_contents {
+					hrmp_watermark = Some(hrmp_message.sent_at);
```
bkchr, line 92: Should we not check that we only update the watermark if `sent_at` is greater?

message-broker/src/lib.rs
```diff
@@ -86,6 +141,50 @@ decl_module! {
 				);
 				*up = up.split_off(num);
 			});
+
+			// Sending HRMP messages is a little bit more involved. On top of the number of messages
+			// per block limit, there is also a constraint that it's possible to send only a single
+			// message to a given recipient per candidate.
+			let mut non_empty_hrmp_channels = <Self as Store>::NonEmptyHrmpChannels::get();
```
bkchr, line 148:
```suggestion
			let mut non_empty_hrmp_channels = NonEmptyHrmpChannels::get();
```

message-broker/src/lib.rs
```diff
@@ -86,6 +141,50 @@ decl_module! {
...
+			let outbound_hrmp_num = cmp::min(HRMP_MSG_NUM_PER_CANDIDATE, non_empty_hrmp_channels.len());
+			let mut outbound_hrmp_messages = Vec::with_capacity(outbound_hrmp_num);
+			let mut prune_empty = Vec::with_capacity(outbound_hrmp_num);
+
+			for &recipient in non_empty_hrmp_channels.iter().take(outbound_hrmp_num) {
```
bkchr, line 153: Doesn't this mean, if there are less parachains we want to send a message to than the upper number of messages per channel, we will waste space?

Assume the upper number is 5, but we only send messages to 2 parachains. This means we waste some pontential message space.


And the current algorithm favors parachains with lower ids. Meaning if we always send a message to one of these parachains, we may starve one of the higher parachains.

bkchr: > And the current algorithm favors parachains with lower ids. Meaning if we always send a message to one of these parachains, we may starve one of the higher parachains.

I see that this is discussed below.

message-broker/src/lib.rs
```diff
@@ -86,6 +141,50 @@ decl_module! {
...
+			// Prune hrmp channels that became empty. Additionally, because it may so happen that we
+			// only gave attention to some channels in `non_empty_hrmp_channels` it's important to
+			// change the order. Otherwise, the next `on_finalize` we will again give attention
+			// only to those channels that happen to be in the beginning, until they are emptied.
+			// This leads to "starvation" of the channels near to the end.
+			//
+			// To mitigate this we shift all processed elements towards the end of the vector using
+			// `rotate_left`. To get intution how it works see the examples in its rustdoc.
+			non_empty_hrmp_channels.retain(|x| !prune_empty.contains(x));
+			non_empty_hrmp_channels.rotate_left(outbound_hrmp_num - prune_empty.len());
```
bkchr, line 181: Doesn't this breaks the `binary_search` we do when inserting into this vector?

collator/src/lib.rs
pepyakin: actually, it meant to be not special (as with the dmq_contents). `Importing` seemed to be "less special" of all, at least according to [the docs](https://crates.parity.io/sp_core/enum.ExecutionContext.html). What are the other options?

collator/src/lib.rs
pepyakin: to see if the runtime tries to trick us? Interesting. I haven't thought about that. 

I would say no. The runtime is always more privileged than the node. We build cunning mechanisms to remove the possibility of nodes to bamboozle the runtime. However, I don't see how we could do the other way around: the runtime always can go rogue and sabotage the parachain (or essentially itself), be it subtly (by, say, not reading a message that it must've read) or blatantly (say, by just panicking unconditionally). And there is nothing that the node can do to prevent that.

So I think we do not need an additional check here - it would be superfluous

message-broker/src/lib.rs
pepyakin: well, yes and no.

Here, we rely on a property of a channel, that each message has `sent_at` always strictly greater than the last one. So it's this should be fine. However, we are not checking right now if the collator submits the right messages. In fact, it can submit any messages and get away with it. That's the concern for #235 though.

I can add an additional check but then with #235 it will get redundant so I'd rather not.

primitives/src/lib.rs
pepyakin: gotcha, will deabbreviate

message-broker/src/lib.rs
pepyakin: this does, good point! We don't need it to be sorted though.

message-broker/src/lib.rs
pepyakin: > Doesn't this mean, if there are less parachains we want to send a message to than the upper number of messages per channel, we will waste space?

Per candidate you mean? Yes, that's correct. We cannot send more than one message in a channel per one block. This is an artificial limitation though, it is not required for HRMP. It is mainly dictated by XCMP - there it is important.

bkchr: Ahh okay.

message-broker/src/lib.rs
bkchr: But we iterate here over all the channels? 

Can there not exist a channel that has `sent_at` being lower than the others?

collator/src/lib.rs
bkchr: Yeah good point

collator/src/lib.rs
bkchr: As this is part of the "block_production", I would have chosen this context. Or just no context at all :D

bkchr (approved):
After the `binary_search` bug is fixed, this is fine to merge.

message-broker/src/lib.rs
pepyakin: oh got it now, yes you are right

shawntabrizi:
bot merge

tolak:
Does this mean we can run XCM V0 on master now?

PR #269, opened 2020-12-21
Plumb polkadot backend into cumulus-collator

pepyakin:
This change plumbs the polkadot backend into the cumulus collator. This enables us to collate storage proofs from the relay-parent to the parachain. (Expected to come in following PRs)

Besides, the type of the inherent data payload was altered so as to incorporate the storage proof.

### DIFF
--- a/collator/src/lib.rs
+++ b/collator/src/lib.rs
@@ -215,7 +215,7 @@ where
 			.ok()?;
 
 		let validation_data = {
-			// TODO:
+			// TODO: Actual proof is to be created in the upcoming PRs.
 			let relay_chain_state = sp_state_machine::StorageProof::empty();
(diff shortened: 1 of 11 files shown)


collator/src/lib.rs
```diff
@@ -208,8 +214,17 @@ where
 			})
 			.ok()?;
 
+		let validation_data = {
+			// TODO: Actual proof is to be created in the upcoming PRs.
```
rphmeier, line 218: issue?

rphmeier (approved):

bkchr (approved):

collator/src/lib.rs
pepyakin: Implementation for this literally sits in the next commit : )

Issue #203, opened 2020-08-14
Best block is not updated if relay chain block is import before parachain block


bkchr:
When we import a relay chain block that sets a new best block for a parachain, but the parachain was not yet imported, the block will not be set a new best block. This means we need to create some synchronization that handles these race conditions.

PR #271, opened 2020-12-26
Wait for block import in parachain consensus

bkchr:
There was a bug in the parachain consensus that when importing a relay
chain block that sets a new best parachain block, but the required
parachain block was not yet imported. This pr fixes this by waiting for
the block to be imported.

Closes: https://github.com/paritytech/cumulus/issues/203


consensus/src/lib.rs
```diff
@@ -78,165 +70,318 @@ pub trait PolkadotClient: Clone + 'static {
 	) -> ClientResult<Option<Vec<u8>>>;
 }
 
-/// Finalize the given block in the Parachain.
-fn finalize_block<T, Block, B>(client: &T, hash: Block::Hash) -> ClientResult<bool>
+/// Follow the finalized head of the given parachain.
+///
+/// For every finalized block of the relay chain,
```
rphmeier, line 75: trailing doc

consensus/src/lib.rs
```diff
@@ -78,165 +70,318 @@ pub trait PolkadotClient: Clone + 'static {
...
+				tracing::warn!(
+					target: "cumulus-consensus",
+					error = ?err,
+					"Could not decode parachain header while following finalized heads.",
+				);
+				continue;
+			}
+		};
+
+		let hash = header.hash();
```
rphmeier, line 109: does the header actually need to be decoded for this? `blake2(&finalized_head[..])` should give the same hash. We would get an `UnknownBlock` error if it's an invalid header or we haven't yet seen the header.

consensus/src/lib.rs
```diff
@@ -78,165 +70,318 @@ pub trait PolkadotClient: Clone + 'static {
...
-	)
-	.map(|_| ()))
+	let follow_new_best = follow_new_best(
+		para_id,
+		parachain.clone(),
+		relay_chain.clone(),
+		announce_block,
+	);
+	let follow_finalized_head = follow_finalized_head(para_id, parachain, relay_chain);
+	select! {
```
rphmeier, line 167: instead of the `select!` macro, maybe just use `future::select`?

consensus/src/lib.rs
```diff
@@ -78,165 +70,318 @@ pub trait PolkadotClient: Clone + 'static {
...
+			}
+		};
+
+		let hash = header.hash();
+
+		// don't finalize the same block multiple times.
+		if parachain.usage_info().chain.finalized_hash != hash {
+			if let Err(e) = parachain.finalize_block(BlockId::hash(hash), None, true) {
+				match e {
+					ClientError::UnknownBlock(_) => tracing::debug!(
```
rphmeier, line 115: I suppose here there is a similar error to the one that the PR fixes. If the relay chain informs us about a finalized header before we've imported that header, then we won't finalize it, even if we import that same header immediately afterwards.

We will correctly finalize probably soon after, when we get a finality notification corresponding to a header we have already imported.

consensus/src/lib.rs
```diff
@@ -78,165 +70,318 @@ pub trait PolkadotClient: Clone + 'static {
...
-						error!(
-							target: "cumulus-collator",
-							"Trying to set pruned block `{:?}` as new best!",
-							hash,
+	let mut new_best_heads = relay_chain.new_best_heads(para_id)?.fuse();
+	let mut imported_blocks = parachain.import_notification_stream().fuse();
+	// The unset best header of the parachain. Will be `Some(_)` when we have imported a relay chain
+	// block before the parachain block it included. In this case we need to wait for this block to
+	// be imported to set it as new best.
+	let mut unset_best_header = None;
```
rphmeier, line 197: Down the line, we may even need to trigger an availability request to recover the PoV. It is possible that the initial collator is withholding the block.

consensus/src/lib.rs
```diff
@@ -78,165 +70,318 @@ pub trait PolkadotClient: Clone + 'static {
...
+			}
+			Ok(BlockStatus::InChainPruned) => {
+				tracing::error!(
+					target: "cumulus-collator",
+					block_hash = ?hash,
+					"Trying to set pruned block as new best!",
+				);
+			}
+			Ok(BlockStatus::Unknown) => {
+				*unset_best_header = Some(parachain_head);
```
rphmeier, line 336: overwriting does seem like the correct behavior.

consensus/src/lib.rs
```diff
@@ -78,165 +70,318 @@ pub trait PolkadotClient: Clone + 'static {
...
+		}
+	};
+
+	let hash = parachain_head.hash();
+
+	if parachain.usage_info().chain.best_hash == hash {
+		tracing::debug!(
+			target: "cumulus-consensus",
+			block_hash = ?hash,
+			"Skipping set new best block, because block is already the best.",
```
rphmeier, line 318: i'm curious if we do see duplicate notifications like this often.

consensus/src/lib.rs
```diff
@@ -78,165 +70,318 @@ pub trait PolkadotClient: Clone + 'static {
...
+		)
+	} else {
+		// Make sure the block is already known or otherwise we skip setting new best.
+		match parachain.block_status(&BlockId::Hash(hash)) {
+			Ok(BlockStatus::InChainWithState) => {
+				unset_best_header.take();
+
+				import_block_as_new_best(hash, parachain_head, parachain, announce_block);
```
rphmeier, line 326: why do we import as new best even though it's already `InChainWithState`? Is it a fork-choice rule thing?

consensus/src/lib.rs
```diff
@@ -78,165 +70,318 @@ pub trait PolkadotClient: Clone + 'static {
...
+			}
+			Ok(BlockStatus::InChainPruned) => {
+				tracing::error!(
+					target: "cumulus-collator",
+					block_hash = ?hash,
+					"Trying to set pruned block as new best!",
```
rphmeier, line 332: I guess this might happen as a chain reversion, but probably not since pruning happens after finality

rphmeier (approved):
Thanks for tests

rphmeier:
What was the downside of that bug? I suppose collators would still build on the right head because that's determined by the collation generation subsystem. So the RPC would show incorrect data as the "best block" was not updated?

bkchr:
> What was the downside of that bug? I suppose collators would still build on the right head because that's determined by the collation generation subsystem. So the RPC would show incorrect data as the "best block" was not updated?

This is not mainly about collators. Collators would still not be able to build a block, if they don't know the parachain block yet. So, we may need a similar fix for the collation as well. We already have something similar for block announcement validation, but there we need to wait for a relay chain block.

This is fix is more for full nodes like rpc or whatever nodes. With rococo v0 we had seen slow nodes like raspberry pi suffering under this problem that one of the blocks was always imported first. Maybe it was a combination of multiple things that lead to this to the seen problems. However, I more like to have this thing fixed to be sure we don't see this.
This could for example also lead to observations of a higher block time with rococo v1, because the rpc parachain rpc node imported the relay chain block before the parachain block.

consensus/src/lib.rs
JoshOrndorff: Anecdotally, I've never noticed it.

bkchr: Every relay chain block that does not includes a new parachain block, will trigger this log line. So, with rococo v1 at least every second block.

consensus/src/lib.rs
bkchr: I thought that we don't wanted to prevent to ask the relay chain validators to recover the povblocks?

But yeah, in general you are right. However this is not yet implemented anywhere.

consensus/src/lib.rs
bkchr: That is right, but also done intentionally by me. I do not expect that we are that much behind the tip that we can not finalize the block, because we did not have this block yet. If that is happening, we probably have other problems.

consensus/src/lib.rs
JoshOrndorff: Ah, yeah, it's a debug line. I've never seen it because I never run with that debug target.

consensus/src/lib.rs
rphmeier: Maybe should be filed for the "wishlist" but yeah I agree. Even the slight lag in finality should mean that blocks are well propagated before finality.

consensus/src/lib.rs
bkchr: AFAIK there is no other API for this. In the end we just import the header a second time and say that it should be imported as best block.

There once existed an API for this, but it was removed at some point.

consensus/src/lib.rs
bkchr: As the header is generic, a user could have whatever implementation for `hash()` they wants.

bkchr:
bot merge

Issue #221, opened 2020-09-22
Get parachain ID from relaychain


xlc:
Currently to deploy a parachain, we will need to know the parachain ID upfront and have it included in the genesis state.

This can be error-prone and requires generate genesis state after a parachain win a slot. We cannot get everything ready before a crowdsale starts.

Relaychain should be able to send some message to the parachain to tell what's its parachain id so there is one less way to configure it incorrectly.

gavofyork:
> This can be error-prone and requires generate genesis state after a parachain win a slot

No longer true. The parachain ID is pre-reserved ahead of time and has nothing to do with slot leases any more. That should alleviate this issue.

gavofyork:
> Relaychain should be able to send some message to the parachain to tell what's its parachain id so there is one less way to configure it incorrectly.

I doubt any of cumulus could work to deliver any such message unless its parachain ID is already known.

xlc:
Right. So we will know parachain ID prior genesis block so we can just hardcode the value.

Still a nice to have but not a priority at all.

gavofyork:
I really don't see how it's even logically possible. How could cumulus even connect with validators if it doesn't already know its parachain ID? what if two otherwise equal parachains begin at the same time? how are the collator nodes to know which is which?

xlc:
Make sense. We should be able to deploy two identical parachain on same slot and without hardcode the parachain ID upfront it will be hard to distinguish things.

PR #279, opened 2021-01-04
Inform the PVF with the latest relevant relay chain state

### LINKED ISSUE #261: Inform the cumulus runtime about a subset of the relay-chain state
In order a candidate has a chance to get included, it must pass the acceptance criteria check. The check is parametrized by the relay-chain configuration and state. Part of the parameters can change each session and other part up to each block.

Messaging is the biggest user of these data, it has quite some parameters. However, it's not the only. Due to the transient validation data is prone for manipulation (see https://github.com/paritytech/polkadot/issues/2085) we also have to pass the data about `code_upgrade_allowed` and `max_code_size`

Here is the enumeration of data we need in the cumulus runtime:

- configuration (from the runtime configuration module)
    - the number of UMP messages allowed per candidate
    - the number of HRMP messages allowed per candidate
    - e.g. the code size (i.e. the transient validation data pieces)
- ump
    - UMP queue number of messages and total size (`RelayDispatchQueueSize`)
- channels
    - open channels (i.e all) or channel openness (`HrmpEgressChannelsIndex(P)`)
    - the number of messages and the total size (`HrmpChannels(P)`)

pepyakin:
This PR changes the way that we inform cumulus runtime of the required state.

Background: The initial idea was to use the transient validation data, but then it was discovered that it doesn't work for the use-cases anticipated. Thus we decided to go with the merkle proof taken from the relay parent.

The result of this PR is that now the collator cannot trick the parachain not prevent upgrades and secondly the messaging now cannot exceed the limits invalidating the candidate.

I recommend going over each commit individually

Closes #261 
<s>Depends on https://github.com/paritytech/polkadot/pull/2194</s>
Unblocks https://github.com/paritytech/polkadot/issues/2085


wal004:
> Depends on [paritytech/polkadot#2194](https://github.com/paritytech/polkadot/pull/2194)

Would you be up for automatically managing dependent PRs? I built [Dpulls](https://github.com/marketplace/dpulls) to handle that - could be helpful if you often deal with dependent PRs - totally free for open source.

pepyakin:
I've rebased this PR to update to the latest polkadot master. No other changes were made.

message-broker/src/lib.rs
```diff
@@ -210,20 +310,57 @@ pub enum SendHorizontalErr {
 
 impl<T: Config> Module<T> {
 	pub fn send_upward_message(message: UpwardMessage) -> Result<(), SendUpErr> {
-		// TODO: check the message against the limit. The limit should be sourced from the
-		// relay-chain configuration.
+		let host_config = <cumulus_parachain_upgrade::Module<T>>::host_configuration().unwrap();
```
bkchr, line 313: Unwrap?

message-broker/src/lib.rs
```diff
@@ -127,53 +126,154 @@ decl_module! {
...
+				3 + hrmp_max_message_num_per_candidate as u64,
+				4 + hrmp_max_message_num_per_candidate as u64,
 			);
 
 			weight
 		}
 
 		fn on_finalize() {
+			let host_config = <cumulus_parachain_upgrade::Module<T>>::host_configuration().unwrap();
+			let relevant_messaging_state = <cumulus_parachain_upgrade::Module<T>>::relevant_messaging_state().unwrap();
```
bkchr, line 165: Unwraps?

message-broker/src/lib.rs
```diff
@@ -127,53 +126,154 @@ decl_module! {
...
+					.scan(
+						(available_capacity as usize, available_size as usize),
+						|(cnt, size), msg| match (cnt.checked_sub(1), size.checked_sub(msg.len())) {
+							(Some(cnt), Some(size)) => Some((cnt, size)),
+							_ => None,
+						},
+					)
+					.count();
+
+				// TODO: #274 Return back messages that do not longer fit into the queue.
```
bkchr, line 189: By that you mean messages that are now bigger than the maximum allowed message size?

bkchr: Why do we even support to change the max message size for an open channel?

bkchr: Ahh that is an upward message, this makes somewhat sense.

message-broker/src/lib.rs
```diff
@@ -127,53 +126,154 @@ decl_module! {
...
-			// per block limit, there is also a constraint that it's possible to send only a single
-			// message to a given recipient per candidate.
+			// Sending HRMP messages is a little bit more involved. There are the following
+			// constraints:
+			//
+			// - a channel should exist (and it can be closed while a message is buffered),
+			// - at most one message can be sent in a channel,
+			// - the sent out messages should be ordered by ascension of recipient para id.
+			// - the capacity and total size of the channel is limited,
+			// - the maximum size of a message is limited (and can potentially be changed),
```
bkchr, line 202: Now here is the comment correct, why do we allow to change the max message limit at all? :D

message-broker/src/lib.rs
```diff
@@ -127,53 +126,154 @@ decl_module! {
...
+				let became_empty = pending.is_empty();
+
+				if channel_meta.total_size + message_payload.len() as u32 > channel_meta.max_total_size {
+					// Sending this message will make the channel total size overflow. Skip it for now.
+					continue;
+				}
+
+				// If we reached here, then the channel has capacity to receive this message. However,
+				// it doesn't mean that we are sending it just yet.
+				<Self as Store>::OutboundHrmpMessages::insert(&recipient, pending);
```
bkchr, line 253:
```suggestion
				OutboundHrmpMessages::insert(&recipient, pending);
```

message-broker/src/lib.rs
```diff
@@ -127,53 +126,154 @@ decl_module! {
...
+				if became_empty {
+					prune_empty.push(recipient);
+				}
```
bkchr, line 256:
```suggestion
				if became_empty {
				        OutboundHrmpMessages::remove(&recipient);
					prune_empty.push(recipient);
				} else {
					OutboundHrmpMessages::insert(&recipient, pending);
				}
```

bkchr: WDYT of this?

bkchr (approved):
Besides my comments, it looks good.

collator/src/lib.rs
```diff
@@ -195,6 +195,77 @@ where
...
+			.transpose()
+			.map_err(|e| {
+				error!(
+					target: "cumulus-collator",
+					"Cannot decode the hrmp egress channel index: {:?}",
+					e,
+				)
+			})
+			.ok()?
+			.unwrap_or(vec![]);
```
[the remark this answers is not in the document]
pepyakin: FWIW, that's intentional, but I don't mind

parachain-upgrade/src/lib.rs
```diff
@@ -81,17 +105,6 @@ decl_module! {
...
-			System::<T>::can_set_code(&validation_function)?;
-			Self::schedule_upgrade_impl(validation_function)?;
-		}
-
-		/// Schedule a validation function upgrade without further checks.
-		///
-		/// Same as [`Module::schedule_upgrade`], but without checking that the new `validation_function`
-		/// is correct. This makes it more flexible, but also opens the door to easily brick the chain.
-		#[weight = (0, DispatchClass::Operational)]
-		pub fn schedule_upgrade_without_checks(origin, validation_function: Vec<u8>) {
```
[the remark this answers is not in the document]
pepyakin: oh good q: there were no difference in the previous version. Yes, in `schedule_upgrade` if the upgrade is prohibited by polkadot, the code would bail from `can_set_code` if `code_upgrade_allowed` is none. In `schedule_upgrade_without_checks` we would go straight to `schedule_upgrade_impl` and it would still check `code_upgrade_allowed` and bail if it's none, which makes the behavior identical to that of `schedule_upgrade`.

That seemed poppycocky to me, so I removed it

message-broker/src/lib.rs
pepyakin: That's embarassing

parachain-upgrade/src/lib.rs
```diff
@@ -53,6 +55,9 @@ pub trait Config: frame_system::Config {
 
 	/// Something which can be notified when the validation data is set.
 	type OnValidationData: OnValidationData;
+
+	/// Returns the parachain ID we are running with.
+	type SelfParaId: Get<ParaId>;
```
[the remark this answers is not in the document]
pepyakin: I was adding this with tears on my eyes, but yeah, turns out that we do need to know the parachains ID. 

Now, we still can get away without it right now. Incidentally, we only need it for stuff that is required for sending messages. So we can just return `None` from this and that should either disable or pause sending until the chain will get the correct para id, presumably set by the governance.

However, if we want to remove `dmq_mqc_head` and `hrmp_mqc_heads` from the `PersistentValidationData`, as per https://github.com/paritytech/polkadot/issues/2244, then the parahchain will have to know its para id otherwise should be there, say, a downward message and a parachain must handle at least one if there is any per parablock.

So I see two solutions here:

1. Add para id to the persisted validation data. It was even mentioned in issues under another angle https://github.com/paritytech/cumulus/issues/221
2. Push polkadot to always generate para id for a parachain before submitting the code.

message-broker/src/lib.rs
pepyakin: Yeah looks more sensible from 10kft

message-broker/src/lib.rs
pepyakin: Even funnier thing is that it's not even possible to renegotiate the limits for an existing channel.

For now it's hard to see how the cross chain messaging will turn out and how exactly it will be used. For now, it seems to me that allow the renegotiation of the sizes may come in handy for the parachains due to various reasons.

They off course can shut down the channel and open another one, but that would take some latency. Since we already have this possibility that the messages may be returned to the sender for some reason, we can use it as well for this case. So it doesn't really add a lot of extra complexity.

parachain-upgrade/src/lib.rs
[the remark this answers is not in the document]
bkchr: `can_set_code` is a Substrate function that checks that the given runtime is following some conventions, aka upgrading the spec_version etc. Please bring it back!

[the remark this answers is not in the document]
pepyakin: ah, right, of course, I'll bring it back

PR #285, opened 2021-01-13
upgrade log crate version to 0.4.13

atenjin:


bkchr (approved):

JoshOrndorff:
I'm curious about this PR because I've been struggling with dependencies lately. Would you mind sharing some knowledge?

1. Is it necessary that Cumulus use the same version of `log` that Substrate does?
2. If not necessary, is it desirable?
3. Was this upgrade solving a specific problem?

bkchr:
> * Is it necessary that Cumulus use the same version of `log` that Substrate does?

No. This is only a patch version difference.

> * If not necessary, is it desirable?

Less dependencies to compile.

PR #291, opened 2021-01-14
Replace ValidationData with PersistedValidationData

pepyakin:
Depends on https://github.com/paritytech/polkadot/pull/2272

This PR doesn't update the Cargo.lock yet since it depends on the polkadot PR. FWIW, this PR passes `cargo check` with a proper Cargo.lock locally


pepyakin:
Force pushed, because github didn't react on a normal push for some reason.

bkchr (approved):
This will require that all parachains are restarted. Otherwise looks good.

PR #295, opened 2021-01-18
Compact proof.

cheme:
This PR depends on :
https://github.com/paritytech/substrate/pull/8574

This PR adds code proof compression (the state proof).


cheme:
I switch this PR to also extract ':code' from proof.


The idea is that, with cumulus setting, validation code is the same as runtime code, so we have access to it while validating.

Therefore the code is removed from proof when compressing and added back when decompressing (in a lazy maner by some validation host function).



Notice that on polka size, code is pretty hacky, probably running
`
		let validation_code = runtime_api_request(
			ctx,
			descriptor.relay_parent,
			RuntimeApiRequest::ValidationCode(
				descriptor.para_id,
				assumption,
				code_tx,
			),
			code_rx,
		).await?;
`
would be less impacting (still need to have these parameters attached to validation data.


Also design could be more generic (maybe this can be applied to other key than ':code'):
- host function targetting known content: something like 'known_storage_key' returning 'Option<Option<Vec<u8>>'.
- trie decompress could use a lazy trait that take key as input and trigger on ALL encoded 0 length content.
(currently we need to know skip key when decompressing)
This requires to escape the compressed 0 length content to some known byte sequence (so it changes the encoding
in the rather rare case of empty content).


Regarding test done:
At this point only the fact that integration test is running (I did see root mismatch when I inserted &[] as code before linking the host function).

So clearly there is more work to do (polkadot and alternative design), but this seems pretty ok, so this question is as relevant as either.

cheme:
CI is failing with all the patched context, runs ok locally (with the given Cargo.lock)

cheme:
I did remove the skip wasm code (and polkadot host function), to keep this PR simple.

pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -65,11 +65,17 @@ pub fn validate_block<B: BlockT, E: ExecuteBlock<B>, PSC: crate::Config>(
...
-	}
+	// Uncompress
+	let mut db = MemoryDB::default();
+	let root = match sp_trie::decode_compact::<sp_trie::Layout<HashFor<B>>, _, _>(
+		&mut db,
+		storage_proof.iter_compact_encoded_nodes(),
+		Some(parent_head.state_root()),
+	) {
+		Ok(root) => root,
+		Err(_e) => panic!("Compact proof decoding failure."),
```
bkchr, line 76:
```suggestion
		Err(e) => panic!("Compact proof decoding failure: {:?}", e),
```

pallets/parachain-system/src/validate_block/tests.rs
```diff
@@ -134,6 +143,7 @@ fn validate_block_no_extra_extrinsics() {
 		witness,
 		validation_data,
 	} = build_block_with_witness(&client, vec![], parent_head.clone());
```
bkchr, line 145: Why isn't this function directly returning the `CompactProof`?

bkchr (approved):
2 small nitpicks

pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -65,11 +65,17 @@ pub fn validate_block<B: BlockT, E: ExecuteBlock<B>, PSC: crate::Config>(
...
-	if !HashDB::<HashFor<B>, _>::contains(&db, &root, EMPTY_PREFIX) {
-		panic!("Witness data does not contain given storage root.");
-	}
+	// Uncompress
+	let mut db = MemoryDB::default();
+	let root = match sp_trie::decode_compact::<sp_trie::Layout<HashFor<B>>, _, _>(
+		&mut db,
+		storage_proof.iter_compact_encoded_nodes(),
+		Some(parent_head.state_root()),
+	) {
```
bkchr, line 74: Ahh and didn't you added a method to the compactproof for uncompressing it?

cheme: We would have to rebuild the db afterward, here we directly got the db with the right content.

pallets/parachain-system/src/validate_block/implementation.rs
cheme: I will revert it (no debug implemented when using wasm :cry:

pallets/parachain-system/src/validate_block/implementation.rs
bkchr: Ahh okay

pallets/parachain-system/src/validate_block/tests.rs
cheme: :facepalm: done

pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -65,11 +65,17 @@ pub fn validate_block<B: BlockT, E: ExecuteBlock<B>, PSC: crate::Config>(
...
+		Ok(root) => root,
+		Err(e) => panic!("Compact proof decoding failure.", _e),
```
bkchr, line 76:
```suggestion
		Err(_) => panic!("Compact proof decoding failure."),
```

bkchr:
bot merge

PR #301, opened 2021-01-21
Complete telemetry for parachain & relaychain

### LINKED ISSUE #149: Telemetry cannot be used for parachain and for relaychain at the same time
This is a limitation in substrate. The telemetry must be initialized only once.

Related to https://github.com/paritytech/substrate/pull/6666 (this is not a fix as this doesn't allow using the telemetry on the parachain and the relaychain at the same time)

### LINKED ISSUE #154: Telemetry should be splitted
Currently when submitting telemetry messages, we will be sending relay chain and parachain messages always to the same endpoint. This will not work, because the telemetry can not distinguish them.

As for logging, we need some sort of scopes to distinguish them. This can be done with slogger and the feature for setting a scope per future. 

After we have scopes, we need to make the telemetry aware of these scopes and send the messages to the correct endpoint.

cecton:
screenshot

screenshot2

Closes #149 
Closes #154


rococo-parachains/src/service.rs
```diff
@@ -126,6 +126,7 @@ where
 
 	let params = new_partial(&parachain_config)?;
 	let telemetry_span = params.other;
+	let _telemetry_span_entered = telemetry_span.as_ref().map(|x| x.enter());
```
bkchr, line 129: Why do we need to do this here manually?

cecton: I need to investigate but I think there is something in a function somewhere below that reports something to the telemetry.

I know it's vague and it's probably not needed for the telemetry to work. I just wanted to make sure we don't miss a telemetry message at all.

bkchr: Please investigate this. The problem here is that this isn't any code that is reusable, so people would need to copy this.

cecton: Fixed by https://github.com/paritytech/substrate/pull/7951

cecton:
@bkchr this is also ready for review

bkchr:
@cecton fails to compile :P

rococo-parachains/src/command.rs
```diff
@@ -31,7 +30,7 @@ use sc_service::{
 	config::{BasePath, PrometheusConfig},
 	PartialComponents,
 };
-use sp_core::hexdisplay::HexDisplay;
+use sp_core::{hexdisplay::HexDisplay, Encode};
```
bkchr, line 33: Hmm?

test/service/src/genesis.rs
```diff
@@ -14,10 +14,10 @@
 // You should have received a copy of the GNU General Public License
 // along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
 
-use codec::Encode;
 use cumulus_primitives::{genesis::generate_genesis_block, ParaId};
 use cumulus_test_runtime::Block;
 use polkadot_primitives::v0::HeadData;
+use sp_core::Encode;
```
bkchr, line 20: What? Why you start importing this from `sp_core`?

bkchr (approved):
Please do not import `Encode` from `sp_core`. Otherwise it looks good.

test/service/src/genesis.rs
cecton: It doesn't work with the one of codec for some reason... I will check the dependencies

cecton:
```
error[E0599]: no method named `encode` found for reference `&sp_runtime::generic::Header<u32, BlakeTwo256>` in the current scope
  --> test/service/src/genesis.rs:27:37
   |
27 |     let genesis_state = block.header().encode();
   |                                        ^^^^^^ method not found in `&sp_runtime::generic::Header<u32, BlakeTwo256>`
   | 
  ::: /home/cecile/.cargo/git/checkouts/substrate-7e08433d4c370a21/93b231e/primitives/runtime/src/generic/header.rs:40:1
   |
40 | pub struct Header<Number: Copy + Into<U256> + TryFrom<U256>, Hash: HashT> {
   | -------------------------------------------------------------------------
   | |
   | doesn't satisfy `<_ as Deref>::Target = _`
   | doesn't satisfy `_: parity_scale_codec::Encode`
   | doesn't satisfy `_: parity_scale_codec::WrapperTypeEncode`
   |
   = note: the method `encode` exists but the following trait bounds were not satisfied:
           `<sp_runtime::generic::Header<u32, BlakeTwo256> as Deref>::Target = _`
           which is required by `sp_runtime::generic::Header<u32, BlakeTwo256>: parity_scale_codec::Encode`
           `sp_runtime::generic::Header<u32, BlakeTwo256>: parity_scale_codec::WrapperTypeEncode`
           which is required by `sp_runtime::generic::Header<u32, BlakeTwo256>: parity_scale_codec::Encode`
   = help: items from traits can only be used if the trait is in scope
   = note: the following trait is implemented but not in scope; perhaps add a `use` for it:
           `use sp_core::Encode;`
```

cecton: Ok I found it! It's hmm... single quote... so I failed my regex and one parity codec was not updated XD

Issue #303, opened 2021-01-21
Add relay parent into digest


bkchr:
We should put the relay parent into a digest in the header. This will be useful to find out on which relay parent a parachain block was build on.

bkchr:
This could probably be used for https://github.com/paritytech/cumulus/issues/320

rphmeier:
Seems to require https://github.com/paritytech/polkadot/issues/7191 to do safely.

PR #304, opened 2021-01-21
Add relay parent to inherent and to digest

JoshOrndorff:
This is my attempt at solving #303.

Feedback and suggestions welcome.

### DIFF
--- a/collator/src/lib.rs
+++ b/collator/src/lib.rs
@@ -297,6 +297,7 @@ where
 				horizontal_messages,
 				validation_data: validation_data.clone(),
 				relay_chain_state,
+				relay_parent,
 			}
 		};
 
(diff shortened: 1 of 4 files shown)


bkchr:
Ty for starting this!

So, I imagined it a little bit different. The digest is put as pre-runtime digest to the block in the collator implementation here: https://github.com/paritytech/cumulus/blob/b5f6580da1df1895b0b0984ac855a1a2a69ccc26/collator/src/lib.rs#L506

We would also need to check here that the `pre-runtime` digest exists:
https://github.com/paritytech/cumulus/blob/master/consensus/src/import_queue.rs

Regarding interfaces I would like to see something similar to this here:
https://github.com/paritytech/substrate/blob/master/primitives/consensus/babe/src/digests.rs#L158-L213

JoshOrndorff:
I might return to this, but I'm setting it aside for now. Thanks for the feedback on my initial draft.

PR #306, opened 2021-01-22
Add a command to purge the relay chain only

### LINKED ISSUE #259: Subcommand to purge embedded relay chain
Because parachain node include embedded relay chain nodes, they also have a relay chain block database. It will be useful to be able to purge that relay chain just as we can purge the parachain.

Maybe something like
```
./parachain-node purge-relay --base-path /collator1
```

cecton:
Fixes #259


cecton:
Tested and working:

```
[0] [07:55:34] ~/r/cumulus cecton-purge-relay > ./target/release/rococo-collator purge-chain --base-path /tmp/rococo
Are you sure to remove "/tmp/rococo/chains/local_testnet/db"? [y/N]: y
"/tmp/rococo/chains/local_testnet/db" removed.
[0] [07:55:43] ~/r/cumulus cecton-purge-relay > ./target/release/rococo-collator purge-relay-chain --base-path /tmp/rococo
Are you sure to remove "/tmp/rococo/polkadot/chains/westend_dev/db"? [y/N]: y
"/tmp/rococo/polkadot/chains/westend_dev/db" removed.
```

bkchr:
Commented on the issue

cecton:
```
[0] [11:29:44] ~/r/cumulus cecton-purge-relay > ./target/debug/rococo-collator purge-chain --parachain
/home/cecile/.local/share/rococo-collator/chains/local_testnet/db
Are you sure to remove? [y/N]: ^C⏎
[130] [11:36:24] ~/r/cumulus cecton-purge-relay > ./target/debug/rococo-collator purge-chain
/home/cecile/.local/share/rococo-collator/chains/local_testnet/db
/home/cecile/.local/share/rococo-collator/polkadot/chains/westend_dev/db
Are you sure to remove? [y/N]: ^C⏎
[130] [11:36:30] ~/r/cumulus cecton-purge-relay > ./target/debug/rococo-collator purge-chain --relay
/home/cecile/.local/share/rococo-collator/polkadot/chains/westend_dev/db
Are you sure to remove? [y/N]: ^C⏎
[130] [11:36:34] ~/r/cumulus cecton-purge-relay > ./target/debug/rococo-collator purge-chain --relay --para
/home/cecile/.local/share/rococo-collator/chains/local_testnet/db
/home/cecile/.local/share/rococo-collator/polkadot/chains/westend_dev/db
Are you sure to remove? [y/N]: y
"/home/cecile/.local/share/rococo-collator/chains/local_testnet/db" did not exist.
"/home/cecile/.local/share/rococo-collator/polkadot/chains/westend_dev/db" did not exist.
```

rococo-parachains/src/cli.rs
```diff
@@ -143,16 +147,103 @@ pub struct RelayChainCli {
...
 	}
 }
+
+#[derive(Debug, StructOpt)]
+pub struct PurgeChainCmd {
+	#[structopt(flatten)]
+	base: sc_cli::PurgeChainCmd,
+
+	/// Only delete the para chain database
+	#[structopt(long = "para-chain", aliases = &["para", "parachain"])]
```
bkchr, line 175:
```suggestion
	#[structopt(long = "parachain", aliases = &["para"])]
```

rococo-parachains/src/cli.rs
```diff
@@ -143,16 +147,103 @@ pub struct RelayChainCli {
...
 		Self {
 			base_path,
 			chain_id,
 			base: polkadot_cli::RunCmd::from_iter(relay_chain_args),
 		}
 	}
 }
+
+#[derive(Debug, StructOpt)]
+pub struct PurgeChainCmd {
```
bkchr, line 170: As described in the ticket. Could we please move this to its own crate to make it reusable? :D

bkchr (commented):
Please move and also please write some tests for the command on its own.

rococo-parachains/src/cli.rs
cecton: I missed that part :D ok I will do

rococo-parachains/src/command.rs
```diff
@@ -204,6 +204,27 @@ pub fn run() -> Result<()> {
...
+						.iter()
+						.chain(cli.relaychain_args.iter()),
+				);
+
+				let polkadot_config = SubstrateCli::create_configuration(
+						&polkadot_cli,
+						&polkadot_cli,
+						config.task_executor.clone(),
+						None,
+					).map_err(|err| format!("Relay chain argument error: {}", err))?;
```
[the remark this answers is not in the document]
cecton: I think that is solved...

cli/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+
+use sc_cli;
+use std::{
+	fs,
+	io::{self, Write},
+};
+use structopt::StructOpt;
+
+#[derive(Debug, StructOpt)]
+pub struct PurgeChainCmd {
```
bkchr, line 25: Docs? :D

bkchr (commented):
(and please don't forget the tests :see_no_evil: )

cecton:
> (and please don't forget the tests see_no_evil )

I'm on it. You're too fast

bkchr:
> > (and please don't forget the tests see_no_evil )
> 
> I'm on it. You're too fast

Sorry :P

cli/src/lib.rs
cecton: Done

cecton:
There is an issue with the `--dev` flag. There is one for the para chain, one for the relay chain. For some reasons (limitation in structopt or maybe clap) I have problems to retrieve the parameters after `--` for the `purge-chain` subcommand... and that is a big problem to solve.

The other problem with this is that it is also impacting the `--chain`. The only way I see right now would be to add parameters `--relay-dev` and `--relay-chain` to the cumulus purge-chain command.

Meanwhile I have a few questions:
- Are we sure we need to support polkadot args in the purge-chain command?
- Do I need to use the `--dev` flag at all in the tests? Before it was required to avoid connecting the outside world I think but now when I look at the example in the README and we don't use --dev at all anywhere.

One other (very far fetched) alternative would be to give a big thought about moving away from structopt and changing completely sc-cli's API to use clap directly. It's less declarative but this is actually a good thing in term of flexibility. We might be able to achieve more with less complexity.

bkchr:
> There is an issue with the `--dev` flag. There is one for the para chain, one for the relay chain. For some reasons (limitation in structopt or maybe clap) I have problems to retrieve the parameters after `--` for the `purge-chain` subcommand... and that is a big problem to solve.
> 
> The other problem with this is that it is also impacting the `--chain`. The only way I see right now would be to add parameters `--relay-dev` and `--relay-chain` to the cumulus purge-chain command.

`--dev` is just a shortcut for `--chain something`. So, IMHO we could omit this here because it doesn't make any sense.

> 
> Meanwhile I have a few questions:
> 
>     * Are we sure we need to support polkadot args in the purge-chain command?
> 
>     * Do I need to use the `--dev` flag at all in the tests? Before it was required to avoid connecting the outside world I think but now when I look at the example in the README and we don't use --dev at all anywhere.
> 

You should make yourself familiar what this command is doing. From the perspective of cumulus the `--dev` never worked, but we could probably make it work. As explained above, the `--dev` is just a shortcut for `--chain bla` and you can use that in your tests.

> 
> One other (very far fetched) alternative would be to give a big thought about moving away from structopt and changing completely sc-cli's API to use clap directly. It's less declarative but this is actually a good thing in term of flexibility. We might be able to achieve more with less complexity.

We used clap before and it was horrible :P Yeah, it is more flexible, but you also loose all your shit when you try to modify something or to add a new parameter. So, no switch.

cecton:
> As explained above, the --dev is just a shortcut for --chain bla and you can use that in your tests.

I know that very well but my problem is still there. I will try to explain better:

When you run cumulus you can give a `--chain` to the relay chain and another one to the para chain. It works because of how things are parse in the command line.

Unfortunately, because of limitations with structopt, I can't parse the arguments after `--` when using the purge command. Because of that, the test fails because the process is first started with one chain (dev) but the purge command (ran later on during the test) is using another chain (westend-dev).

What I can do to circumvent this issue:
1. Add a new parameter `--relay-chain` to the purge command
2. Store the chain name in a file in the base path of the para chain (so it can locate the directory)

(I'm mostly using the parameter `--dev` in the test to avoid the nodes to connect to anything.)

cecton:
> Add a new parameter `--relay-chain` to the purge command

Wait, there is already a parameter `--relay-chain`. I meant a parameter `--relay-chain-chain <chain-file>` (probably with a better name lol)

cecton:
> Wait, there is already a parameter `--relay-chain`. I meant a parameter `--relay-chain-chain <chain-file>` (probably with a better name lol)

Or maybe just make `--relaychain` takes an optional path argument to the chain file

cecton:
Ok I get it, the relay chain is actually found from the parachain's extension. So I just fixed the test. @bkchr this is ready for review then

cli/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+/// The `purge-chain` command used to remove the whole chain: the parachain and the relaychain.
+#[derive(Debug, StructOpt)]
+pub struct PurgeChainCmd {
+	/// The base struct of the purge-chain command.
+	#[structopt(flatten)]
+	pub base: sc_cli::PurgeChainCmd,
+
+	/// Only delete the para chain database
+	#[structopt(long = "parachain", aliases = &["para"])]
```
bkchr, line 36:
```suggestion
	#[structopt(long, aliases = &["para"])]
```

cli/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+	pub parachain: bool,
+
+	/// Only delete the relay chain database
+	#[structopt(long = "relaychain", aliases = &["relay"])]
```
bkchr, line 40:
```suggestion
	#[structopt(long, aliases = &["relay"])]
```

cli/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+			(true, true) | (false, false) => vec![para_config.database, relay_config.database],
+			(true, false) => vec![para_config.database],
+			(false, true) => vec![relay_config.database],
+		};
+
+		let db_paths = databases
+			.iter()
+			.map(|x| {
+				x.path().ok_or_else(|| {
+					sc_cli::Error::Input("Cannot purge custom database implementation".into())
```
bkchr, line 61: Maybe a hint which on this is, would be nice :D

cli/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+		for db_path in &db_paths {
+			match fs::remove_dir_all(&db_path) {
+				Ok(_) => {
+					println!("{:?} removed.", &db_path);
+				}
+				Err(ref err) if err.kind() == io::ErrorKind::NotFound => {
+					eprintln!("{:?} did not exist.", &db_path);
+				}
+				Err(err) => return Result::Err(err.into()),
```
bkchr, line 94:
```suggestion
				Err(err) => return Err(err.into()),
```

rococo-parachains/tests/purge_chain_works.rs
```diff
@@ -15,50 +15,63 @@
...
 
 use assert_cmd::cargo::cargo_bin;
-use std::{convert::TryInto, fs, path::PathBuf, process::Command, thread, time::Duration};
+use std::{convert::TryInto, process::Command, thread, time::Duration};
 
 mod common;
 
 #[test]
 #[cfg(unix)]
 fn purge_chain_works() {
```
bkchr, line 24: Would be really nice to have these tests part of the `/tests/` "parachain". But this can be done in a later pr.

bkchr (approved):
Sorry for the late review. Some last nitpicks, otherwise looks good.

cli/src/lib.rs
cecton: Done in 670b674

rococo-parachains/tests/purge_chain_works.rs
```diff
@@ -15,50 +15,63 @@
...
-		unistd::Pid,
-	};
+	fn run_node_and_stop() -> tempfile::TempDir {
+		use nix::{
+			sys::signal::{kill, Signal::SIGINT},
+			unistd::Pid,
+		};
 
-	let base_path = "purge_chain_test";
+		let base_path = tempfile::tempdir().unwrap();
```
JoshOrndorff, line 31: @cecton I'm curious why you did this after you already gave us the `--tmp` flag in https://github.com/paritytech/substrate/pull/6221

JoshOrndorff: Oh, I get it. So you can check the directory later on.

PR #308, opened 2021-01-22
MQC authorization

### LINKED ISSUE #235: Authorize incoming downward and horizontal messages

pepyakin:
Closes #235 

Apart from that this changes the order of message dispatching. After this PR the horizontal messages will be dispatched in the order they were relayed (the property `sent_at` of a message) and messages that were relayed at the same block will have a tiebreaker of para id.


parachain-system/src/lib.rs
```diff
@@ -1351,4 +1524,300 @@ mod tests {
...
+			.with_validation_data(
+				|_, relay_block_num, validation_data| match relay_block_num {
+					1 => {
+						// 200 - doesn't exist yet
+						// 300 - one new message
+						validation_data.hrmp_mqc_heads.push((
+							ParaId::from(300),
+							hex![
+								"d8d73e427fbcb01793aae939418c28af9f570da4f3d8430076a58844397c78a5"
+							]
```
bkchr, line 1578: Maybe we should have a function for calculating this hash. We could use it in the production code and also in tests to not hardcode these values.

bkchr (approved):
Nice pr! :)

parachain-system/src/lib.rs
pepyakin: Yeah, I considered that.

I deliberately chose this hardcoded snapshot like approach. My reasoning was that if you inadvertently changed the function, it will likely pass the tests, whereas it should not. If you do want to change the function, that will involve changing the hashes which is admittedly annoying, but still not that bad and unlikely to happen.

pepyakin: On the second thought, we can extract the function (or functions, because there are two actually) and then test that it returns expected results.

parachain-system/src/lib.rs
```diff
@@ -183,41 +206,10 @@ decl_module! {
 			RelevantMessagingState::put(relevant_messaging_state);
 			HostConfiguration::put(host_config);
 
-			<T::OnValidationData as OnValidationData>::on_validation_data(vfp);
+			<T::OnValidationData as OnValidationData>::on_validation_data(vfp.clone());
```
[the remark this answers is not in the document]
pepyakin: Yeah that makes sense

parachain-system/src/lib.rs
```diff
@@ -96,6 +106,19 @@ decl_storage! {
...
 		HostConfiguration get(fn host_configuration): Option<AbridgedHostConfiguration>;
 
+		/// The last downward message queue chain head we have observed or all zeroes if we haven't observed
+		/// any.
+		///
+		/// This value is loaded before and saved after processing inbound downward messages carried
+		/// by the system inherent.
+		LastDmqMqcHead: relay_chain::Hash;
+		/// The message queue chain heads we have observed per each channel incoming channel. If the
+		/// value is missing it is treated as all zeroes.
```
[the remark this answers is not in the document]
pepyakin: The reason why a missing value is treated with all zeroes is because we state that on the polkadot side, [here](https://github.com/paritytech/polkadot/blob/1dd53d4345b13b15f67d9c4f54d8b63cc0ba1902/runtime/parachains/src/hrmp.rs#L78-L85)

It says to us that the previous message queue chain head for an empty queue is all zeros. So whenever we get an inherent that has some MQCs that we previously didn't know about, we treat them as all zeros.

However, after processing the batch we overwrite the whole structure with the MQC heads that we observed in the last round.

Does this make sense?

parachain-system/src/lib.rs
```diff
@@ -443,6 +435,119 @@ decl_module! {
 }
 
 impl<T: Config> Module<T> {
+	/// Process all inbound downward messages relayed by the collator.
+	///
+	/// Checks if the sequence of the messages is valid, dispatches them and communicates the number
+	/// of processed messages to cumulus via a storage update.
```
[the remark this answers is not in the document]
pepyakin: 🙈  ok I meant the collator )

parachain-system/src/lib.rs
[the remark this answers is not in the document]
bkchr: Yeah that also makes sense from the code, however isn't it this case that this storage item will never have the empty head? Because when we insert some value for a parachain, it will always be a non-all-zero head?

parachain-system/src/lib.rs
```diff
@@ -443,6 +435,119 @@ decl_module! {
...
+		horizontal_messages.sort_by(|a, b| {
+			// first sort by sent-at and then by the para id
+			match a.1.sent_at.cmp(&b.1.sent_at) {
+				cmp::Ordering::Equal => a.0.cmp(&b.0),
+				ord => ord,
+			}
+		});
+
+		let last_mqc_heads = LastHrmpMqcHeads::get();
+		let mut running_mqc_heads = BTreeMap::new();
```
bkchr, line 500: When a parachain didn't send us a message in this round, don't we currently delete the last head? Don't we require this when we receive the next message in the future?

parachain-system/src/lib.rs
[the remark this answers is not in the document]
pepyakin: Actually, the values may have an empty head. This can happen when we observe a new MQC head from persisted validation data but receive no messages. In this case a zero will end up in this storage item.

parachain-system/src/lib.rs
```diff
@@ -443,6 +435,119 @@ decl_module! {
...
+				.entry(sender)
+				.or_insert_with(|| last_mqc_heads.get(&sender).cloned().unwrap_or_default());
+			*cur_mqc_head = BlakeTwo256::hash_of(&(
+				*cur_mqc_head,
+				horizontal_message.sent_at,
+				BlakeTwo256::hash_of(&horizontal_message.data),
+			));
+
+			ensure!(
+				vfp.hrmp_mqc_heads
```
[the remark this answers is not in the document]
pepyakin: Not sure how that could be done below, but I moved it before we flatten the BTreeMap.

parachain-system/src/lib.rs
```diff
@@ -96,6 +106,19 @@ decl_storage! {
 		/// This data is also absent from the genesis.
 		HostConfiguration get(fn host_configuration): Option<AbridgedHostConfiguration>;
 
+		/// The last downward message queue chain head we have observed or all zeroes if we haven't observed
+		/// any.
+		///
+		/// This value is loaded before and saved after processing inbound downward messages carried
+		/// by the system inherent.
+		LastDmqMqcHead: relay_chain::Hash;
```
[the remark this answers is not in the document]
pepyakin: I didn't do this change, because I changed `relay_chain::Hash` to `MessageQueueChain`, which hopefully makes it clear that we rely on the `Default` implementation here.

I am not fundamentally opposed to change it to `Option` though, so just tell

parachain-system/src/lib.rs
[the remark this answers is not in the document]
pepyakin: https://github.com/paritytech/cumulus/pull/308/commits/2ab6df90b9e0818bf2bd8a2530f63268c5759ed2

parachain-system/src/lib.rs
pepyakin: https://github.com/paritytech/cumulus/pull/308/commits/388168e300246237f3cfecc9f1a8740e208128dd

parachain-system/src/lib.rs
pepyakin: Fixed by https://github.com/paritytech/cumulus/pull/308/commits/bf6fff56206c3a0f40c9f0549ca3e2a97b80f50b, good catch

parachain-system/src/lib.rs
```diff
@@ -443,6 +432,126 @@ decl_module! {
...
+		}
+
+		// Check that the MQC heads for each channel provided by the relay chain match the MQC heads
+		// we have after processing all incoming messages.
+		//
+		// Along the way we also carry over the relevant entries from the `last_mqc_heads` to
+		// `running_mqc_heads`. Otherwise, in a block where no messages were sent in a channel
+		// it won't get into next block's `last_mqc_heads` and thus will be all zeros, which
+		// would corrupt the message queue chain.
+		for &(ref sender, ref target_head) in &vfp.hrmp_mqc_heads {
```
bkchr, line 536: I assume these are always all heads, even from the paras that didn't send us any messages in this block? :D

bkchr (approved):

parachain-system/src/lib.rs
pepyakin: Yeah, those are all. Good idea to leave a comment about that on `hrmp_mqc_heads`.

pepyakin: Ah, wait, I just removed it in the [next PR](https://github.com/paritytech/cumulus/pull/317) 🤦‍♂️

pepyakin:
bot merge

PR #309, opened 2021-01-24
Fix HRMP message source location 

akru:
**Question**

Are HRMP messages should be tagged as routed from the relay chain for the XCM executor, didn't they?

### DIFF
--- a/xcm-handler/src/lib.rs
+++ b/xcm-handler/src/lib.rs
@@ -122,10 +122,11 @@ impl<T: Config> HrmpMessageHandler for Module<T> {
 		frame_support::debug::print!("Processing HRMP XCM: {:?}", &hash);
 		match VersionedXcm::decode(&mut &msg.data[..]).map(Xcm::try_from) {
 			Ok(Ok(xcm)) => {
-				match T::XcmExecutor::execute_xcm(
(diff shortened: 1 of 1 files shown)


shawntabrizi:
@akru as far as I understand, horizontal messages are directly from Parachain to Parachain, so this change does not make sense to me.

pepyakin:
Right

akru:
@shawntabrizi probably, you didn’t write XCM executor. With your position is difficult to explain this naming https://github.com/paritytech/polkadot/blob/master/xcm/xcm-builder/src/location_conversion.rs#L92

So, let’s make it clear. HRMP message routing is parachain-to-parachain, as you said, right? Routing information about location should be removed, but in fact hrmp routes message, it’s not direct way, right?

shawntabrizi (commented):
@akru my apologies, I believe you were totally correct here.

An important note that I got from Gav was that the multilocation is completely unaffected by the messaging/delivery system. This was a fundamental misunderstanding that I had.

As such, when you have `Parachain A` and `Parachain B` connected to a `Parent Relay Chain`, the multi location in the eyes of `Parachain B` will always be `Location(Parent, Parachain A)`.

shawntabrizi (approved):

PR #315, opened 2021-01-29
Block announce validation should use the correct `Validation` result

bkchr:
The error variant is just for internal errors and we need to return
`Failure` always when the other node send us an invalid statement.

### DIFF
--- a/network/src/lib.rs
+++ b/network/src/lib.rs
@@ -54,9 +54,11 @@ use futures::{
 };
 use log::trace;
 
-use std::{fmt, marker::PhantomData, pin::Pin, sync::Arc, convert::TryFrom};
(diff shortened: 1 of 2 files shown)


network/src/lib.rs
```diff
@@ -143,10 +152,12 @@ impl BlockAnnounceData {
 		let signer = match authorities.get(validator_index as usize) {
 			Some(r) => r,
 			None => {
-				return Err(BlockAnnounceError(
-					"block accouncement justification signer is a validator index out of bound"
-						.to_string(),
-				));
+				log::debug!(
+					target: LOG_TARGET,
+					"Block accouncement justification signer is a validator index out of bound",
```
pepyakin, line 157:
```suggestion
					"Block announcement justification signer is a validator index out of bound",
```

pepyakin (approved):
I don't feel confident, but the change makes sense to me

PR #317, opened 2021-01-29
Migrate from MQCs in persisted validation data to merkle proofs

pepyakin:
- [x] Run this locally to check if it works

### DIFF
--- a/test/relay-sproof-builder/src/lib.rs
+++ b/test/relay-sproof-builder/src/lib.rs
@@ -22,7 +22,16 @@ use sp_std::collections::btree_map::BTreeMap;
 /// Builds a sproof (portmanteau of 'spoof' and 'proof') of the relay chain state.
 #[derive(Clone)]
 pub struct RelayStateSproofBuilder {
+	/// The para id of the current parachain.
+	///
+	/// This doesn't get into the storage proof produced by the builder, however, it is used for
+	/// generation of the storage image and by auxilary methods.
+	///
+	/// It's recommended to change this value once in the very beginning of usage.
+	///
+	/// The default value is 200.
 	pub para_id: ParaId,
+
 	pub host_config: AbridgedHostConfiguration,
 	pub dmq_mqc_head: Option<relay_chain::Hash>,
 	pub relay_dispatch_queue_size: Option<(u32, u32)>,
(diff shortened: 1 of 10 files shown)


pepyakin:
Some changes related to tests were lost somehow. WIP

pepyakin:
I cannot test this locally, blocked by https://github.com/paritytech/polkadot/issues/2373

I think we can still review and land this and test the whole thing asynchronously

test/relay-sproof-builder/src/lib.rs
```diff
@@ -44,14 +55,43 @@ impl Default for RelayStateSproofBuilder {
...
+	///
+	/// If there is no channel, a new default one is created.
+	///
+	/// It also updates the `hrmp_ingress_channel_index`, creating it if needed.
+	pub fn upsert_inbound_channel(&mut self, sender: ParaId) -> &mut AbridgedHrmpChannel {
+		let in_index = self.hrmp_ingress_channel_index.get_or_insert_with(Vec::new);
+		match in_index.binary_search(&sender) {
+			Ok(_idx) => {}
+			Err(idx) => in_index.insert(idx, sender),
+		}
```
bkchr, line 78:
```suggestion
		if let Err(idx) = in_index.binary_search(&sender) {
			in_index.insert(idx, sender);
		}
```

test/relay-sproof-builder/src/lib.rs
```diff
@@ -44,14 +55,43 @@ impl Default for RelayStateSproofBuilder {
...
 	}
 }
 
 impl RelayStateSproofBuilder {
+	/// Returns a mutable reference to HRMP channel metadata for a channel (`sender`, `self.para_id`).
+	///
+	/// If there is no channel, a new default one is created.
+	///
+	/// It also updates the `hrmp_ingress_channel_index`, creating it if needed.
+	pub fn upsert_inbound_channel(&mut self, sender: ParaId) -> &mut AbridgedHrmpChannel {
```
bkchr, line 73: `upsert` I needed to google this :D But I also got no better idea.

pepyakin: Rust uses `get_or_insert`, [here](https://doc.rust-lang.org/1.48.0/core/option/enum.Option.html#method.get_or_insert). Perhaps we could use it, but that seemed to me a bit overly long.

bkchr: Yeah IDK

bkchr: I also thought about this, but it would probably too long as you said.

Issue #36, opened 2020-01-15
Switch from `Proposer` to `Slots`


bkchr:
Instead of using `Proposer` as interface for building the blocks in the collator, we probably want to switch over to `Slots` as the interface. At the point when we want to support custom authoring, using `Slots` is easier as Babe and Aura already implemented using this trait.

We would need to change the trait in Substrate to return the `Proposal`, because we need this in the collator.

JoshOrndorff:
You mean switching to [`SlotWorker`](https://crates.parity.io/sc_consensus_slots/trait.SlotWorker.html) right? Or maybe [`SimpleSlotWorker`](https://crates.parity.io/sc_consensus_slots/trait.SimpleSlotWorker.html)?

Would this change eliminate the possibility of using PoW for collating? Is that a compromise we really want?

bkchr:
> You mean switching to [`SlotWorker`](https://crates.parity.io/sc_consensus_slots/trait.SlotWorker.html) right? Or maybe [`SimpleSlotWorker`](https://crates.parity.io/sc_consensus_slots/trait.SimpleSlotWorker.html)?

`SlotWorker`.

> 
> Would this change eliminate the possibility of using PoW for collating? Is that a compromise we really want?

We don't really support PoW at the moment at all. So, yes.

JoshOrndorff:
I started looking at making the change to the `SlotWorker` trait, and I have a few questions about it.

1. [`Proposal`](https://crates.parity.io/sp_consensus/struct.Proposal.html) is generic over the `Transaction` type used in the backend. Should I make the `SlotWorker` trait also generic over this type?

2. Currently the only implementation of the `SlotWorker` trait come from the [blanket implementation](https://github.com/paritytech/substrate/blob/26f3d6826a181674257b8e9fc708aa04b49dc73c/client/consensus/slots/src/lib.rs#L385-L391) over the `SimpleSlotWorker` trait.
That means that I'll need to update the provided `on_slot` method for the `SimpleSlotWorker` trait. I'm stuck because I can't figure out how to return the proposal [here](https://github.com/paritytech/substrate/blob/26f3d6826a181674257b8e9fc708aa04b49dc73c/client/consensus/slots/src/lib.rs#L378). After it's data has already been moved [eg here](https://github.com/paritytech/substrate/blob/26f3d6826a181674257b8e9fc708aa04b49dc73c/client/consensus/slots/src/lib.rs#L345). It seems Proposal doesn't implement `Clone`.

[...]

PR #329, opened 2021-02-15
Parachain Consensus abstractions

bkchr:
This pr introduces abstractions around the parachain consensus that is driven by the collator. It moves the current implementation into the `RelayChainConsensus` as a consensus that is entirely dependent on the relay chain.

Fixes: https://github.com/paritytech/cumulus/issues/36


client/consensus/relay-chain/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+//!
+//! 1. Each node that sees itself as a collator is free to build a parachain candidate.
+//!
+//! 2. This parachain candidate is send to the parachain validators that are part of the relay chain.
+//!
+//! 3. The parachain validators validate at most X different parachain candidates, where X is the
+//! total number of parachain validators.
+//!
+//! 4. The parachain candidate with the highest number of approvals is choosen by the relay-chain
+//! block producer to be backed.
```
burdges, line 32: Approvals means supporting parachain validators I guess.

client/consensus/common/src/lib.rs
```diff
@@ -511,6 +510,52 @@ where
...
+#[async_trait::async_trait]
+pub trait ParachainConsensus<B: BlockT>: Send + Sync + dyn_clone::DynClone {
+	/// Produce a new candidate at the given parent block.
+	///
+	/// Should return `None` if the consensus implementation decided that it shouldn't build a
+	/// candidate or if there occurred any error.
+	///
+	/// # NOTE
+	///
+	/// It is expected that the block is
```
rphmeier, line 536: trailing

client/consensus/relay-chain/src/lib.rs
rphmeier: Yeah, prefer the term backing as approval means something else in parachain consensus.

client/consensus/common/src/lib.rs
```diff
@@ -511,6 +510,52 @@ where
 	}
 }
 
+/// The result of [`ParachainConsensus::produce_candidate`].
+pub struct ParachainCandidate<B> {
+	/// The block that was build for this candidate.
```
JoshOrndorff, line 515:
```suggestion
	/// The block that was built for this candidate.
```

client/consensus/common/src/lib.rs
```diff
@@ -511,6 +510,52 @@ where
...
+	/// The proof that was recorded while building the block.
+	pub proof: sp_trie::StorageProof,
+}
+
+/// A specific parachain consensus implementation that can be used by a collator to produce candidates.
+///
+/// The collator will call [`Self::produce_candidate`] every time there is a free core for the parachain
+/// this collator is collating for. It is the job of the consensus implementation to decide if this
+/// specific collator should build candidate for the given relay chain block. The consensus
+/// implementation could for example check if this specific collator is part of the validator.
```
JoshOrndorff, line 526:
```suggestion
/// The collator will call [`Self::produce_candidate`] every time there is a free core for the parachain
/// this collator is collating for. It is the job of the consensus implementation to decide whether this
/// specific collator should build a candidate for the given relay chain block. The consensus
/// implementation could, for example, check whether this specific collator is part of a staked set.
```

client/collator/src/lib.rs
```diff
@@ -343,14 +260,15 @@ where
 		relay_parent: PHash,
 		validation_data: PersistedValidationData,
 	) -> Option<CollationResult> {
-		trace!(target: LOG_TARGET, "Producing candidate");
+		tracing::trace!(target: LOG_TARGET, "Producing candidate");
```
pepyakin, line 263: worth adding `relay_parent` tag here?

client/collator/src/lib.rs
```diff
@@ -20,219 +20,131 @@ use cumulus_client_network::WaitToAnnounce;
...
 	block_status: Arc<BS>,
+	parachain_consensus: Box<dyn ParachainConsensus<Block>>,
 	wait_to_announce: Arc<Mutex<WaitToAnnounce<Block>>>,
 	backend: Arc<Backend>,
-	polkadot_client: Arc<PClient>,
-	polkadot_backend: Arc<PBackend2>,
 }
 
-impl<Block: BlockT, PF, BI, BS, Backend, PBackend, PClient, PBackend2> Clone
-	for Collator<Block, PF, BI, BS, Backend, PBackend, PClient, PBackend2>
```
pepyakin, line 78: Yay!

pepyakin (approved):
Looks good overall. I am not sure how it fixes #36 though

bkchr:
> Looks good overall. I am not sure how it fixes #36 though

#36 was a rather shitty issue. In the end I settled with myself on requiring that people will need to implement `ParachainConsensus` for each consensus. Maybe I will provide an implementation of `ParachainConsensus` for `T` where `T: Slots`, but for that I will first need to have some examples :D

JoshOrndorff (commented):
I was a little late here, but I'll submit this anyway. Thanks for this PR.

client/consensus/relay-chain/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+//! permission-less consensus algorithm that doesn't require any staking or similar to join as a
+//! collator. In this algorithm the consensus is provided by the relay-chain. This works in the
+//! following way.
+//!
+//! 1. Each node that sees itself as a collator is free to build a parachain candidate.
+//!
+//! 2. This parachain candidate is send to the parachain validators that are part of the relay chain.
+//!
+//! 3. The parachain validators validate at most X different parachain candidates, where X is the
+//! total number of parachain validators.
```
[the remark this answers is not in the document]
JoshOrndorff: Mostly suggested this one because I worry that people who haven't read all the docs might think that "parachain validators" means collators.

bkchr:
@JoshOrndorff could you open a pr with your suggestions please?

JoshOrndorff:
Could you explain a little bit more about the `RelayChainConsensusBuilder`?

Why is it necessary? I don't quite understand the comment:

> /// Builds a [`RelayChainConsensus`] for a parachain. As this requires
/// a concrete relay chain client instance, the builder takes a [`polkadot_service::Client`]
/// that wraps this concrete instanace. By using [`polkadot_service::ExecuteWithClient`]
/// the builder gets access to this concrete instance.

I mostly understand the explanation, but I don't understand why the builder needs a concrete relay chain client instance. Is this pattern used somewhere else? This seems more complex than the analogous code in sc_consensus_aura

JoshOrndorff:
I guess this is the part that is really bending my brain:
```rust
	/// Build the relay chain consensus.
	fn build(self) -> Box<dyn ParachainConsensus<Block>> {
		self.relay_chain_client.clone().execute_with(self)
	}
```

bkchr:
This works around some missing type system features :P 

A polkadot client can either be Kusama/Polkadot/Rococo/Test and all use some small different types. Actually most of them are equal, but the compiler can not resolve this. To circumvent this problem on the polkadot side, there is this `Client` enum that combines all of them into one enum. Than we use this machinery here to get access to the real type inside this `Client` enum that implements all the required traits etc.

JoshOrndorff:
Thanks for explaining. I see the part you mean in Polkadot now.

https://github.com/paritytech/polkadot/blob/71d46a027be3a8435c9500491e9fa0ee2c6944b3/node/service/src/client.rs#L97-L163

PR #334, opened 2021-02-18
Author filter logic

JoshOrndorff:
## Not to be merged

This PR is **not intended to be merged**. Rather it is to share code that I think is worth discussion for three reasons.

1. It clearly demonstrates a sync issue that I reported in https://github.com/paritytech/cumulus/issues/335. It may help diagnose that issue.
2. It serves as an example of why I want to use the data in one inherent when checking another as I reported in https://github.com/paritytech/substrate/issues/8108.
3. I will eventually seek peer review on the parachain consensus system we're working on. (Although this is only the on-chain part, so it may be hard to review at this point.)

## Design

This PR introduces the minimal necessary "consensus" logic to demonstrate the points above.

### Author Inherent

It introduces a pallet called Author Inherent. This pallet allows block authors to insert an onchain identity that represents them. This pallet is minimal, so it does not involve signing, or verifying signatures, it just allows block authors to insert an onchain identity. To check this inherent, we use the trait `CanAuthor` which takes an id and returns bool.

### Author Filter
The second pallet, called Author Filter implements the `CanAuthor` trait. It compares against a pseudo-randomly (using `Randomness` trait) selected subset of an approved author set. Concretely, we set Alice and Bob as authors in genesis config, and we select one of them to be eligible at each height.

## How to run this?
This is very close to stock cumulus. The scripts in the repo may even work. This is how I ran two validators and two collators.

```bash
#Alice Validator
./polkadot-ce03f3715-approval-checking --chain rococo-local-raw.json --base-path /tmp/alice --alice

# Bob Validator
./polkadot-ce03f3715-approval-checking --chain rococo-local-raw.json --base-path /tmp/bob --bob --port 30334

# Alice Collator
./rococo-collator --base-path /tmp/collator-alice --collator --alice --parachain-id 1000 --port 30335 -- --chain rococo-local-raw.json --port 40335

# Bob Collator
./rococo-collator --base-path /tmp/collator-bob --collator --bob --port 30336 --parachain-id 1000 -- --chain rococo-local-raw.json --port 40336
```


pallets/author-filter/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+
+			for i in 0..num_eligible {
+				// A context identifier for grabbing the randomness. Consists of two parts
+				// - The constant string *b"filter" - to identify this pallet
+				// - The index `i` when we're selecting the ith eligible author
+				// Currently this has the weakness that the authors are based only on para-block
+				// height. This will be aleviated in the future by adding entropy from the relay
+				// chain inherent.
+				let subject: [u8; 7] = [b'f', b'i', b'l', b't', b'e', b'r', i as u8];
+				let randomness : sp_core::H256 = T::RandomnessSource::random(&subject);
```
JoshOrndorff, line 80: This is why I want to use data from the parachain inherent when checking the author inherent.

See how we do it in Moonbeam where we use the relay block height. The problem is this logic can't be used until block execution. https://github.com/PureStake/moonbeam/blob/2427474e054f5dea16bab141622aa72ded492190/pallets/author-filter/src/lib.rs#L82-L97

JoshOrndorff:
I forgot this was still open. Sorry about that. Objectives 1 and 2 are accomplished, and I have updated code on which I'll request review shortly, so I'm closing this.

PR #346, opened 2021-02-26
XCM handler: make send origin configurable.

### LINKED ISSUE #339: Configurable send origin in XCM handler
XCM handler module has dispatchable calls to send messages, but the origin must be root. https://github.com/paritytech/cumulus/blob/d8126cdc13c4c370fa1c360421843796c8411ca9/pallets/xcm-handler/src/lib.rs#L77-L78

It will be helpful to make the origin configurable, so that it can work with `pallet-collective`. For instance, config the send origin as `EnsureRootOrHalfCouncil`.

shaunxw:
Closes #339


bkchr (approved):

tolak:
Hi i noticed that this PR can solved the situation that passed origin account into xcm executor, it would be better if we cherry pick this PR to rococov1 branch case we actually use rococov1 branch as dependence  in our project

bkchr:
The rococo branch will be updated today for the Rococo restart. :)

PR #352, opened 2021-03-03
Question: How can I add a seal digest

JoshOrndorff:
I'm working toward having the consensus engine sign blocks. I'm stuck on how to add the seal digest. I've reduced my confusion down to this minimal example, in which I add a simple silly digest item which contains no real data.

I'm able to successfully add the digest item and get my node to import the blocks. I haven't written the verifier yet, although I believe I understand the principles. Before I get to that, I'm stuck because the relay chain validators can't import my blocks.

I guess my final question is: what should I return from the `produce_candidate` function?
* The complete block including the seal
* The original block as it came out of the runtime (with no seal)
* We should change the interface and return both. Both are needed later by the cumulus collator.

### DIFF
--- a/client/consensus/relay-chain/src/lib.rs
+++ b/client/consensus/relay-chain/src/lib.rs
@@ -203,8 +203,14 @@ where
 
 		let (header, extrinsics) = block.clone().deconstruct();
 
-		let mut block_import_params = BlockImportParams::new(BlockOrigin::Own, header);
-		block_import_params.body = Some(extrinsics);
+		// Add a silly test digest, just to get familiar with how it works
+		let test_digest = sp_runtime::generic::DigestItem::Seal(*b"test", Vec::new());
+
+		let mut block_import_params = BlockImportParams::new(BlockOrigin::Own, header.clone());
+
+		// Add the test digest to the block import params
+		block_import_params.post_digests.push(test_digest.clone());
+		block_import_params.body = Some(extrinsics.clone());
 		// Best block is determined by the relay chain.
 		block_import_params.fork_choice = Some(ForkChoiceStrategy::Custom(false));
 		block_import_params.storage_changes = Some(storage_changes);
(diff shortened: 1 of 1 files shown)


client/consensus/relay-chain/src/lib.rs
```diff
@@ -224,7 +230,25 @@ where
...
+
+
+		// My question is about this part. If I don't include this, my own node imports the block,
+		// but cannot retrieve the state, because it has the wrong block hash.
+		//
+		// But if I leave this in place, then the relay chain cannot validate the candidate.
+		// It fails with:
+		// 2021-03-03 15:58:54  panicked at 'assertion failed: `(left == right)`
+		// left: `1`,
+		// right: `0`: Number of digest items must match that calculated.', /home/joshy/.cargo/git/checkouts/substrate-7e08433d4c370a21/74d5612/frame/executive/src/lib.rs:421:9
```
bkchr, line 244: You need to remove the digest before importing the block. Check how Babe and AuRa are doing this.

JoshOrndorff: Thanks for taking a look. I understand that I need to remove the digest before executing the block. I'm not sure where that is expected to happen with the current parachain  consensus trait.

It's the relay chain validators that are giving this error when they execute the block. Should produce candidate return the block with the seal digest or without it?

JoshOrndorff: To say a little more. I know my verifier will have to pop the seal off in order for other parachain nodes to import this block. But validators won't even run my verifier. So the verifier alone will not solve this problem.

bkchr: https://github.com/paritytech/cumulus/blob/8c1f788aa34ea5213544f813903a347d33ce6d47/rococo-parachains/runtime/src/lib.rs#L441

https://github.com/paritytech/cumulus/blob/8c1f788aa34ea5213544f813903a347d33ce6d47/pallets/aura/src/lib.rs#L53-L99

;)

JoshOrndorff: :exploding_head: :pray:

JoshOrndorff:
Do you expect it will be a common pattern in parachains to move stuff that typically happened in the verifier into the runtime so the relay chain validators can do it too?

JoshOrndorff:
One more note. After using this trick, I'm seeing this error on the validators. Everything seems fine otherwise.
```
2021-03-04 14:34:18  Failed to write file with rename, lock path: /tmp/substrateQsUXoG/chains/rococo_local_testnet/db/wasmtime/modules/wasmtime-/g0dSHvsZ_8UxtkBv45KAg_lEDOU8XhwXftSNb2UBE8s.wip-atomic-write-mod, target path: /tmp/substrateQsUXoG/chains/rococo_local_testnet/db/wasmtime/modules/wasmtime-/g0dSHvsZ_8UxtkBv45KAg_lEDOU8XhwXftSNb2UBE8s, err: No such file or directory (os error 2)
```

PR #357, opened 2021-03-07
Support xcm local execution in xcm-handler.

shaunxw:
With xcm local execution support, modules like `xtokens` don't have to do workarounds and use xcm-executor to do the execution anymore.

As discussed with @shawntabrizi and @xlc, updating xcm-handler's `SendXcm` implementation can't achieve the same goal, unless adding a `origin` parameter to the `send_xcm` signature, which is not necessary IMO as in other cases this param is not needed:

```rust
fn send_xcm(origin: MultiLocation, dest: MultiLocation, msg: Xcm) -> Result<(), XcmError> {}
```

### DIFF
--- a/pallets/xcm-handler/src/lib.rs
+++ b/pallets/xcm-handler/src/lib.rs
@@ -27,12 +27,13 @@ use cumulus_primitives_core::{
 	DownwardMessageHandler, HrmpMessageHandler, HrmpMessageSender, InboundDownwardMessage,
 	InboundHrmpMessage, OutboundHrmpMessage, ParaId, UpwardMessageSender,
 };
-use frame_support::{decl_error, decl_event, decl_module, sp_runtime::traits::Hash, traits::EnsureOrigin};
+use frame_support::{decl_error, decl_event, decl_module, dispatch::DispatchResult, sp_runtime::traits::Hash, traits::EnsureOrigin};
 use sp_std::convert::{TryFrom, TryInto};
 use xcm::{
 	v0::{Error as XcmError, ExecuteXcm, Junction, MultiLocation, SendXcm, Xcm},
 	VersionedXcm,
 };
+use xcm_executor::traits::LocationConversion;
 
 pub trait Config: frame_system::Config {
 	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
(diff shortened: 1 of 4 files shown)


shawntabrizi:
How will you make use of this function? WIth tight coupling?

shaunxw:
I used some abstraction over the handler, firstly defined a trait in `orml-xcm-support`:

```rust
pub trait XcmHandler<AccountId> {
	fn execute_xcm(origin: AccountId, xcm: Xcm) -> DispatchResult;
}
```

and added a type in `Config` of `orml-xtokens`:

```rust
pub trait Config: frame_system::Config {
	type XcmHandler: XcmHandler<Self::AccountId>;
}
```

And in runtime integration, a handler struct to wrap the `XcmHandler`:

```rust
pub struct HandleXcm;
impl XcmHandlerT<AccountId> for HandleXcm {
	fn execute_xcm(origin: AccountId, xcm: Xcm) -> DispatchResult {
		XcmHandler::execute_xcm(origin, xcm)
	}
}

impl orml_xtokens::Config for Runtime {
	// -- snip --
	type XcmHandler = HandleXcm;
}
```

shawntabrizi (approved):

shaunxw:
The abstraction is a bit verbose, as I'm not sure if it's good to defined this handler trait in cumulus where it's not used. 

Probably the `origin` parameter could be changed to `MultiLocation`:

```rust
pub trait XcmHandler<AccountId> {
	fn handle_xcm(origin: MultiLocation, xcm: Xcm) -> DispatchResult;
}
```

In this way the trait would be more general. Both [`DownwardMessageHandler`](https://github.com/paritytech/cumulus/blob/b50848ea53344f239866512b68ed9fbd259cf1b4/pallets/xcm-handler/src/lib.rs#L110-L113) and [`HrmpMessageHandler`](https://github.com/paritytech/cumulus/blob/b50848ea53344f239866512b68ed9fbd259cf1b4/pallets/xcm-handler/src/lib.rs#L132-L135) impl in xcm handler can also call it to avoid repeat.

pallets/xcm-handler/src/lib.rs
```diff
@@ -101,6 +107,20 @@ decl_module! {
 	}
 }
 
+impl<T: Config> Module<T> {
+	pub fn execute_xcm(origin: T::AccountId, xcm: Xcm) -> DispatchResult {
```
bkchr, line 111: Some docs would be nice.

shaunxw: Docs added.

bkchr (approved):

bkchr:
@shaunxw done :)

Issue #335, opened 2021-02-18
Paranodes occasionally fail to import a block and don't try again


JoshOrndorff:
A parachain node will occasionally fail to import a parachain block and consequently not update its best block. When this happens the node does not try again to request the failed-import block from a peer even though it is accepted into the relay chain.

This issue happens fairly often (every 10 - 100 blocks) but it is not always obvious. For example in a single collator network, there is no other collator which could fail to import a block, so you won't see it. But in a multi-collator network you will.

The good news is that the node gets back on track once it imports the next parachain block after the one it failed to import. However, if the failed node is a collator, and the network is waiting for it to author, this represents a liveness stall.

Although the issue happens often, it is much easier to notice that it happened and diagnose it if it causes a network stall. To help with that I've opened PR #334. This introduces the minimal necessary author assignment logic necessary to make this bug cause a network stall.

bkchr:
> However, if the failed node is a collator, and the network is waiting for it to author, this represents a liveness stall.

This is your main problem. You need a way to advance to the next author if one doesn't build a block.

What does it mean it fails to import a block? Is there an error somewhere? Or it just doesn't import the block?

JoshOrndorff:
> You need a way to advance to the next author

I totally agree. We've already implemented this by including the relay chain block height as part of our author selection. But I opened #334 with this less robust scheme because this is how I found it most useful for reproducing the import error.

> What does it mean it fails to import a block? Is there an error somewhere? Or it just doesn't import the block?

No error. It just never prints the log saying it imported the block, and never updates its best block (even though its peers do).

bkchr:
Could you run the node with `sync=trace` until this happens and give me the log? :D

crystalin:
@bkchr 
I was able to have the same effect, but I'm not sure if it is the same issue (only few blocks got delayed here (2 or 3), while we observed sometime a delay of many blocks (like 5-10)).

The block height that got delayed was `53568..53570`
[moonbeam_slow_import.log](https://github.com/paritytech/cumulus/files/6047012/moonbeam_slow_import.log)

bkchr:
@crystalin which Substrate version are you using?

crystalin:
@bkchr 
https://github.com/paritytech/substrate#965950969f3fca2d9e225e4988afbc0a6b851a56
https://github.com/paritytech/polkadot#71a528e822af63b96bdb03efac9b4927fb11a25e
[...]

PR #363, opened 2021-03-11
Fix block propagation between non-collator nodes

bkchr:
There was an issue that non collator, actually all nodes that didn't produced a particular block, did not propagate this block to its peers. This pr fixes this by propagating every imported block that was not build by the local block to its peers. Actually this is a hack, until it is fixed in Substrate ;)

Fixes: https://github.com/paritytech/cumulus/issues/335

To make the tests work, the following pr in Substrate is required: https://github.com/paritytech/substrate/pull/8325

(I have executed the tests locally with the mentioned pr and everything is green)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1208,6 +1208,7 @@ dependencies = [
 name = "cumulus-client-network"
 version = "0.1.0"
 dependencies = [
+ "cumulus-primitives-core",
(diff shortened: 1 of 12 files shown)


crystalin:
Awesome work @bkchr 
We will include it in Moonbeam v7 and let you know if it improves our block propagation

PR #364, opened 2021-03-12
 Integration test for runtime upgrades

### LINKED ISSUE #360: Integration test for runtime upgrades
We need an integration test for runtime upgrades.

1. Use `schedule_upgrade` to schedule the upgrade. This will require that we bump the `spec_version`. We will need to add an `if` to this function https://github.com/paritytech/cumulus/blob/bkchr-update-yea-yeah-yeah/test/runtime/src/lib.rs#L298. Basically we want to check the storage for some value to be set (this value needs to be set at genesis). If this value is set, we return `VERSION` and otherwise we return the "custom" version where `spec_version` is bumped. This will make it possible to make the checks happy in `schedule_upgrade` without having to build the runtime twice.
2. There should a custom `OnRuntimeUpgrade` code for the runtime that sets some value which we will query to ensure that the upgrade actually happened.
3.
4. Profit

TLDR: Add a new integration test that schedules a runtime upgrade and we check that this runtime upgrade was executed.

cecton:
Closes #360


test/runtime/src/lib.rs
```diff
@@ -295,7 +297,14 @@ decl_runtime_apis! {
 impl_runtime_apis! {
 	impl sp_api::Core<Block> for Runtime {
 		fn version() -> RuntimeVersion {
-			VERSION
+			if VersionUpgrade::get() {
```
bkchr, line 300: This will not work. You need to use some value that is set at genesis and removed at the runtime upgrade.

And here you need to check if this value isn't set.

cecton: Ok I got it. Like this 1215f65 ?

test/service/src/lib.rs
```diff
@@ -545,4 +578,72 @@ impl TestNode {
...
+		function: impl Into<runtime::Call>,
+		caller: Sr25519Keyring,
+	) -> Result<RpcTransactionOutput, RpcTransactionError> {
+		let extrinsic = construct_extrinsic(&*self.client, function, caller);
+
+		self.rpc_handlers.send_transaction(extrinsic.into()).await
+	}
+
+	/// Register a parachain at this relay chain.
+	pub async fn schedule_upgrade(&self) -> Result<(), RpcTransactionError> {
```
bkchr, line 594: Please pass the wasm here as parameter.

test/service/Cargo.toml
```diff
@@ -48,6 +52,7 @@ cumulus-client-network = { path = "../../client/network" }
 cumulus-client-service = { path = "../../client/service" }
 cumulus-primitives-core = { path = "../../primitives/core" }
 cumulus-test-runtime = { path = "../runtime" }
+cumulus-test-runtime-upgrade = { path = "../runtime-upgrade" }
```
bkchr, line 55: Should be a dev dependency.

test/service/tests/runtime_upgrade.rs
```diff
@@ -0,0 +1,105 @@
...
+
+	while let Some(notification) = import_notification_stream.next().await {
+		if notification.is_new_best {
+			let res = charlie.client.runtime_api()
+				.has_upgraded(&BlockId::Hash(notification.hash));
+			if matches!(res, Ok(false)) {
+				break;
+			}
+		}
+	}
```
bkchr, line 84: Why do we need this?

test/service/tests/runtime_upgrade.rs
```diff
@@ -0,0 +1,105 @@
...
+
+	// schedule runtime upgrade
+	charlie.schedule_upgrade().await.unwrap();
+
+	while let Some(notification) = import_notification_stream.next().await {
+		if notification.is_new_best {
+			let res = charlie.client.runtime_api()
+				.has_upgraded(&BlockId::Hash(notification.hash));
+			if matches!(res, Ok(true)) {
```
bkchr, line 93:
```suggestion
			if res.unwrap_or(false) {
```

bkchr (approved):

test/service/src/lib.rs
cecton: Fixed in d83d1ef

test/service/Cargo.toml
cecton: Fixed in ae01e69

PR #373, opened 2021-03-19
Fix truly tiny typo in overview document

Byron:


bkchr:
I'm clearly no native speaker, however in Polkadot it is `PoV` everywhere, so I think `PoVBlock` is right :see_no_evil:

Byron:
In that case, the other occurrences of `PovBlock` could be turned into `PoVBlock` for consistency.

bkchr:
> In that case, the other occurrences of `PovBlock` could be turned into `PoVBlock` for consistency.

Sounds good! :)

bkchr (approved):

PR #378, opened 2021-03-26
add validation code hash in collation

thiolliere:
companion to polkadot https://github.com/paritytech/polkadot/pull/2593

* add `CODE_HASH` in well_known_keys to store the hash of the code.
* `CODE_HASH` on genesis config, and migration is provided in on_runtime_upgrade
* client fetch `CODE_HASH` value or query `CODE` if the value at `CODE_HASH` is empty

## TODO

* test with polkadot-launch

## NOTES

* The client will send collation with the validation_code_hash, thus the relay chain nodes must accept these collation with validation_code_hash or any collation will be invalid.
* The runtime of parachain doesn't really need to be upgraded. Even if the `CODE_HASH` key has no value in storage, the client will fetch the code from `CODE` and compute the hash.

### DIFF
--- a/client/collator/src/lib.rs
+++ b/client/collator/src/lib.rs
@@ -21,7 +21,7 @@ use cumulus_primitives_core::{
 	well_known_keys, OutboundHrmpMessage, ParachainBlockData, PersistedValidationData,
 };
 
-use sc_client_api::BlockBackend;
+use sc_client_api::{BlockBackend, StateBackend};
 use sp_consensus::BlockStatus;
 use sp_core::traits::SpawnNamed;
 use sp_runtime::{
@@ -44,9 +44,13 @@ use futures::{channel::oneshot, FutureExt};
 use std::sync::Arc;
 use parking_lot::Mutex;
 use tracing::Instrument;
+use lru::LruCache;
 
 /// The logging target.
 const LOG_TARGET: &str = "cumulus-collator";
+
+/// The size of lru cache for code hash
+const CODE_HASH_LRU_SIZE: usize = 8;
 
 /// The implementation of the Cumulus `Collator`.
 pub struct Collator<Block: BlockT, BS, Backend> {
@@ -54,6 +58,8 @@ pub struct Collator<Block: BlockT, BS, Backend> {
 	parachain_consensus: Box<dyn ParachainConsensus<Block>>,
 	wait_to_announce: Arc<Mutex<WaitToAnnounce<Block>>>,
 	backend: Arc<Backend>,
+	/// An LRU cache mapping code storage hash to validation code hash.
+	code_hash_lru: Arc<Mutex<LruCache<Block::Hash, Hash>>>,
 }
 
 impl<Block: BlockT, BS, Backend> Clone for Collator<Block, BS, Backend> {
@@ -63,6 +69,7 @@ impl<Block: BlockT, BS, Backend> Clone for Collator<Block, BS, Backend> {
 			wait_to_announce: self.wait_to_announce.clone(),
 			backend: self.backend.clone(),
 			parachain_consensus: self.parachain_consensus.clone(),
+			code_hash_lru: self.code_hash_lru.clone(),
 		}
 	}
 }
@@ -82,12 +89,14 @@ where
 		parachain_consensus: Box<dyn ParachainConsensus<Block>>,
 	) -> Self {
 		let wait_to_announce = Arc::new(Mutex::new(WaitToAnnounce::new(spawner, announce_block)));
+		let code_hash_lru = Arc::new(Mutex::new(LruCache::new(CODE_HASH_LRU_SIZE)));
 
 		Self {
 			block_status,
 			wait_to_announce,
 			backend,
 			parachain_consensus,
+			code_hash_lru,
 		}
 	}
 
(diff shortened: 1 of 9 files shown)


client/collator/src/lib.rs
```diff
@@ -241,6 +241,40 @@ where
...
+					Some(code) => ValidationCode(code),
+					None => {
+						tracing::error!(
+							target: LOG_TARGET,
+							"Failed to get validation code."
+						);
+						return None;
+					}
+				};
+				code.hash()
```
thiolliere, line 275: this branch when code is not found could be removed once all current runtimes have upgraded and put their code hash in storage.

bkchr (commented):
Looks good in general. As there is no production parachain out there, you can remove any migrations.

I'm also thinking if we need this in the state at all?

thiolliere:
without taking into account the cost of querying code from trie, I benchmarked blake2_256 for big input, result is 1.3ms per MB.

so 13ms for 10MB, so indeed it might not be really needed to put hash code in the trie.

bkchr:
Don't remove it yet, I will need to think a little bit about it :D

client/collator/src/lib.rs
```diff
@@ -241,6 +241,40 @@ where
 				}
 			};
 
+			let code_hash = sp_io::storage::get(well_known_keys::CODE_HASH);
```
thiolliere, line 244: IIRC this code is executed after the block is executed, thus the code is the code after runtime upgrade. Instead we need to fetch the code before runtime upgrade. So I need to find a way to get the code hash from the code used to build the code.

If not then I mistake somewhere else my manual test show that the produced candidate is considered invalid by polkadot client.

bkchr: > EDIT: I tried to fetch code from `header.parent_hash()` but it doesn't seem to work either

This should work.

bkchr:
So, I thought about this and got an idea.

1. Use `storage_hash(CODE)` to get the hash of the code.
2. Have some lru-cache that uses `Block::Hash -> BlakeTwo256Hash` mapping.
3. If the current code hash isn't known, insert the hash and compute the BlakeTwo256Hash of the code.

We should do it that way because the code hash is cached by Substrate automatically and thus, we don't need to always recalculate it.

client/collator/src/lib.rs
thiolliere: indeed I mixed up, I triggered a successful runtime upgrade.
I fail to trigger a second one, but I think this is because of some other constraint I'm not aware of.

rphmeier:
I also wonder, if this is the right place to add the validation code hash.

1. The `collation-generation` subsystem in Polkadot could also add this and we avoid any Cumulus change.
2. The receiving validator must check that this code hash is correctly set

bkchr:
But how would the `collation-generation` subsystem know the current validation code? Isn't the final goal of this feature to not have the validation code in the relay chain storage at all?

thiolliere:
I'll add a test so that code is correctly fetch from parent, the test will produce collation for a block with a transaction sudo set code, and test the validation_code_hash is the expected one.

I'll also do a final test with polkadot-launch.

But otherwise code is good

client/collator/src/lib.rs
```diff
@@ -159,6 +168,62 @@ where
...
+				return None;
+			},
+		};
+
+		let maybe_code_hash = self.code_hash_lru.lock().get(&code_storage_hash).cloned();
+
+		let code_hash = if let Some(code_hash) = maybe_code_hash {
+			code_hash
+		} else {
+			parent_state.inspect_state(|| {
```
bkchr, line 209: Here you can also just use `parent_state.storage(CODE)`

thiolliere:
I struggle to create an enviroment which compiles, I'll test in a branch with diener on thursday.

(for info, I tried multiple stuff, using diener to point to a branch in polkadot, this fails because it takes an outdated substrate, then I tried to update substrate it fails to peek a version for fs-swap, if I try to to update on fs-swap it fails for another crates which says that a features doesn't exists.
Then I tried to have a branch on substrate/polkadot/cumulus, having polkadot pointing the substrate branch and cumulus pointing to the polkadot and the substrate branch. it failed too.
Then I tried just a regular cargo update on everything, fails too.
Maybe I should try to use diener patch instead of diener update. I continue on thursday)

rphmeier:
> as polkadot has no mecanism to fetch the code from storage. putting it in polkadot would require mecanism for polkadot to get the code in some way which sounds less ideal

The `RuntimeApiSubsystem` should expose this, no?

rphmeier:
> Isn't the final goal of this feature to not have the validation code in the relay chain storage at all

I believe we'd have validation code hash or perhaps `validation_code_erasure_root` in storage even if code was entirely removed from the relay chain

thiolliere:
OK I try now to get the code hash from the relay chain instead of having it in the Collation.
This way it shouldn't require any change from cumulus and we can close this PR.

The commit is https://github.com/paritytech/polkadot/pull/2593/commits/0db1f2c53a5f31dc2187dd1022d8b6009fd2d51e but it is not working when I tried with the changes on top of rococo. https://github.com/paritytech/polkadot/pull/2593#issuecomment-811998724

thiolliere:
We now fetch from relay chain to get the validation code hash in the collation generation in polkadot.

Thus this PR is no longer needed. Even though current implementation in polkadot is not optimized, fetching the code hash from relay chain should be relatively cheap.

Issue #368, opened 2021-03-16
`--collator` does not imply `--validator`


bkchr:
Both should be doing the same, but `--collator` is not enabling the same features as `--validator`.

Maybe we should just drop `--collator` all together.

JoshOrndorff:
I vote to keep it. We use it at Moonbeam. Of course we could still have it there even if you don't have it here I suppose. I can take a look at this next week.

bkchr:
The `--collator` flag is not provided by Cumulus in a "reusable" way. Meaning that you need to copy the code anyway and could keep it or do whatever you want :D

PR #380, opened 2021-03-29
Make `--collator` reusable and imply `--validator`

JoshOrndorff:
This is my attempt at solving #368 

This addresses the primary concern about the `--collator` flag not being equivalent to the `--validator` flag, as well as  the part about the collator flag not being reusable. With this PR, parachain projects can import the cumulus RunCmd which includes the collator flag.

I was really hoping structopt would allow me to simply add an annotation `#[structopt(implies = validator)]` but it seems like there is no notion of "implies" in structopt.


JoshOrndorff:
@cecton Feedback on my use of structopt here is very welcome.

client/cli/src/lib.rs
```diff
@@ -112,3 +112,37 @@ impl sc_cli::CliConfiguration for PurgeChainCmd {
...
+	fn role(&self, is_dev: bool) -> sc_cli::Result<sc_cli::Role> {
+		let is_authority = self.base.validator || self.collator || is_dev;
+
+		Ok(if is_authority {
+			sc_service::Role::Authority
+		} else {
+			sc_service::Role::Full
+		})
+	}
+}
```
cecton, line 148: I'm afraid this won't be enough. You need to call every method that [the original `RunCmd` implements](https://github.com/paritytech/substrate/blob/master/client/cli/src/commands/run_cmd.rs#L269-L445). Just like you did for `shared_params`.

You can get inspiration from this code we do for the relay chain: https://github.com/paritytech/cumulus/blob/master/rococo-parachains/src/command.rs#L349-L457

rococo-parachains/src/cli.rs
```diff
@@ -88,24 +88,6 @@ pub struct ExportGenesisWasmCommand {
...
-	pub parachain_id: Option<u32>,
-}
-
-impl std::ops::Deref for RunCmd {
-	type Target = sc_cli::RunCmd;
-
-	fn deref(&self) -> &Self::Target {
-		&self.base
-	}
-}
```
cecton, line 107: See how the previous implementation circumvent the need of calling the "base" `RunCmd` by using `Deref`. Unfortunately you can't do the same because you want to override the `fn role()`.

cecton (changes requested):

rococo-parachains/src/cli.rs
JoshOrndorff: Ohh, I wondered what the point of this `Deref` was. I see how it was useful now.

cecton: It was a bit lazy :sweat:

JoshOrndorff:
I've filled in the rest of the methods now. I'm glad you pointed it out too because there were two others that needed to be overridden. I have one more question for you, and it is about the helper function `rpc_interface`. I copied it from `sc_cli`, but I wonder if it should be made public there. What do you think?

cecton:
>  I copied it from `sc_cli`, but I wonder if it should be made public there. What do you think?

Normally I would say it wouldn't be bad but this is so "lonely" that probably nobody would notice this is here and it is useful...

... :thinking: ... maybe if we can move it to some object it would feel better as part of the API... like an method on an instance of RpcMethods?

```
fn choose_rpc_interface(&self, is_external, is_unsafe, is_validator)
-> sc_cli::Result<IpAddr> { ... }
```

What do you think @JoshOrndorff

JoshOrndorff:
Are we waiting on paritytech/substrate#8505 before reviewing this one? Or shall we merge this one first and refactor it after?

client/cli/src/lib.rs
```diff
@@ -112,3 +118,161 @@ impl sc_cli::CliConfiguration for PurgeChainCmd {
...
+	}
+
+	if is_external || is_unsafe_external {
+		// if rpc_methods == RpcMethods::Unsafe {
+		// 	log::warn!(
+		// 		"It isn't safe to expose RPC publicly without a proxy server that filters \
+		// 	available set of RPC methods."
+		// 	);
+		// }
+
```
cecton, line 273: Remove (or uncomment) commented code?

cecton (approved):
Just a small knit to fix but otherwise I think we can merge for now and we will update later if the other PR gets merged at some point

bkchr (changes requested):
Sorry for the later answer, however it took some time until I found some solution I'm happy with.

The current implementation is not making me happy, especially as you have manually overwritten some of the methods to respect `collator`.

I think we should do the following way:
```
/// The `run` command used to run a node.
#[derive(Debug, StructOpt)]
pub struct RunCmd {
	/// Our run command inherents from sc_cli's
	#[structopt(flatten)]
	pub base: sc_cli::RunCmd,
	/// Id of the parachain this collator collates for.
	#[structopt(long)]
	pub parachain_id: Option<u32>,
	/// Run node as collator.
	///
	#[structopt(long, conflicts_with = "validator")]
	pub collator: bool,
}

// as you already have it. But we don't implement `CliConfiguration` for this struct.
// CliConfiguration should be implemented for some other struct that we transform `RunCmd` into.

impl RunCmd {
    /// Find better name ;P
    fn transform(self) -> Transformed {
         self.base.validator = self.base.validator || self.collator;
         // Transformed should implement `CliConfiguration`, but this time we can forward everything to the `base` cmd.
         Transformed { self.base, self.parachain_id }
    }
}
```

bkchr:
ping @JoshOrndorff

JoshOrndorff:
Thanks for your review. I understand what you mean.

I just haven't made time to get back to this. I can take another look by next week.

bkchr:
Ty

JoshOrndorff:
I've taken an attempt at the design change you recommended. I'm not quite sure where you want the transformation / normalization to happen though.

I wonder if the example code you gave is really supposed to take ownership of the original RunCmd?
```
impl RunCmd {
    /// Find better name ;P
    fn transform(self) -> Transformed {
         self.base.validator = self.base.validator || self.collator;
         // Transformed should implement `CliConfiguration`, but this time we can forward everything to the `base` cmd.
         Transformed { self.base, self.parachain_id }
    }
}
```

I've tried it this way and also with only borrowing, and I've hit a different error either way. The code explains it in context.Feedback welcome.

client/cli/src/lib.rs
```diff
@@ -112,3 +118,161 @@ impl sc_cli::CliConfiguration for PurgeChainCmd {
...
+	pub fn normalize(&self) -> NormalizedRunCmd {
+		let mut new_base = self.base.clone();
+
+		 new_base.validator = self.base.validator || self.collator;
+
+		 NormalizedRunCmd { 
+			 base: new_base,
+			 parachain_id: self.parachain_id,
+		}
+	}
```
bkchr, line 164: https://github.com/paritytech/substrate/pull/8731

using this we can make use of the self reference

JoshOrndorff:
This PR now also updates Cumulus's Substrate dependency by three commits to include @bkchr's work from paritytech/substrate#8731

These are the three commits. Two of them are harmless and one is the change I want, so I think this should be fine.

image

client/cli/src/lib.rs
```diff
@@ -112,3 +118,135 @@ impl sc_cli::CliConfiguration for PurgeChainCmd {
...
+pub struct NormalizedRunCmd {
+	/// The cumulus RunCmd inherents from sc_cli's
+	pub base: sc_cli::RunCmd,
+	/// Id of the parachain this collator collates for.
+	pub parachain_id: Option<u32>,
+}
+
+impl RunCmd {
+	/// Transform this RunCmd into a NormalizedRunCmd which does not have a redundant `collator` field.
+	pub fn normalize(&self) -> NormalizedRunCmd {
```
bkchr, line 152:
```suggestion
	/// Transform this RunCmd into a NormalizedRunCmd which does not have a redundant `collator` field.
	pub fn normalize(&self) -> NormalizedRunCmd {
```

Do you maybe have a better idea for the function name?

And the docs should also be updated

client/cli/src/lib.rs
```diff
@@ -112,3 +118,135 @@ impl sc_cli::CliConfiguration for PurgeChainCmd {
...
+		// TODO This clone issue should be fixed by https://github.com/paritytech/substrate/pull/8731 right?
```
bkchr, line 153:
```suggestion
```

client/cli/src/lib.rs
JoshOrndorff: Do you not like `normalize`? I like it because it indicates that we aren't really changing any of the data, just writing it in a more standard form. `standardize` would also be fine. I like both of those better than the original `transform`.

I'm fine with any, of them, but I can't spend too long thinking about it because I have a bike shed I have to paint today ;-)

JoshOrndorff: I updated the docs as you requested.

bkchr (approved):

client/cli/src/lib.rs
```diff
@@ -112,3 +118,134 @@ impl sc_cli::CliConfiguration for PurgeChainCmd {
...
+/// This is how we make `--collator` imply `--validator`.
+pub struct NormalizedRunCmd {
+	/// The cumulus RunCmd inherents from sc_cli's
+	pub base: sc_cli::RunCmd,
+	/// Id of the parachain this collator collates for.
+	pub parachain_id: Option<u32>,
+}
+
+impl RunCmd {
+	/// Create a NormalizedRunCmd which does not have a redundant `collator` field.
```
bkchr, line 151:
```suggestion
	/// Create a [`NormalizedRunCmd`] which merges the `collator` cli argument into `validator` to have only one.
```

client/cli/src/lib.rs
bkchr: Fine

client/cli/src/lib.rs
bkchr: One last call @JoshOrndorff. Does this sounds reasonable in english? :D

JoshOrndorff: Sounds great.

bkchr:
bot merge

Issue #374, opened 2021-03-23
Bring back `can_set_code`


bkchr:
`can_set_code` is used in `schedule_upgrade` in `parachain-system`. It was removed because the current implementation is too slow for parachains.

JoshOrndorff:
Since this is closed, it means we can "upgrade" to a lower spec version or different spec name and the upgrade should still go through right?

bkchr:
No

JoshOrndorff:
Oh actually, now that the normal upgrade path is with `system::set_code` we are already using the normal checks from https://github.com/paritytech/substrate/blob/master/frame/system/src/lib.rs#L1537-L1557

Are those the ones that were too heavy?

bkchr:
Before we instantiated the WASM and called into it. Now, we just can fetch the version from a WASM section. This is much faster and better.

PR #383, opened 2021-03-30
Improve Storage and Add `set_upgrade_block` to Validation Function Upgrade

shawntabrizi:
The first change in this PR is to the underlying storage structure of how Validation Function Upgrades are handled.

Before, we had a single storage item which has both the block when the upgrade would happen, and the whole wasm we would upgrade with. We would then check every block if it was the right block to upgrade on, reading the whole wasm each time!

Here we split storage into two storage items, one where we can more quickly check the block number being correct, and the other where only after we confirm we need the new validation function, we get it.

----

The second change allows Root to set the specific block where they expect the relay chain to upgrade their wasm.

This is useful if you want to coordinate with the relay chain for some different upgrade schedule than would be set up by default.

Lots of warnings here on using this function if you don't know what you are doing...

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -80,7 +80,7 @@ decl_storage! {
 		/// We need to store the new validation function for the span between
 		/// setting it and applying it. If it has a
 		/// value, then [`PendingValidationFunction`] must have a real value, and
-		/// together will coordinate what block number where the upgrade will happen.
+		/// together will coordinate the block number where the upgrade will happen.
 		PendingRelayChainBlockNumber: Option<RelayChainBlockNumber>;
 
 		/// The new validation function we will upgrade to when the relay chain
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -78,14 +78,14 @@ pub trait Config: frame_system::Config {
 decl_storage! {
 	trait Store for Module<T: Config> as ParachainSystem {
 		/// We need to store the new validation function for the span between
-		/// setting it and applying it. If `PendingRelayChainBlockNumber` has a
-		/// value, then `PendingValidationFunction` must have a real value, and
+		/// setting it and applying it. If it has a
+		/// value, then [`PendingValidationFunction`] must have a real value, and
 		/// together will coordinate what block number where the upgrade will happen.
 		PendingRelayChainBlockNumber: Option<RelayChainBlockNumber>;
 
 		/// The new validation function we will upgrade to when the relay chain
-		/// reaches `PendingRelayChainBlockNumber`. A real validation function must
-		/// exist here as long as `PendingRelayChainBlockNumber` is set.
+		/// reaches [`PendingRelayChainBlockNumber`]. A real validation function must
+		/// exist here as long as [`PendingRelayChainBlockNumber`] is set.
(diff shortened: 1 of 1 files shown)


pallets/parachain-system/src/lib.rs
```diff
@@ -155,6 +155,19 @@ decl_module! {
 			Self::schedule_upgrade_impl(validation_function)?;
```
shawntabrizi, line 155: Whats going on up here?

`schedule_upgrade` and `schedule_upgrade_without_checks` are exactly the same?

pallets/parachain-system/src/lib.rs
```diff
@@ -155,6 +155,19 @@ decl_module! {
...
+		/// Force an already scheduled validation function upgrade to happen on a particular block.
+		///
+		/// Note that coordinating this block for the upgrade has to happen independently on the relay
+		/// chain and this parachain. Synchronizing the block for the upgrade is sensitive, and this
+		/// bypasses all checks and and normal protocols. Very easy to brick your chain if done wrong.
+		#[weight = (0, DispatchClass::Operational)]
+		pub fn set_upgrade_block(origin, relay_chain_block: RelayChainBlockNumber) {
+			ensure_root(origin)?;
+			if let Some((_, validation_function)) = PendingValidationFunction::get() {
+				PendingValidationFunction::put((relay_chain_block, validation_function));
```
shawntabrizi, line 167: Note that this storage pattern is really inefficient

I see in the inherent below that we read this storage item every block just to look at the `relay_chain_block`.

We def want to split this into two storage items or a map so that we don't read the entire validation function every block just to check the upgrade block.

Am I missing anything @pepyakin ?

pallets/parachain-system/src/lib.rs
bkchr: https://github.com/paritytech/cumulus/issues/374

pallets/parachain-system/src/lib.rs
```diff
@@ -155,6 +155,19 @@ decl_module! {
...
 
+		/// Force an already scheduled validation function upgrade to happen on a particular block.
+		///
+		/// Note that coordinating this block for the upgrade has to happen independently on the relay
+		/// chain and this parachain. Synchronizing the block for the upgrade is sensitive, and this
+		/// bypasses all checks and and normal protocols. Very easy to brick your chain if done wrong.
+		#[weight = (0, DispatchClass::Operational)]
+		pub fn set_upgrade_block(origin, relay_chain_block: RelayChainBlockNumber) {
+			ensure_root(origin)?;
+			if let Some((_, validation_function)) = PendingValidationFunction::get() {
```
bkchr, line 166: Should we not return an error if this is not set?

pallets/parachain-system/src/lib.rs
bkchr: Good point. AFAIK this was not introduced by Sergei. 

But we should fix this.

bkchr (approved):

pallets/parachain-system/src/lib.rs
shawntabrizi: yeah can do it

pallets/parachain-system/src/lib.rs
shawntabrizi: I will add it to this PR

PR #386, opened 2021-04-01
Ping & XCMP message aggregation & rewrite

gavofyork:
Introduces the ping pallet for pinging between chains.

Also includes XCMP message dispatch system reimagining:
- XCM messages now get aggregated into an XCMP message before being sent out
- Similarly, XCMP messages get split into XCM messages before being executed
- Moved most of the logic into xcm-handler pallet
- Altered the outgoing XCMP API from push to pull
- Changed underlying outgoing queue data structures to avoid multi-page read/writes
- Introduced queuing for incoming messages
- Introduced signal messages as a flow-control sub-stream
- Introduced flow-control with basic threshold back-pressure
- Introduced overall weight limitation on messages executed
- Additonal alterations to XCM APIs for the new system

TODO:
- [ ] DMP queuing - messages should be queued and executed as per weight limits.
- [ ] DMP origins - AccountId origins from Relay should have sensible origins on Paras when sent over DMP.
- [ ] Implement proper weight pricing - it should integrate into the chain's usual transaction weight pricing.
- [ ] Implement proper weighing for XCM queueing and dispatch - right now it's basically just ignored.
- [ ] Get block-builder to provide feedback to the XCM system so that the PoV block doesn't become too big when squeezing in messages in `on_idle`.
- [ ] Figure out how to manage origin permissions. We want root and governance Relay-chain XCM origins to have free execution on the Statemint parachain, but not just random accounts that happen to be intructing the Relay-chain to send an XCM on their behalf to the parachain.

Things to consider:
- Message prioritisation/QoS. Currently all messages in a channel must be executed in strict order of being sent. If the receiving system runs out of weight while processing due to a heavy message, then other channels will be tried, even if there are lighter messsages following it which could reasonably be executed. For now this is fine because a) we have an optimised data system for taking only contiguous messages from the beginning of the page stored in state and b) during execution it gives other channels with lighter messages a chance to execute.
- Scheduling. Right now we process all channels' messages, one page at a time in a random-shuffled round-robin, incrementally making more weight available until we have made all weight available and run out of weight to process the next message on every channel. This *works* and is *safe*, but might lead to slightly unexpected stuff happening such as mesages sent much later on a lightly-used channel executing before messages sent on a heavier-used channel.


pallets/parachain-system/src/lib.rs
```diff
@@ -712,24 +764,8 @@ impl MessageQueueChain {
...
-pub enum SendHorizontalErr {
-	/// The message sent is too big.
-	TooBig,
-	/// There is no channel to the specified destination.
-	NoChannel,
-}
-
 impl<T: Config> Module<T> {
-	pub fn send_upward_message(message: UpwardMessage) -> Result<(), SendUpErr> {
+	pub fn send_upward_message(message: UpwardMessage) -> Result<u32, MessageSendError> {
```
shawntabrizi, line 768: What does the u32 represent?

rococo-parachains/pallets/ping/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+		type Origin: From<<Self as SystemConfig>::Origin> + Into<Result<CumulusOrigin, <Self as Config>::Origin>>;
+
+		/// The overarching call type; we assume sibling chains use the same type.
+		type Call: From<Call<Self>> + Encode;
+
+		type XcmSender: SendXcm;
+	}
+
+	#[pallet::storage]
+	/// Details of an asset.
```
kianenigma, line 54:
```suggestion
	/// Vector of `ParaId` targets to spam.
```

rococo-parachains/pallets/ping/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+	pub(super) type Targets<T: Config> = StorageValue<
+		_,
+		Vec<(ParaId, Vec<u8>)>,
+		ValueQuery,
+	>;
+
+	#[pallet::storage]
+	/// Details of an asset.
```
kianenigma, line 62:
```suggestion
	/// Number of pings to send.
```

rococo-parachains/pallets/ping/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+	pub(super) type PingCount<T: Config> = StorageValue<
+		_,
+		u32,
+		ValueQuery,
+	>;
+
+	#[pallet::storage]
+	/// Details of an asset.
```
kianenigma, line 70:
```suggestion
	/// Record of sent pings.
```

(not sure what it is yet)

PR #392, opened 2021-04-07
Two-stage upgrade for parachains

gavofyork:
Authorize an upgrade via XCM/DMP and then actually pass in the code via a normal transaction.

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -32,7 +32,7 @@ use sp_runtime::traits::{BlakeTwo256, Hash};
 use sp_inherents::{InherentData, InherentIdentifier, ProvideInherent};
 use frame_support::{
 	decl_error, decl_event, decl_module, decl_storage,
-	dispatch::{DispatchResult, DispatchError, DispatchResultWithPostInfo},
+	dispatch::{DispatchResult, DispatchError, DispatchResultWithPostInfo, PaysFee},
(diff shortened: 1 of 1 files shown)


shawntabrizi (approved):

shawntabrizi:
cc @kianenigma @joepetrowski @JesseAbram

pallets/parachain-system/src/lib.rs
```diff
@@ -235,10 +238,24 @@ decl_module! {
...
 			ensure_root(origin)?;
-			let _ = Self::send_hrmp_message(message);
+
+			AuthorizedUpgrade::<T>::put(&code_hash);
+
+			Self::deposit_event(RawEvent::UpgradeAuthorized(code_hash));
+		}
+
+		#[weight = 1_000_000]
+		fn enact_authorized_upgrade(origin, code: Vec<u8>) {
```
bkchr, line 251:
```suggestion
		fn enact_authorized_upgrade(_origin, code: Vec<u8>) {
```

Should we maybe pay back the fees if the function was successful?

bkchr (approved):

kianenigma (approved):

pallets/parachain-system/src/lib.rs
shawntabrizi:
```
-> DispatchResultWithPostInfo
```

and

```
Ok(Pays::No.into())
```

PR #393, opened 2021-04-07
Do not accept `--parachain-id` and `--chain` together

bkchr:


JoshOrndorff:
This is because the parachain id is supposed to come from the spec, right?

bkchr:
`--parachain-id` is more like a hack anyway. I don't see why any real parachain should support different parachain ids from the CLI level.

The genesis state is just the genesis header and if you pass `--chain`, we will open the spec to get the header.

The parachain id is only required to generate the spec, when we don't use any named `--chain`, because the parachain id is passed to the genesis config.

JoshOrndorff:
> The parachain id is only required to generate the spec

Should it be associated with the `build-spec` subcommand then?

bkchr:
> > The parachain id is only required to generate the spec
> 
> Should it be associated with the `build-spec` subcommand then?

Depends on your chain.

In the example chain here, yes it could be added, but I don't see the need.

rococo-parachains/src/cli.rs
```diff
@@ -60,15 +60,15 @@ pub struct ExportGenesisStateCommand {
 	pub output: Option<PathBuf>,
 
 	/// Id of the parachain this state is for.
-	#[structopt(long, default_value = "100")]
+	#[structopt(long, default_value = "100", conflicts_with = "chain")]
```
jasl, line 63: `parachain_id` has default value which means you can't pass `chain` arg because `parachain_id` always has value

bkchr: Does it complain at runtime?

jasl: I don't know what your means, but you can simply verify what I said

```
jasl@localhost:~/Workspaces/BlockChain/cumulus on master$ target/release/rococo-collator export-genesis-state --chain tick
error: The argument '--chain <chain>' cannot be used with '--parachain-id <parachain-id>'

USAGE:
    rococo-collator export-genesis-state --chain <chain> --parachain-id <parachain-id>

For more information try --help
```

bkchr: Yeah that is at runtime. Hmmm. That is bad.

PR #407, opened 2021-04-16
Integrate a governance XCM origin 

gavofyork:


rococo-parachains/runtime/src/lib.rs
```diff
@@ -302,6 +303,8 @@ pub type XcmOriginToTransactDispatchOrigin = (
 	// Native signed account converter; this just converts an `AccountId32` origin into a normal
 	// `Origin::Signed` origin of the same 32-byte value.
 	SignedAccountId32AsNative<RococoNetwork, Origin>,
+	// Xcm origins can be represented natively under the Xcm pallet's Xcm origin.
+	XcmPassthough<Origin>,
```
apopiak, line 307:
```suggestion
	XcmPassthrough<Origin>,
```

rococo-parachains/runtime/src/lib.rs
```diff
@@ -64,6 +64,7 @@ use xcm_builder::{
 	AllowUnpaidExecutionFrom, ParentAsSuperuser,
 };
 use xcm_executor::{Config, XcmExecutor};
+use pallet_xcm::{XcmPassthough, EnsureXcm, IsMajorityOfBody};
```
apopiak, line 67:
```suggestion
use pallet_xcm::{XcmPassthrough, EnsureXcm, IsMajorityOfBody};
```

rococo-parachains/runtime/Cargo.toml
```diff
@@ -27,6 +27,7 @@ sp-inherents = { git = "https://github.com/paritytech/substrate", default-featur
 frame-support = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 frame-executive = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 frame-system = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
+pallet-assets = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
```
shawntabrizi, line 30: std import

rococo-parachains/runtime/src/lib.rs
```diff
@@ -376,6 +379,34 @@ impl cumulus_ping::Config for Runtime {
...
 }
 
+parameter_types! {
+	pub const AssetDeposit: Balance = 100;
+	pub const ApprovalDeposit: Balance = 1;
+	pub const StringLimit: u32 = 50;
+	pub const MetadataDepositBase: Balance = 10;
+	pub const MetadataDepositPerByte: Balance = 1;
+	pub const UnitBody: BodyId = BodyId::Unit;
+}
```
shawntabrizi, line 389: Even for Rococo, probably not the right units? Should at least use some kind of `DOLLARS` and stuff.

apopiak:
So when we try something similar in statemint the message doesn't seem to be handled correctly (I'm still figuring out how to debug this better.), but I have a guess at where this is failing: Cumulus handles downward messages [here](https://github.com/paritytech/cumulus/blob/689cb16660c6f9ea956952a365ea621b778a52f8/primitives/utility/src/lib.rs#L71) by giving them the `Parent` origin and I'm wondering whether how that interferes with the origin Passthrough.

pallets/parachain-system/src/lib.rs
```diff
@@ -485,25 +485,38 @@ impl<T: Config> Module<T> {
...
+			Self::deposit_event(RawEvent::DownwardMessagesProcessed(
+				processed_count,
+				weight_used,
+				result_mqc_head.clone(),
+				expected_dmq_mqc_head.clone(),
+			));
+
+			// After hashing each message in the message queue chain submitted by the collator, we should
+			// arrive to the MQC head provided by the relay chain.
+			assert_eq!(result_mqc_head, expected_dmq_mqc_head);
```
shawntabrizi, line 516: we could put this above the deposit event and then avoid the clone?

shawntabrizi: Also we want this panic?

pallets/xcm/src/lib.rs
```diff
@@ -36,16 +40,68 @@ pub mod pallet {
...
+	}
+
+	/// For an incoming downward message, this just adapts an XCM executor and executes DMP messages
+	/// immediately up until some `MaxWeight` at which point it errors. Their origin is asserted to be
+	/// the Parent location.
+	impl<T: Config> DownwardMessageHandler for Pallet<T> {
+		fn handle_downward_message(msg: InboundDownwardMessage) -> Weight {
+			let id = sp_io::hashing::twox_64(&msg.msg[..]);
+			let msg = VersionedXcm::<T::Call>::decode(&mut &msg.msg[..])
+				.map(Xcm::<T::Call>::try_from);
```
apopiak, line 85: the weight for this is left out intentionally, I assume?

pallets/parachain-system/src/lib.rs
gavofyork: yeah - it was relying on an indirect panic before. this just makes it explicit.

pallets/xcm/src/lib.rs
gavofyork: correct. the weight returned is the unknowable weight. stuff for bookkeeping, dispatch, decode and conversion can be benchmarked normally.

gavofyork:
@apopiak @shawntabrizi approvals?

pallets/parachain-system/src/lib.rs
gavofyork: also done in this PR: https://github.com/paritytech/cumulus/pull/413

apopiak (approved):

PR #413, opened 2021-04-27
Panic instead of ensure

### LINKED ISSUE #410: Clean up the parachain-system's ensures
Switch from `ensure` to `panic`s in the parachain-system pallet's `set_validation_data` inherent.

pepyakin:
Instead of `ensure` with dedicated errors use `panic` or `assert`. See
for details #410

Closes #410

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -32,7 +32,7 @@ use sp_runtime::traits::{BlakeTwo256, Hash};
 use sp_inherents::{InherentData, InherentIdentifier, ProvideInherent};
 use frame_support::{
 	decl_error, decl_event, decl_module, decl_storage,
-	dispatch::{DispatchResult, DispatchError, DispatchResultWithPostInfo},
+	dispatch::{DispatchResult, DispatchResultWithPostInfo},
 	ensure, storage,
 	traits::Get,
 	weights::{DispatchClass, Weight, PostDispatchInfo, Pays},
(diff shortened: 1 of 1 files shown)


pallets/parachain-system/src/lib.rs
```diff
@@ -500,34 +504,41 @@ impl<T: Config> Module<T> {
...
 		horizontal_messages: BTreeMap<ParaId, Vec<InboundHrmpMessage>>,
-	) -> Result<Weight, DispatchError> {
+	) -> Weight {
 		// First, check that all submitted messages are sent from channels that exist. The channel
 		// exists if its MQC head is present in `vfp.hrmp_mqc_heads`.
 		for sender in horizontal_messages.keys() {
-			ensure!(
+			// A violation of the assertion below indicates that one of the messages submitted by
+			// the collator was sent from a sender that doesn't have a channel opened to this parachain,
+			// according to the relay-parent state.
```
thiolliere, line 537: maybe you can put this in the assert message, but it is fine here too

pallets/parachain-system/src/lib.rs
```diff
@@ -480,10 +481,13 @@ impl<T: Config> Module<T> {
 	///
 	/// Checks if the sequence of the messages is valid, dispatches them and communicates the number
 	/// of processed messages to the collator via a storage update.
+	///
+	/// **Panics** if it turns out that after processing all messages the Message Queue Chain hash
+	///            doesn't match the expected.
```
thiolliere, line 486: maybe better to panic in the call itself and return a result here, but anyway it is ok to me.

thiolliere (approved):
code is good to me

pallets/parachain-system/src/lib.rs
pepyakin: I deliberately avoid putting such messages in to minimize the runtime overhead. My logic is that it is typically easy to find the location of the assert and thus find the comment above.

pallets/parachain-system/src/lib.rs
pepyakin: Yeah, I thought about that but that doesn't feel to give lot of benifit. I am not against the change and happy to make it if you insist though.

bkchr (approved):

PR #414, opened 2021-04-27
Integrate Shell runtime into collator

gavofyork:


rococo-parachains/src/command.rs
```diff
@@ -146,6 +146,31 @@ fn extract_genesis_wasm(chain_spec: &Box<dyn sc_service::ChainSpec>) -> Result<V
 		.ok_or_else(|| "Could not find wasm file in genesis state!".into())
 }
 
+fn use_shell_runtime(chain_spec: &Box<dyn ChainSpec>) -> bool {
+	chain_spec.id().starts_with("track") || chain_spec.id().starts_with("shell")
```
arkpar, line 150: "track" also gets the new runtime. Is that intended?

gavofyork: yeah - i'm assuming we probably want to have one of the three use this for testing.

arkpar (approved):

PR #421, opened 2021-04-29
Strip down Shell to bare minimum

gavofyork:
Removes:
- Balances
- Signed Transaction support
- Collective Coin Flip randomness (parent_hash is good enough for shell)
- Timestamp
- Session stuff left in
- XCMP & UMP
- Support for accounts in XCM
- Polkadot XCM module (send, execute & Relay/Para origins)

Adds:
- Unsigned transaction support for enacting an authorized upgrade


rococo-parachains/shell-runtime/src/lib.rs
```diff
@@ -389,14 +287,7 @@ pub type SignedBlock = generic::SignedBlock<Block>;
...
 /// The SignedExtension to the basic transaction logic.
-pub type SignedExtra = (
-	frame_system::CheckSpecVersion<Runtime>,
-	frame_system::CheckGenesis<Runtime>,
-	frame_system::CheckEra<Runtime>,
-	frame_system::CheckNonce<Runtime>,
-	frame_system::CheckWeight<Runtime>,
-	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
-);
+pub type SignedExtra = DisallowSigned;
```
kianenigma, line 290: I'd say it is less error-prone in the future to leave the system extensions and just prepend the list with the new barrier.

kianenigma (approved):

rococo-parachains/shell-runtime/src/lib.rs
gavofyork: why reference intentially unused logic?

kianenigma: fair enough 👍

PR #371, opened 2021-03-17
Aura consensus for parachains

bkchr:
This pr introduces Aura for parachains. It is working, but at least one thing is missing to make it secure. We don't verify yet that the slot is "valid", based on the time of the relay block. This will come in some later pr, when we are able to read the relay chain storage more easily.

The pr is also not yet mergable:
- It still depends on some non-merged Substrate branch that I need to clean up.
- Tests are missing and others are maybe not compiling/failing.
- Some docs are missing.

Fixes:  https://github.com/paritytech/cumulus/issues/115

PR #429, opened 2021-05-06
Nimbus Consensus Framework

JoshOrndorff:
This PR introduces Nimbus -- a framework for writing slot-based parachain consensus algorithms. The nimbus framework strives to make non-trivial consensus algorithms a reality for cumulus parachains. This is an alternative approach to #371 that strives to be more reusable. In addition to the framework itself, this PR introduces three example consensus algorithms expressed in the Nimbus framework, and installs one of them in the rococo collator to provide a ready-to-use example of a chain built with Nimbus.

One of the included example consensus engines is the popular Aura round robin engine. I've written a [comparison of Aura implementations](https://github.com/PureStake/cumulus/blob/nimbus/pallets/aura-style-filter/src/lib.rs#L17-L38) to help explain the difference between the two approaches.

I've written a lot of docs about the Nimbus framework's architecture in the [Nimbus Readme](https://github.com/PureStake/cumulus/tree/nimbus/client/consensus/nimbus#readme) (which is part of this PR). I encourage reviewers to read that. Rather than re-iterate that in this PR description, I'll focus on the high-level value add of Nimbus by comparing it to the approach that wraps sc-consensus-aura.

There are two key differences between this approach and #371:
1. **Reusability** - The Nimbus framework is designed so that as much of it as possible can be reused among multiple slot-based consensus algorithms. The entire client-side worker is reused among all three included examples. This is in contrast to the existing Aura and Babe implementations as they exist in the Substrate repo. A development team who wanted to write a custom consensus algorithm in Nimbus will have much less code to write and understand. (Of course this is only true for consensus engines that fit within the slot-based category.) This is evidenced by the <100 lines implementation of Aura.
2. **In-Runtime author checking** - Nimbus checks whether the author of a block is eligible to author in the runtime rather than on the client-side. This makes it easier for relay chain validators, who only run wasm, to make the same eligibility checks. It also makes reusing the client side worker more realistic.


bkchr:
> Nimbus makes the design choice to include the author checking logic in the runtime. This makes it relatively straight forward (but still not entirely straight forward; at least until https://github.com/paritytech/polkadot/issues/2888 is resolved) for relay chain validators to check the authorship information. This is in contrast to the existing implementations of Aura and Babe where the authorship checks are offchain.

I don't understand why you need the linked issue to check the authorship information?

JoshOrndorff:
@bkchr I agree, we don't need paritytech/polkadot#2888 to check the author on the relay chain side. I followed the same approach you showed me [here](https://github.com/paritytech/cumulus/pull/352). I just meant that the interface might get more straight-forward after that PR lands. That work is not blocking this one.

Thanks for looking, any more review is welcome.

pallets/account-set/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	//TODO can I derive default?
+	#[cfg(feature = "std")]
+	impl<T: Config> Default for GenesisConfig<T> {
+		fn default() -> Self {
+			Self {
+				//TODO It would nice to somehow put alice and bob here, but that requires knowledge
+				// of the concrete AccountId type. Maybe I could take it in the config trait and
+				// let the runtime author supply it, but for now, I'm gonna do it in chain_spec.rs
```
shawntabrizi, line 97: You should pass this from the genesis config or a runtime configuration trait

pallets/account-set/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+			StoredAccounts::<T>::get().contains(author)
+		}
+	}
+
+	#[pallet::genesis_config]
+	pub struct GenesisConfig<T: Config> {
+		pub stored_accounts: Vec<T::AuthorId>,
+	}
+
+	//TODO can I derive default?
```
shawntabrizi, line 90: yes, should be

pallets/author-height-filter/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+				// This has the weakness that the authors are based only on para-block height.
+				// For a more robust solution that adds entropy from the parachain inherent, see
+				// ../author-slot-filter
+				let subject: [u8; 7] = [b'f', b'i', b'l', b't', b'e', b'r', i as u8];
+				let (randomness, _) = T::RandomnessSource::random(&subject);
+
+				// TODO why could we not use the same method as in moonbeam?
+				// Why does to_low_u64_be not exist on H256?
+				// https://docs.rs/primitive-types/0.9.0/primitive_types/struct.H256.html#method.to_low_u64_be
+				// Cast to u32 first so we get consistent results on 32- and 64-bit platforms.
```
shawntabrizi, line 92: what 64-bit platforms?

pallets/author-height-filter/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+			Self {
+				eligible_ratio: 50,
+			}
+		}
+	}
+
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig {
+		fn build(&self) {
+			//TODO ensure that the u8 is less than or equal 100 so it can be apercent
```
shawntabrizi, line 160: Should use Percent type in genesis then?

pallets/author-slot-filter/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+			Self {
+				eligible_ratio: 50,
+			}
+		}
+	}
+
+	#[pallet::genesis_build]
+	impl<T: Config> GenesisBuild<T> for GenesisConfig {
+		fn build(&self) {
+			//TODO ensure that the u8 is less than or equal 100 so it can be apercent
```
shawntabrizi, line 175: same comment

bkchr:
>     2\. **In-Runtime author checking** - Nimbus checks whether the author of a block is eligible to author in the runtime rather than on the client-side. This makes it easier for relay chain validators, who only run wasm, to make the same eligibility checks.

You only check that someone passes the correct authorid. This is nothing that makes your implementation more safe or whatever. You only add more computation 🤷

pallets/author-height-filter/src/lib.rs
JoshOrndorff: At first we were just casting directly to `usize` which is 32-bit in wasm which most authoring nodes were using. The problem was full nodes that ran with native execution couldn't import the blocks because they were getting a different index.

pallets/account-set/src/lib.rs
JoshOrndorff: Yep. Agree. I realized later I could probably use the membership pallet from Substrate here. This pallet is just a minimal thing to mock a proper staking solution.

JoshOrndorff:
> This is nothing that makes your implementation more safe or whatever. You only add more computation 🤷

I agree that checking for a valid author in the runtime is not safer or more correct than doing it offchain. I've chosen to do it this way because in a parachain context we need the relay chain validators to do this check as well as the collators, and the validators won't run parachain-specific client-side code. So putting it in the verifier is not enough.

This is the process I'm envisioning.
1. A collator node imports a relay chain block which calls into the `ParachianConsensus` trait asking for a candidate.
2. The collator calls the AuthorFilterAPI runtime api to see whether it will be eligible to author at this slot (let's assume it is).
3. The collator includes its author identity in the author inherent in its block.
4. When the author inherent is executed, the author identity is copied to a consensus digest.
(It may be that steps 3 and 4 should be replaced with a pre-runtime digest. I'm open to learning that, I wrote it this way because I was more familiar with inherents)
5. After the block is constructed, the collator signs it and attaches the signature in a seal digest.
6. When the relay chain validator receives the block, they pop the seal digest and make sure that the signature is valid and made by the same author reported in the first digest which is the same as the one in the inherent. If this signature does not pass, reject the block immediately. If it does pass, we now know that the author in the inherent is the real author who actually authored this block.
7. Begin executing the block using the normal FRAME executive. When the author inherent is executed we find out whether the author is eligible or not.

bkchr:
> I agree that checking for a valid author in the runtime is not safer or more correct than doing it offchain. I've chosen to do it this way because in a parachain context we need the relay chain validators to do this check as well as the collators, and the validators won't run parachain-specific client-side code. So putting it in the verifier is not enough.

I know, but this is already solved by the custom block executor?

JoshOrndorff:
I'll address more of the feedback.

### Weight and Extra Execution

By putting the author checking in a pallet, it is true that we add more execution to the block than if we made this check offchain. Let's consider this from the perspectives of both parachain nodes and validators.

For parachain nodes, this computation will be made either way, checking in the runtime does not change the total amount of computation they do. Since the extrinsic that does the checking is an inherent, it does not cost any fees, so that number is also unchanged. So there is not much difference from the parachain nodes' perspective.

The validators will also have to check the author either way. The difference is whether they do it in the custom executor or the pallet. By putting the execution in the pallet we are able benchmark it and quantify its execution time with weight. This is good. We want to account for all execution time with the weight system. If the author-checking logic is not accounted for in the weight system (as is the case when it goes straight in the executor like in #371), we run the risk of a totally-full block exceeding the execution limit (currently 1/2 sec) because of the unaccounted author check pushes it across the threshold.

In practice this check will be fast and will most likely not be the "last straw" that pushes execution time over the limit, but from a theoretical perspective, I would say accounting the weight is better.

### Reusability

Moving the author checking to a pallet rather than the custom executor, allows a single executor to be reused for any consensus that can be expressed in the nimbus framework (slot-based, signature-sealed algorithms). Before this PR cumulus has two consensus algorithms (RelayChainConsensus and Aura) which each have their own custom executor (admittedly, the one for relay chain consensus is trivial). If you add Babe, you will have another unique custom executor. All of these algorithms can be expressed in nimbus and they would all be able to reuse the same executor.

### HotSwapping Consensus

Parachain teams may want to change, tune, or otherwise adjust their consensus algorithm from time to time. With the nimbus framework, **writing a consensus engine is as easy as writing a pallet**. That means that swapping consensus is as easy as upgrading a pallet. If the author checking were done off-chain this would not be the case. Swappinging consensus would also require a client upgrade and hard fork. I remember @shawntabrizi was disappointed how complex it was to swap consensus in https://github.com/paritytech/substrate/pull/3790, and I believe this approach begins to alleviate some of that.

Obviously this is a baby step toward hot swapping that only allows swapping among consensus engines that fit within nimbus. This misses the main use case of a sovereign chain becoming a parachain or vice versa. Nonetheless, it is a start, and a useful feature for parachain teams who are not yet confident what consensus they will want in the medium to long term.

xlc:
Rick and Morty S05E01 Mort Dinner Rick Andre

bkchr:
So what is the difference between nimbus and making the authority set of Aura configurable?

By making the authority set configurable, I mean the following:

```
Aura frame crate:
trait Config {
   type AuthoritySet: AuthoritySet;
}

trait AuthoritySet {
   fn allowed_authors_for_block() -> Vec<ValidatorId>;
}
```

This is very sketchy, but it would allow you to use whatever authority selection algorithm you want.

JoshOrndorff:
This is about more than changing the authority set. You can change the authority set, but the authorities will still be rotated one at a time in order. So while extending Aura in this way does add generality, it still misses many consensus algorithms that nimbus supports. Here are some examples that would not work in this proposed Aura modification.

* Just like Aura except two authors are selected every slot.
* Authors are selected pseudorandomly such that authors who have not been selected in a long time having higher probability of being selected this time.

Focusing on the consensus layer alone, the core difference between nimbus and Aura is that Aura puts the round-robin logic directly in the consensus layer. So for a block to be aura valid it has to be signed by the author whose turn it is in this slot.

OTOH, Nimbus puts fewer checks in the consensus layer. Nimbus considers any block valid as long as it is signed by the same author id that is injected into the runtime. From there it is up to the runtime to reject blocks that where the author id that was injected is not eligible.

bkchr:
> This is about more than changing the authority set. If you change the authority set in Aura, the authorities will still be rotated one at a time in order. So while extending Aura in this way does add generality, it still misses many consensus algorithms that nimbus supports. Here are some examples that would not work in this proposed Aura modification.
> 
>     * Just like Aura except two authors are selected every slot.
> 
>     * Authors are selected pseudorandomly such that authors who have not been selected in a long time having higher probability of being selected this time.

All this would be supported by my idea. The authorityset trait clearly needs more functions, however it would be responsible for all of this. Aka checking that the author is correct etc. allowing multiple authors per block, having a different set per block. whatever.

JoshOrndorff:
So your idea is about modifying more than the runtime side of Aura, right?

Currently the [rotate-in-order](https://github.com/paritytech/substrate/blob/master/client/consensus/aura/src/lib.rs#L94-L109) logic is coded into the client-side worker. And the [`claim_slot` implementation](https://github.com/paritytech/substrate/blob/master/client/consensus/aura/src/lib.rs#L355) assumes that there will only be one author per slot.

Definitely all of that could be generalized. And once it is generalized, I think the remaining differences are just superficial. Here's one that comes to mind.
```rust
// Your exhaustive set idea:
fn allowed_authors_for_block() -> Vec<ValidatorId>;

// My propositional set idea:
fn can_author(author: AuthorId, slot: u32) -> bool;
```

It seems like our ideas have basically converged :) I think if you want to modify Aura that much, the name "authority round" doesn't make that much sense anymore, so we should change the name. Might I suggest Nimbus ;)

bkchr:
Yeah I also know that this would require touching the client side :P 

I will bring up the name, however I would like to not have that much special code here in cumulus that is mostly the same for Aura.

JoshOrndorff:
> I would like to not have that much special code here in cumulus...

Yeah, I agree with that completely. As this work progressed, I realized most of it is just the plain consensus engine, and only a small part of it is cumulus specific. It would make more sense to have the main consensus engine part directly in Substrate.

> ...that is mostly the same for Aura.
Yeah, I agree with that too. One of my goals here was not duplicating so much code to write a new consensus engine.

> I will bring up the name

Haha cool. No pressure, I was just teasing because it's one of the only remaining differences.

PR #435, opened 2021-05-12
Rename `rococo-collator` to `polkadot-collator`

bkchr:
Also contains some other small changes.


JoshOrndorff:
Is this prep work for #430 ?

Is the intention that the same service (now called polkadot-collator) will run multiple runtimes? Like how it is in the Polkadot repo?

bkchr:
> Is this prep work for #430 ?

No 


> Is the intention that the same service (now called polkadot-collator) will run multiple runtimes? Like how it is in the Polkadot repo?

This was already the case before ;)

PR #437, opened 2021-05-13
Validate timestamps against relay chain slot

### LINKED ISSUE #422: Put relay chain slot number into ParachainInherent
The `ParachainInherent` already contains some data from the relay chain state that is mainly required for messaging. This will be moved to the runtime when we got https://github.com/paritytech/cumulus/issues/320. However, as long as we don't have the functionality to read the relay chain state from the parachain runtime, we should add the relay chain slot number to the relay chain storage proof. 

The relay chain slot number will be required to validate the timestamp inherent when validating the PoV on the relay chain. This should be done by using some Listener which will be called when the `ValidationData` is set and when the timestamp is set. If both are set we should verify that the timestamp is reasonable given the slot number.

andresilva:
Fix #422. 

Depends on https://github.com/paritytech/polkadot/pull/3022.

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -32,17 +32,17 @@ use cumulus_primitives_core::{
 	well_known_keys::{self, NEW_VALIDATION_CODE},
 	AbridgedHostConfiguration, ChannelStatus, DmpMessageHandler, GetChannelInfo,
 	InboundDownwardMessage, InboundHrmpMessage, MessageSendError, OnValidationData,
-	OutboundHrmpMessage, ParaId, PersistedValidationData, UpwardMessage, UpwardMessageSender,
+	OutboundHrmpMessage, ParaId, PersistedValidationData, Slot, UpwardMessage, UpwardMessageSender,
 	XcmpMessageHandler, XcmpMessageSource,
 };
 use cumulus_primitives_parachain_inherent::ParachainInherentData;
 use frame_support::{
-	ensure,
 	dispatch::{DispatchError, DispatchResult},
+	ensure,
+	inherent::{InherentData, InherentIdentifier, ProvideInherent},
 	storage,
 	traits::Get,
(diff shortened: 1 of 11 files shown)


pallets/parachain-system/src/lib.rs
```diff
@@ -466,6 +468,16 @@ pub mod pallet {
...
 
+	/// The relay chain current slot number that was obtained from the relay parent.
+	///
+	/// This field is meant to be updated each block with the validation data inherent. Therefore,
+	/// before processing of the inherent, e.g. in `on_initialize` this data may be stale.
+	///
+	/// This data is also absent from the genesis.
+	#[pallet::storage]
+	#[pallet::getter(fn relay_slot)]
+	pub(super) type RelaySlot<T: Config> = StorageValue<_, Slot>;
```
bkchr, line 479: I don't see any reason to add this here as a storage item.

pallets/parachain-system/src/validate_block/timestamp.rs
```diff
@@ -0,0 +1,117 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus. If not, see <http://www.gnu.org/licenses/>.
+
+//! Utility module to validate cumulus-runtime timestamps against the
+//! relay chain slot.
+
+use cumulus_primitives_core::{OnValidationData, Slot};
```
bkchr, line 20: This should be its own pallet. Similar to `aura-ext`. Maybe `timestamp-ext` or something.

The idea with using environmental was also a bad idea by me. We should probably store the intermediate data in the storage.

bkchr (changes requested):

bkchr:
Superseded by https://github.com/paritytech/cumulus/pull/494

Issue #344, opened 2021-02-25
Extract data for collation via runtime api


bkchr:
Currently we use some well known keys to read all the relevant data that we require to create the `Collation` struct in the collator. Instead of using well known keys, we should use a runtime api with one function that returns all the required data at once. This way we could remove these well known keys completely.

For the validate block function, we would use the same way as done for the validation data in https://github.com/paritytech/cumulus/pull/342 to access it.

bkchr:
1. Create a runtime api declaration ([See](https://docs.rs/sp-api/3.0.0/sp_api/macro.decl_runtime_apis.html) for more infos).
```
struct CollationInfo {
   // all the relevant fields
}

decl_runtime_apis! {
    /// I'm happy about any other name :P 
    trait FetchCollationInfo {
         fn fetch() -> CollationInfo;
    }
}
```

2. Implement the runtime api for the relevant runtimes in Cumulus.

3. Provide a function in `cumulus-pallet-parachain-system` that can be called by the runtime api and that returns `CollationInfo`.

4. Change [this](https://github.com/paritytech/cumulus/blob/master/client/collator/src/lib.rs#L152-L254) code to use the new runtime api. Here you also can see which fields `CollationInfo` requires. 

5. Make [this](https://github.com/paritytech/cumulus/blob/master/pallets/parachain-system/src/validate_block/implementation.rs#L143L175) code switch to this new interface of the parachain system pallet as well. Remove the well known keys completely.

6. ....

7. Profit!

PR #443, opened 2021-05-17
Introduce `CollectCollationInfo` runtime api

bkchr:
Instead of using well known keys to communicate information about a
collation between the runtime and the collator, we now use a runtime api
for this.

Fixes: https://github.com/paritytech/cumulus/issues/344


pallets/parachain-system/src/lib.rs
```diff
@@ -481,6 +477,39 @@ pub mod pallet {
...
+	pub(super) type NewValidationCode<T: Config> = StorageValue<_, Vec<u8>, OptionQuery>;
+
+	/// HRMP watermark that was set in a block.
+	///
+	/// This will be cleared in `on_initialize` of each new block.
+	#[pallet::storage]
+	pub(super) type HrmpWatermark<T: Config> =
+		StorageValue<_, relay_chain::v1::BlockNumber, ValueQuery>;
+
+	/// HRMP messages that were send in a block.
```
pepyakin, line 499:
```suggestion
	/// HRMP messages that were sent in a block.
```

pallets/parachain-system/src/lib.rs
```diff
@@ -481,6 +477,39 @@ pub mod pallet {
...
 
+	/// Number of downward messages processed in a block.
+	///
+	/// This will be cleared in `on_initialize` of each new block.
+	#[pallet::storage]
+	pub(super) type ProcessedDownwardMessages<T: Config> = StorageValue<_, u32, ValueQuery>;
+
+	/// New validation code that was set in a block.
+	///
+	/// This will be cleared in `on_initialize` of each new block.
```
pepyakin, line 488: It's not so simple here due to the `DidSetValidationCode` check.

pepyakin (approved):
Looks good. I really love this change. 

A couple of nits and typos.

Issue #388, opened 2021-04-04
Recovery of PoVs


akru:
> This report is not only an idea but already happens on the test network.

**Description**

Let parachain A has four collators: Alice, Bob, Charlie, and Dave. Let Dave is a malicious collator. Currently having at least one collator is possible to stop block producing on a parachain. Let's see following steps:

1. Alice, Bob, Charlie, and Dave proposing blocks to relay chain validators.
2. All collators are synced up and can propose a block.
3. If the block proposal (or block sentence) from Dave accepted, then Dave stops to share blocks with each other.
4. As a result Dave only one can propose a block because of having a full chain.

**Possible solutions**

Probably collator should be able to recover the block from the relay when it can't get the best block from sync process.

bkchr:
Yeah, we are aware of it.

And somewhere I have an todo to implement recovery from the relay chain.

akru:
@bkchr thank you, looking forward to updates.

burdges:
I've explained this in chat and https://github.com/paritytech/polkadot/issues/2203, but might as well spell it out here..

Safer Option:  We've at least 2f+1 out of 3f+1 validators who signed that they have availability pieces of Dave's block, so a collator could contact any f+1 of those, obtain their piece, and reconstruct Dave's block from the erasure coding.  This works fine, but requires opening connections to 256 < f+1 < 512 validators.  It's fine if only the next upcoming collator does so, but not good if many do so.

Faster Option:  There are three node classifications who know the full parachain block, the collator who created it (Dave), the two-ish validators who acted as backing checkers for it (maybe Dave's buddies), and the approval checkers who checked the block.  Approval checkers can only be known by listening to the relay chain gossip, but Alice, Bob, Charlie already do so anyways.  Approval checkers do not currently retain Dave's block, but they could do so temporarily, which enables downloading the whole block by talking to only one validator. 

Approval checkers cannot be known by Dave in advance, meaning he cannot plan upon this working.  Yet, if Dave controls proportions q < 1/3 of the collators and p<1/3 of the validators then he waits an expected 6 p^{-16}/q seconds between attempts, roughly 1 year with q=p=1/3.   Acceptable risk since stalls are not soundness violations.  

There is nothing wrong with either option, so we'll implement which ever looks simplest, almost surely the safer slower option.  If everyone behaves then no problem.  :)  We could later add an opt-in form of the faster option, if we're worried both about this attack and about approval checkers holding blocks too long.  

---

[...]

PR #445, opened 2021-05-19
Pov recovery for parachains

bkchr:
This implements pov recovery for parachain collators.

The idea is that when a collator observes an unknown candidate, for that it doesn't yet has imported a block, we will start a timer. If the timer is ready and we don't have imported the block, we will start to recover the PoV from the relay chain. 

In the current model with 12 seconds slot time, this should be working relative smoothly. In the future we probably need to fine tune this.

Missing:
- Some docs for the crate
- Fetch the relay-chain slot duration from the relay chain client

Fixes: https://github.com/paritytech/cumulus/issues/388


client/pov-recovery/src/active_candidate_recovery.rs
```diff
@@ -0,0 +1,112 @@
...
+	///
+	/// If the returned [`AvailableData`] is `None`, it means that the recovery failed.
+	pub async fn wait_for_recovery(&mut self) -> (Block::Hash, Option<AvailableData>) {
+		loop {
+			if self.recoveries.is_empty() {
+				futures::pending!()
+			} else if let Some(res) = self.recoveries.next().await {
+				self.candidates.remove(&res.0);
+				return res;
+			}
```
drahnr, line 109:
```suggestion
			if let Some(res) = self.recoveries.next().await {
				self.candidates.remove(&res.0);
				return res;
			} else {
				futures::pending!()
			}
```

client/pov-recovery/src/lib.rs
```diff
@@ -0,0 +1,434 @@
...
+	/// Run the pov-recovery.
+	pub async fn run(mut self) {
+		let mut imported_blocks = self.parachain_client.import_notification_stream().fuse();
+		let mut finalized_blocks = self.parachain_client.finality_notification_stream().fuse();
+		let pending_candidates =
+			pending_candidates(self.relay_chain_client.clone(), self.para_id).fuse();
+		futures::pin_mut!(pending_candidates);
+
+		loop {
+			select! {
```
drahnr, line 353: Doesn't this drop one of the futures if they'd complete at the same time?

PR #448, opened 2021-05-21
Some changes for Statemint

bkchr:
Some changes that are required for Statemint to upgrade from Shell to Statemint.

This is mainly about exposing types and stuff, to make it usable from the outside ;)

### DIFF
--- a/client/consensus/common/src/lib.rs
+++ b/client/consensus/common/src/lib.rs
@@ -561,6 +561,50 @@ impl<B: BlockT> ParachainConsensus<B> for Box<dyn ParachainConsensus<B> + Send +
 	}
 }
 
+/// Parachain specific block import.
+///
+/// This is used to set `block_import_params.fork_choice` to `false` as long as the block origin is
+/// not `NetworkInitialSync`. The best block for parachains is determined by the relay chain. Meaning
+/// we will update the best block, as it is included by the relay-chain.
+pub struct ParachainBlockImport<I>(I);
+
+impl<I> ParachainBlockImport<I> {
+	/// Create a new instance.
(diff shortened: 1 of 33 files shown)


client/consensus/common/src/lib.rs
```diff
@@ -561,6 +561,50 @@ impl<B: BlockT> ParachainConsensus<B> for Box<dyn ParachainConsensus<B> + Send +
...
+	async fn import_block(
+		&mut self,
+		mut block_import_params: sp_consensus::BlockImportParams<Block, Self::Transaction>,
+		cache: std::collections::HashMap<sp_consensus::import_queue::CacheKeyId, Vec<u8>>,
+	) -> Result<sp_consensus::ImportResult, Self::Error> {
+		// Best block is determined by the relay chain, or if we are doing the intial sync
+		// we import all blocks as new best.
+		block_import_params.fork_choice = Some(sp_consensus::ForkChoiceStrategy::Custom(
+			block_import_params.origin == sp_consensus::BlockOrigin::NetworkInitialSync,
+		));
```
JoshOrndorff, line 603: I wondered about this import logic because we have occasionally seen parachain nodes sync all the way to the tip, and then fail to keep up with the chain.

Imagine you are syncing from a particular collator. Right as you reach the tip of the chain, that collator sends you its recently-authored block. That block has not yet been included in the relay chain. You import it as best anyway because you are still syncing. Eventually a different candidate gets included in the relay chain.

Will the current logic handle this gracefully?

bkchr: Yes, you are not in initial sync when reaching the tip. And even if it is enabled, we would just reorg.

PR #452, opened 2021-05-29
Add Statemint

gavofyork:
This used statemint repo commit `9664809ad5de6ce43b6b78a6b3d12fbfece7f147`.


pallets/collator-selection/src/lib.rs
```diff
@@ -0,0 +1,434 @@
...
+
+//! Collator Selection pallet.
+//!
+//! A pallet to manage collators in a parachain.
+//!
+//! ## Overview
+//!
+//! The Collator Selection pallet manages the collators of a parachain. **Collation is _not_ a
+//! secure activity** and this pallet does not implement any game-theoretic mechanisms to meet BFT
+//! safety assumptions of the chosen set.
```
JoshOrndorff, line 24: > Collation is not a secure activity.

I agree with this for some definitions of "secure" but not for all. I agree that no matter how badly collators behave they can't cause the relay chain to finalize invalid parachain state transitions.

But there are still important ways in which a corrupt collator set can misbehave and cause nontrivial problems for a parachain:
1. A collator can skip slots.
2. A collator can censor transactions.

If this is couple with an author-selection mechanism that selects only one author per slot in the consensus layer (like Aura) a single malicious collator can cause the parachain to produce no block in that slot. The more collators do this the more the chain throughput decreases.

Regarding censorship, if the collator set ever becomes entirely colluding, they can retain unilateral control forever by censoring any transactions that would bond other accounts as collators.

JoshOrndorff: Actualy, I think we could construct a minimal liveness proof that a block will be authored eventually from the kick_mechanism and an assumption that the chain is launched with at least one honest collator. But the practical concern still exists. One whale could stake a lot of accounts and skip all the corresponding slots.

Adding a slash to the kick mechanism would ensure that this whale attacker would eventually bleed enough stake to not be a problem, but there would still be an unstable transient period after launch before settling down to a steady state.

gavofyork:
> Regarding censorship, if the collator set ever becomes entirely colluding, they can retain unilateral control forever by censoring any transactions that would bond other accounts as collators.

We generally assume at most 33% byzantine.

Worst-case, relay-chain governance can jump in, forcibly alter the state to one which kicks (and slashes, perhaps) all the bad collators.

pallets/collator-selection/src/lib.rs
gavofyork: There could be a slash, but it work similarly to the relay-chain staking where the slash is non-zero only for when a threshold go offline at once (and can be superlinear afterwwards).

polkadot-parachains/src/command.rs
```diff
@@ -161,6 +169,14 @@ fn use_shell_runtime(chain_spec: &dyn ChainSpec) -> bool {
...
 
+fn use_statemine_runtime(chain_spec: &dyn ChainSpec) -> bool {
+	chain_spec.id().starts_with("statemine")
+}
+
+fn use_westmint_runtime(chain_spec: &dyn ChainSpec) -> bool {
+	chain_spec.id().starts_with("westmint")
+}
+
 use crate::service::{new_partial, RococoParachainRuntimeExecutor, ShellRuntimeExecutor};
```
riusricardo, line 180: Do the other runtimes executors need to be added?
`StatemintRuntimeExecutor`, `StatemineRuntimeExecutor`, `WestmintRuntimeExecutor`

riusricardo: https://github.com/paritytech/statemint/blob/9e8d640dfe0311652d848bfd0aea3ae62bc54b0f/node/src/service.rs#L41
https://github.com/paritytech/statemint/blob/9e8d640dfe0311652d848bfd0aea3ae62bc54b0f/node/src/service.rs#L49
https://github.com/paritytech/statemint/blob/9e8d640dfe0311652d848bfd0aea3ae62bc54b0f/node/src/service.rs#L57

apopiak:
Looks fine to me, can you mention the Statemint repo commit that you copied so we can determine which other changes we need to move as well?

xlc:
So statemint will live here instead of its own repo?

polkadot-parachains/src/command.rs
gavofyork: Done.

joepetrowski:
> So statemint will live here instead of its own repo?

yes

apopiak:
@gavofyork @bkchr Correct me if I'm wrong, but:
I don't see the statemint node changes from https://github.com/paritytech/statemint/pull/80/files#diff-594440db8adfc5d1b37711309312bfa06b49e488309186ad062a896778cf9fbfR84-R96 anywhere so I don't think the transition to AuRa would work smoothly using the `polkadot-collator`

polkadot-parachains/src/chain_spec.rs
```diff
@@ -193,3 +192,534 @@ fn shell_testnet_genesis(parachain_id: ParaId) -> shell_runtime::GenesisConfig {
...
+						 hex!("1256436307dfde969324e95b8c62cb9101f520a39435e6af0f7ac07b34e1931f").into(),
+						 hex!("1256436307dfde969324e95b8c62cb9101f520a39435e6af0f7ac07b34e1931f").unchecked_into()
+					 ),
+					 (
+						 hex!("98102b7bca3f070f9aa19f58feed2c0a4e107d203396028ec17a47e1ed80e322").into(),
+						 hex!("98102b7bca3f070f9aa19f58feed2c0a4e107d203396028ec17a47e1ed80e322").unchecked_into()
+					 ),
+				],
+				vec![],
+				hex!("6648d7f3382690650c681aba1b993cd11e54deb4df21a3a18c3e2177de9f7342").into(),
```
apopiak, line 671:
```suggestion
				// re-use the Westend sudo key
				hex!("6648d7f3382690650c681aba1b993cd11e54deb4df21a3a18c3e2177de9f7342").into(),
```

bkchr (approved):

riusricardo (approved):

polkadot-parachains/src/command.rs
```diff
@@ -112,7 +151,11 @@ impl SubstrateCli for Cli {
 	}
 
 	fn native_runtime_version(chain_spec: &Box<dyn ChainSpec>) -> &'static RuntimeVersion {
-		if use_shell_runtime(&**chain_spec) {
+		if chain_spec.is_statemint() {
+			&statemint_runtime::VERSION
+		} else if chain_spec.is_statemine() {
+			&statemine_runtime::VERSION
+		} else if chain_spec.is_shell() {
```
apopiak, line 158: westmint is missing here

polkadot-parachains/src/service.rs
```diff
@@ -510,15 +519,16 @@ pub fn shell_build_import_queue(
 	.map_err(Into::into)
 }
 
-/// Start a rococo-shell parachain node.
+/// Start a polkadator-shell parachain node.
```
apopiak, line 522:
```suggestion
/// Start a polkadat-shell parachain node.
```

apopiak (approved):
LGTM

polkadot-parachains/src/service.rs
```diff
@@ -486,15 +519,16 @@ pub fn shell_build_import_queue(
 	.map_err(Into::into)
 }
 
-/// Start a rococo-shell parachain node.
+/// Start a polkadat-shell parachain node.
```
apopiak, line 522:
```suggestion
/// Start a polkadot-shell parachain node.
```

PR #462, opened 2021-06-02
Add srtool + subwasm workflow

chevdor:
This PR adds a new github workflow to the CI.
It does the following:
- run `srtool`
- save the wasm
- install/run `subwasm`
- fetch the info, including the blake2_256 hash from the wasm
- store relevant artifacts

### DIFF
--- a/.github/workflows/srtool.yml
+++ b/.github/workflows/srtool.yml
@@ -23,6 +23,6 @@ jobs:
       - name: Archive Runtime
         uses: actions/upload-artifact@v2
         with:
-          name: ${{ matrix.chain }}-runtime-${{ github.ref }}
+          name: ${{ matrix.chain }}-runtime-${{ github.sha }}
           path: |
             ${{ steps.srtool_build.outputs.wasm }}
--- a/.github/workflows/srtool.yml
+++ b/.github/workflows/srtool.yml
@@ -23,6 +23,6 @@ jobs:
       - name: Archive Runtime
         uses: actions/upload-artifact@v2
         with:
-          name: ${{ matrix.chain }}-runtime
+          name: ${{ matrix.chain }}-runtime-${{ github.ref }}
           path: |
-            "${{ steps.srtool_build.outputs.wasm }}"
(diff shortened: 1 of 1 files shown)


chevdor:
A sample run can be seen here:
https://github.com/chevdor/cumulus/runs/2734510623?check_suite_focus=true

.github/workflows/srtool.yml
```diff
@@ -0,0 +1,58 @@
+name: Srtool build
+
+on: push
```
apopiak, line 3: I don't think we want to run the workflow on every push?

apopiak: @s3krit Can we run it on a release? E.g. on certain branch names or tags?

.github/workflows/srtool.yml
```diff
@@ -0,0 +1,58 @@
...
+
+jobs:
+  srtool:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        chain: ["statemine"]
```
apopiak, line 10: would also be cool to run it for westmint

apopiak, line 10:
```suggestion
        chain: ["statemine", "westmint"]
```
so change to this?

apopiak (commented):
Very cool!
Might want to not run it on every push and run it for westmint, too

chevdor:
Since I am off today, my prio was to have it work first so you get wasm and
blake2_256.

Sure, on push is too much. The extra runs can be cancelled manually for now
though.

The last fix i made in subwasm should make the CI as it is also pass for
westmint.

chevdor:
The first run including westmint can be seen here: https://github.com/chevdor/cumulus/actions/runs/902147585

.github/workflows/srtool.yml
s3krit: Yep dead simple - if you know the pattern of what cumulus release tags look like, you can match like so: https://github.com/paritytech/polkadot/blob/master/.github/workflows/publish-draft-release.yml#L7

s3krit (approved):

Issue #466, opened 2021-06-03
Parachain runtimes depend on`polkadot-service` and therefore all polkadot runtimes.


JoshOrndorff:
Every parachain runtime needs to include pallet parachain system, which indirectly depends on polkadot-service (when building to native). This means that to build a parachain runtime (not even the entire node) we have to wait to build all the polkadot runtimes.

Exploring this dependency further we see
```
cumulus/polkadot-parachains/rococo-runtime$ cargo tree -i polkadot-service

polkadot-service v0.9.3 (https://github.com/paritytech/polkadot?branch=master#6077ed06)
└── cumulus-primitives-parachain-inherent v0.1.0 (/home/joshy/ProgrammingProjects/cumulus/primitives/parachain-inherent)
    └── cumulus-pallet-parachain-system v0.1.0 (/home/joshy/ProgrammingProjects/cumulus/pallets/parachain-system)
        └── cumulus-rococo-parachain-runtime v0.1.0 (/home/joshy/ProgrammingProjects/cumulus/polkadot-parachains/rococo-runtime)
```

The culprit is `cumulus-primitves-parachain-inherent`. This crate includes the definition `ParachainInherentData`, as well as a bunch of client side stuff that is not needed in the runtime.

Currently,`cumulus-primitves-parachain-inherent` uses (abuses) the `std` feature to gate all of the client-side stuff. That means that when we build the runtime to std we also build all the client side stuff which is actually not necessary. I recommend using a separate `client-side` feature to gate that stuff so that the `polkadot-service` dependency doesn't get pulled into the runtime.

As a side note, this unintentional dependency is causing grief when writing runtime benchmarks for a parachain. The parachain runtime has to include some otherwise-unnecessary dependencies such as pallet-society just so we can gate their runtime benchmark features. Adding pallet-society as a dependency to a runtime that does not use that pallet is a sign that something is wrong.

PR #467, opened 2021-06-03
Feature-gate client-side of cumulus-primitive-parachain-inherent

JoshOrndorff:
This is my attempt at solving #466 

It doesn't seem to work though because when I `cargo check -p cumulus-pallet-parachain-system` it still builds all the polkadot runtimes.


bkchr:
https://github.com/paritytech/cumulus/pull/489 this is the best we currently can do.

PR #470, opened 2021-06-03
Add raw chainspecs for Westmint and Statemine

lovelaced:


polkadot-parachains/res/westmint.json
```diff
@@ -1,20 +1,21 @@
...
+             "/ip4/34.90.183.162/tcp/30334/p2p/12D3KooWDPmzizJMgfb1tz4MvPtfAw88gfbttdSv56iofxA1TgdM",
+             "/ip4/35.204.49.222/tcp/30334/p2p/12D3KooWRG7deBzjHA94DqVovULegLq3JYTap27dP1rXJDjmtrCx",
+             "/ip4/35.204.177.184/tcp/30334/p2p/12D3KooWML4JGiNCCo8mzohDhbVZjJLgwVzFFPZBfzwU22HYdupF"
+     ],
+   "telemetryEndpoints": null,
   "protocolId": null,
-  "properties": {
-    "tokenDecimals": 12,
-    "tokenSymbol": "WND"
-  },
```
apopiak, line 17: was there a reason for removing the properties?

jacogr: Yes, needs to be re-added. (Same with Statemine, e.g. as deployed to the actual RPC nodes)

There are also other RPC operators that use these specs, so they should reflect the properties otherwise weird things will and do happen...

apopiak:
I added the properties for statemine as well. Can you guys confirm that that's fine? @lovelaced @jacogr ?
then we can merge this soon

apopiak (approved):

lovelaced:
> I added the properties for statemine as well. Can you guys confirm that that's fine? @lovelaced @jacogr ?
> then we can merge this soon

Looks good to me, thanks

PR #471, opened 2021-06-04
feat(ci): update runtime build workflow

chevdor:
- switch to a tag (v0.1.0) for the `chevdor/srtool-actions` (previously using a `draft` branch)
- trigger to only tags & ignore folders not involved in the runtime
- add weekly build schedule (every Monday @ 2am)
- add timestamp to the artifacts

### DIFF
--- a/.github/workflows/srtool.yml
+++ b/.github/workflows/srtool.yml
@@ -21,11 +21,8 @@ jobs:
       matrix:
         chain: ["statemine", "westmint"]
     steps:
-      - name: Get Time
-        id: time
-        uses: nanzm/get-time-action@v1.1
-        with:
-          format: "YYYY_MM_DD-HH_mm_ss"
+      - name: Get Timestamp
+        run: echo "TMSP=$(date '+%Y%m%d_%H%M%S')" >> $GITHUB_ENV
       - uses: actions/checkout@v2
       - name: Srtool build
         id: srtool_build
@@ -43,7 +40,7 @@ jobs:
         env:
           TMSP: "${{ steps.time.outputs.time }}"
         with:
-          name: ${{ matrix.chain }}-runtime-$TIME-${{ github.sha }}
+          name: ${{ matrix.chain }}-runtime-${{ env.TMSP }}-${{ github.sha }}
           path: |
             ${{ steps.srtool_build.outputs.wasm }}
             ${{ matrix.chain }}-srtool-digest.json
@@ -72,7 +69,7 @@ jobs:
         env:
           TMSP: "${{ steps.time.outputs.time }}"
         with:
-          name: ${{ matrix.chain }}-runtime-$TIME-${{ github.sha }}
+          name: ${{ matrix.chain }}-runtime-${{ env.TMSP }}-${{ github.sha }}
           path: |
             ${{ matrix.chain }}-info.json
             ${{ matrix.chain }}-metadata.json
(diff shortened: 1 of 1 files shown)


chevdor:
Waiting for https://github.com/chevdor/cumulus/actions/runs/905829184 to pass

.github/workflows/srtool.yml
```diff
@@ -1,6 +1,18 @@
 name: Srtool build
 
-on: push
+on:
+  push:
+    tags:
+      - "*"
```
s3krit, line 6: Is there a specific pattern of tab we could match on rather than any tag?

.github/workflows/srtool.yml
```diff
@@ -9,10 +21,15 @@ jobs:
       matrix:
         chain: ["statemine", "westmint"]
     steps:
+      - name: Get Time
+        id: time
+        uses: nanzm/get-time-action@v1.1
```
s3krit, line 26: It's weird to me that we'd need to rely on a third party action to just get the time. As a general rule I like to only use third party actions where necessary. Maybe this is necessary though.

s3krit (commented):
Looks good. Just leaving a comment for now till my questions are addressed

bkchr:
>     * add weekly build schedule (every Monday @ 2am)

What is the point of this?

.github/workflows/srtool.yml
chevdor: I checked the existing tag~s~ (https://github.com/paritytech/cumulus/tags) and for now, the list is too short for a pattern. Since the frequency of new tags is still low, I think picking up all tags is better than every single and fair for now. My proposal would be to refine the pattern as soon as we start using tags , that would trigger the runtime build but that we also want to exclude.

.github/workflows/srtool.yml
chevdor: We actually don't and I agree. I guess we may use that more so a homebrewed action maybe better. For now, I think I can remove it entirely.

chevdor:
> * add weekly build schedule (every Monday @ 2am)
>What is the point of this?

Since the frequency of tagging is low (at least currently), I want to ensure we don't end up not building the runtime for a too long period of time and then discover potential issues with the CI when we do need it wot run and work.

Why monday 2am: I think a weekly build is fair and not overdone. It runs when it will disturb no one and we will have potential issues visible mondays first hour. We totally can reconsider the frequency up or down if the activity of the repo changes one way or another.

I think it can also be convenient for people who want to test to have a fairly fresh runtime build.

bkchr:
> > * add weekly build schedule (every Monday @ 2am)
> >   What is the point of this?
> 
> Since the frequency of tagging is low (at least currently), I want to ensure we don't end up not building the runtime for a too long period of time and then discover potential issues with the CI when we do need it wot run and work.
> 

And how will you find out that it doesn't build anymore? Remember to check every monday if the build succeeded?

chevdor:
> And how will you find out that it doesn't build anymore? Remember to check every monday if the build succeeded?

While that would be an option, it would be far from convenient...

I prefer github to simply let me know.
image

.github/workflows/srtool.yml
chevdor: I removed it

bkchr:
Okay, I didn't know that github actions support to send email automatically. That is fine.

s3krit (approved):

PR #477, opened 2021-06-07
Add session weight

JesseAbram:
This as talked about creates a session benchmark for statemint that does not rely on staking pallet and thus gives truer weights (weights were run on my laptop but once approved I can run them on a proper benchmachine and add them in).


scripts/benchmarks.sh
```diff
@@ -0,0 +1,44 @@
...
+statemintOutput=./polkadot-parachains/statemint-runtime/src/weights
+statemineChain=statemine-dev
+statemintChain=statemint-dev
+pallets=(
+    pallet_assets
+	pallet_balances
+	pallet_collator_selection
+	pallet_multisig
+	pallet_proxy
+    pallet_session
```
shawntabrizi, line 15:
```suggestion
	pallet_proxy
	pallet_session
```

scripts/benchmarks.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+steps=50
+repeat=20
+statemineOutput=./polkadot-parachains/statemine-runtime/src/weights
+statemintOutput=./polkadot-parachains/statemint-runtime/src/weights
+statemineChain=statemine-dev
```
shawntabrizi, line 7: should really get the bot working on this repo :)

shawntabrizi (approved):

PR #486, opened 2021-06-10
Remove All Assets and Uniques from `NonTransfer` Proxy

shawntabrizi:
Given the nature of the Assets and Uniques pallet, it seems very sensible to disable all access to this pallets from the `NonTransfer` proxy.

Combinations of calls like `mint` and `burn` can be used to emulate transfer behaviors, and who knows what new functions will be added.

Given that we have explicit Proxies for `AssetOwner` and `AssetManager`, which specifically handle management of these pallets, I see little reason for `NonTransfer` proxy to have this risk.

Fixes: https://github.com/paritytech/srlabs_findings/issues/91


xlc:
I am thinking if it make sense to introduce an event based proxy filtering mechanism. i.e. fail the transaction if a specific event is emitted.

shawntabrizi:
@xlc would only work for transactional extrinsics?

xlc:
Right. So that't depends on making everything `transactional` first.
Another way could be annotate each calls with tags and filter based on whitelist / blacklist of the tags. I guess this won't happen though... But the current proxy filter is really error prone.

bkchr (approved):
Unique is not in statemint and westmint?

shawntabrizi:
@bkchr nope. should it be?

shawntabrizi:
bot merge

PR #488, opened 2021-06-10
Asset Transaction Payment

apopiak:
Add a pallet to allow for transactionpayment in assets.


apopiak:
@jacogr We'll need a different SignedExtra for the transactions here.

pallets/asset-tx-payment/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+		inherent::Vec,
+		traits::{
+			Currency, ReservableCurrency, EnsureOrigin, ExistenceRequirement::KeepAlive,
+		},
+		PalletId,
+	};
+	use frame_system::pallet_prelude::*;
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config + pallet_transaction_payment::Config + pallet_balances::Config {
```
shawntabrizi, line 84: just starting to read this, but do we need to hard-depend on pallets here?

pallets/asset-tx-payment/src/tests.rs
```diff
@@ -0,0 +1,819 @@
...
+#[test]
+fn transaction_payment_in_asset_possible() {
+	let base_weight = 5;
+	ExtBuilder::default()
+		.balance_factor(10)
+		.base_weight(base_weight)
+		.build()
+		.execute_with(||
+	{
+		// work around visibility limitations in the assets pallet
```
shawntabrizi, line 323: can be updated now?

pallets/asset-tx-payment/src/tests.rs
```diff
@@ -0,0 +1,819 @@
...
+		assert_ok!(Assets::mint_into(asset_id, &beneficiary, balance));
+		assert_eq!(Assets::balance(asset_id, caller), balance);
+		let weight = 5;
+		// we have a ratio of 1::2 for native vs asset minimum balance
+		let fee = (base_weight + weight + len as u64) * 2;
+		let pre = ChargeAssetTxPayment::<Runtime>::from(0, Some(asset_id))
+			.pre_dispatch(&caller, CALL, &info_from_weight(weight), len)
+			.unwrap();
+		// assert that native balance is not used
+		assert_eq!(Balances::free_balance(caller), balance);
```
shawntabrizi, line 346: would be cool not only to assert this, but also assert that the user has zero balance to start with (maybe in another test)

pallets/asset-tx-payment/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+		len: usize,
+	) -> Result<
+		(
+			BalanceOf<T>,
+			InitialPayment<T>,
+		),
+		TransactionValidityError,
+	> {
+		let tip = self.0;
+		let fee = pallet_transaction_payment::Module::<T>::compute_fee(len as u32, info, tip);
```
apopiak, line 131: @shawntabrizi We reuse `compute_fee` of `pallet_transaction_payment` here, thus the the dependence on its `Config` trait.

pallets/asset-tx-payment/src/lib.rs
apopiak: for `pallet_transaction_payment` see below comment.
`pallet_balances` might be a remnant from an earlier implementation, will try to remove and see what happens :-D

pallets/asset-tx-payment/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+	/// maximum corresponding limit.
+	///
+	/// For example, if a transaction consumed 1/4th of the block length and half of the weight, its
+	/// final priority is `fee * min(2, 4) = fee * 2`. If it consumed `1/4th` of the block length
+	/// and the entire block weight `(1/1)`, its priority is `fee * min(1, 4) = fee * 1`. This means
+	///  that the transaction which consumes more resources (either length or weight) with the same
+	/// `fee` ends up having lower priority.
+	fn get_priority(len: usize, info: &DispatchInfoOf<T::Call>, final_fee: BalanceOf<T>) -> TransactionPriority {
+		let weight_saturation = T::BlockWeights::get().max_block / info.weight.max(1);
+		let max_block_length = *T::BlockLength::get().max.get(DispatchClass::Normal);
```
tomusdrw, line 160:
```suggestion
		let max_block_length = *T::BlockLength::get().max.get(info.class);
```
Why is it always `Normal`? Maybe we should get a total length instead? With current implementation we might have large non-normal transactions getting their priority heavily reduced compared to normal transactions.

pallets/asset-tx-payment/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+		let len_saturation = max_block_length as u64 / (len as u64).max(1);
+		let coefficient: BalanceOf<T> = weight_saturation.min(len_saturation).saturated_into::<BalanceOf<T>>();
```
tomusdrw, line 162: I think this should be at least `1`. For larger-than-block transactions (`Mandatory`) the `coefficient` might end up being `0`.

pallets/asset-tx-payment/src/payment.rs
```diff
@@ -0,0 +1,125 @@
...
+use pallet_assets::BalanceConversion;
+use sp_runtime::{
+	traits::{AtLeast32BitUnsigned, DispatchInfoOf, MaybeSerializeDeserialize, PostDispatchInfoOf, Saturating, Zero},
+	transaction_validity::InvalidTransaction,
+};
+use sp_std::{fmt::Debug, marker::PhantomData};
+
+pub trait HandleCredit<AccountId, B: Balanced<AccountId>> {
+	fn handle_credit(credit: CreditOf<AccountId, B>);
+}
```
tomusdrw, line 18: docs?

pallets/asset-tx-payment/src/payment.rs
```diff
@@ -0,0 +1,125 @@
...
+	/// Note: The `fee` already includes the `tip`.
+	fn withdraw_fee(
+		who: &T::AccountId,
+		_call: &T::Call,
+		_info: &DispatchInfoOf<T::Call>,
+		asset_id: Self::AssetId,
+		fee: Self::Balance,
+		_tip: Self::Balance,
+	) -> Result<Self::LiquidityInfo, TransactionValidityError> {
+		let converted_fee = CON::to_asset_balance(fee, asset_id)
```
tomusdrw, line 89: I guess this is what will handle the conversion rate between token and a particular asset, right? It's going to be oracle-based (or dex-based?)?

pallets/asset-tx-payment/src/payment.rs
```diff
@@ -0,0 +1,125 @@
...
+			.map_err(|_| -> TransactionValidityError { InvalidTransaction::Payment.into() })?;
```
tomusdrw, line 90:
```suggestion
			.map_err(|_| TransactionValidityError::from(InvalidTransaction::Payment))?;
```
Previous notation felt a bit confusing :)

pallets/asset-tx-payment/src/payment.rs
```diff
@@ -0,0 +1,125 @@
...
+		let can_withdraw = <T::Fungibles as Inspect<T::AccountId>>::can_withdraw(asset_id.into(), who, converted_fee);
+		if !matches!(can_withdraw, WithdrawConsequence::Success) {
+			return Err(InvalidTransaction::Payment.into());
+		}
+		<T::Fungibles as Balanced<T::AccountId>>::withdraw(asset_id.into(), who, converted_fee)
+			.map_err(|_| -> TransactionValidityError { InvalidTransaction::Payment.into() })
```
tomusdrw, line 96:
```suggestion
			.map_err(|_| TransactionValidityError::from(InvalidTransaction::Payment))
```

pallets/asset-tx-payment/src/payment.rs
```diff
@@ -0,0 +1,125 @@
...
+		paid: Self::LiquidityInfo,
+	) -> Result<(), TransactionValidityError> {
+		let converted_fee = CON::to_asset_balance(corrected_fee, paid.asset().into())
+			.map_err(|_| -> TransactionValidityError { InvalidTransaction::Payment.into() })?;
+		// Calculate how much refund we should return
+		let (final_fee, refund) = paid.split(converted_fee);
+		// refund to the the account that paid the fees. If this fails, the
+		// account might have dropped below the existential balance. In
+		// that case we don't refund anything.
+		// TODO: what to do in case this errors?
```
tomusdrw, line 119: I guess this should be addressed? Under what conditions the errors can happen?

pallets/asset-tx-payment/src/tests.rs
```diff
@@ -0,0 +1,819 @@
...
+			ChargeAssetTxPayment::<Runtime>
+				::post_dispatch(pre, &info_from_weight(weight), &post_info_from_weight(final_weight), len, &Ok(()))
+		);
+		let final_fee = fee_with_tip - (weight - final_weight) * 2;
+		assert_eq!(Assets::balance(asset_id, caller), balance - (final_fee));
+		assert_eq!(Assets::balance(asset_id, BLOCK_AUTHOR), fee + final_fee);
+	});
+}
+
+// 	#[test]
```
tomusdrw, line 385: What's up with these?

polkadot-parachains/statemint-common/src/impls.rs
```diff
@@ -16,17 +16,19 @@
 //! Auxillary struct/enums for Statemint runtime.
 //! Taken from polkadot/runtime/common (at a21cd64) and adapted for Statemint.
 
-use frame_support::traits::{Currency, Imbalance, OnUnbalanced};
+use frame_support::traits::{Currency, Imbalance, OnUnbalanced, fungibles::{self, CreditOf}};
+use sp_std::marker::PhantomData;
 
 pub type NegativeImbalance<T> = <pallet_balances::Pallet<T> as Currency<<T as frame_system::Config>::AccountId>>::NegativeImbalance;
 
+pub type AccountIdOf<R> = <R as frame_system::Config>::AccountId;
```
tomusdrw, line 24: docs?

polkadot-parachains/statemint-common/src/impls.rs
```diff
@@ -43,12 +45,11 @@ where
 	}
 }
 
-pub struct DealWithFees<R>(sp_std::marker::PhantomData<R>);
+pub struct DealWithFees<R>(PhantomData<R>);
```
tomusdrw, line 48: docs?

polkadot-parachains/statemint-common/src/impls.rs
```diff
@@ -61,6 +62,21 @@ where
 	}
 }
 
+pub struct AssetsToBlockAuthor<R>(PhantomData<R>);
```
tomusdrw, line 65: docs?

polkadot-parachains/statemint-common/src/impls.rs
```diff
@@ -61,6 +62,21 @@ where
...
+impl<R> pallet_asset_tx_payment::HandleCredit<AccountIdOf<R>, pallet_assets::Pallet<R>> for AssetsToBlockAuthor<R>
+where
+	R: pallet_authorship::Config + pallet_assets::Config,
+	AccountIdOf<R>: From<polkadot_primitives::v1::AccountId> + Into<polkadot_primitives::v1::AccountId>,
+{
+	fn handle_credit(credit: CreditOf<AccountIdOf<R>, pallet_assets::Pallet<R>>) {
+		let author = pallet_authorship::Pallet::<R>::author();
+		// TODO: what to do in case paying the author fails (e.g. because `fee < min_balance`)
```
tomusdrw, line 73: Is this really a TODO?

polkadot-parachains/statemint-runtime/src/lib.rs
```diff
@@ -659,7 +665,7 @@ pub type SignedExtra = (
 	frame_system::CheckEra<Runtime>,
 	frame_system::CheckNonce<Runtime>,
 	frame_system::CheckWeight<Runtime>,
-	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
+	pallet_asset_tx_payment::ChargeAssetTxPayment<Runtime>,
```
tomusdrw, line 668: Doesn't this require a transaction version bump or sth?

pallets/asset-tx-payment/src/lib.rs
tomusdrw: Could this be made a standalone function?

pallets/asset-tx-payment/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+			BalanceOf<T>,
+			InitialPayment<T>,
+		),
+		TransactionValidityError,
+	> {
+		let tip = self.0;
+		let fee = pallet_transaction_payment::Module::<T>::compute_fee(len as u32, info, tip);
+
+		if fee.is_zero() {
+			return Ok((fee, InitialPayment::Nothing));
```
tomusdrw, line 134: What about the `tip`? Is it guaranteed that if `fee == 0` then `tip == 0` as well?

pallets/asset-tx-payment/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+	/// final priority is `fee * min(2, 4) = fee * 2`. If it consumed `1/4th` of the block length
+	/// and the entire block weight `(1/1)`, its priority is `fee * min(1, 4) = fee * 1`. This means
+	///  that the transaction which consumes more resources (either length or weight) with the same
+	/// `fee` ends up having lower priority.
+	fn get_priority(len: usize, info: &DispatchInfoOf<T::Call>, final_fee: BalanceOf<T>) -> TransactionPriority {
+		let weight_saturation = T::BlockWeights::get().max_block / info.weight.max(1);
+		let max_block_length = *T::BlockLength::get().max.get(DispatchClass::Normal);
+		let len_saturation = max_block_length as u64 / (len as u64).max(1);
+		let coefficient: BalanceOf<T> = weight_saturation.min(len_saturation).saturated_into::<BalanceOf<T>>();
+		final_fee.saturating_mul(coefficient).saturated_into::<TransactionPriority>()
```
tomusdrw, line 163: Did it work the same way in `pallet_transaction_payment`? I'm not sure if having linear relation here is good. For super small transactions (both len and weight) the priority increase seems to be extremely high (absolute values of `min(block_length, block_weight)`. Given that the only way to counter that is with a tip, I feel it might not be tuned well (like there is no relation between extra `tip` cost to frontrun such transaction and absolute value of `block_weight/block_length`).

tomusdrw: Ugh, just realised the whole code here is actually the same as in `pallet_transaction_payment` and I happen to be the author. So I guess the discussion is not relevant to this PR and can continued somewhere else.

pallets/asset-tx-payment/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+		match initial_payment {
+			InitialPayment::Native(already_withdrawn) => {
+				<<T as pallet_transaction_payment::Config>::OnChargeTransaction as OnChargeTransaction<T>>::correct_and_deposit_fee(
+					&who, info, post_info, actual_fee, tip, already_withdrawn)?;
+			},
+			InitialPayment::Asset(already_withdrawn) => {
+				T::OnChargeAssetTransaction::correct_and_deposit_fee(
+					&who, info, post_info, actual_fee.into(), tip.into(), already_withdrawn.into())?;
+			},
+			// TODO: do anything else than just asserting that actual_fee is zero?
```
tomusdrw, line 248: `tip` maybe?

pallets/asset-tx-payment/src/payment.rs
```diff
@@ -0,0 +1,125 @@
...
+	fn handle_credit(credit: CreditOf<AccountId, B>);
+}
+
+/// Handle withdrawing, refunding and depositing of transaction fees.
+pub trait OnChargeAssetTransaction<T: Config> {
+	/// The underlying integer type in which fees are calculated.
+	type Balance: AtLeast32BitUnsigned + FullCodec + Copy + MaybeSerializeDeserialize + Debug + Default;
+	type AssetId: FullCodec + Copy + MaybeSerializeDeserialize + Debug + Default + Eq;
+	type LiquidityInfo;
+
```
tomusdrw, line 26: docs?

pallets/asset-tx-payment/src/lib.rs
apopiak: No, it depends on details of the transaction payment pallet (notably the `FeeMultiplier`).
We could introduce a trait and implement it for the tx payment pallet.

pallets/asset-tx-payment/src/payment.rs
apopiak: Right now it's based on the ratio between minimum balances.
It could be oracle-based in the future (DEX-based seems very unlikely)

pallets/asset-tx-payment/src/tests.rs
apopiak: Sorry, realised after asking for review that I left these :grimacing: 
They're from the original transaction-payment pallet and meant as inspiration. Will convert to tests for this pallet or remove.

pallets/asset-tx-payment/src/lib.rs
apopiak: `fee` here includes the tip, so yes

pallets/asset-tx-payment/src/lib.rs
apopiak: yeah, I copied it :see_no_evil:

pallets/asset-tx-payment/src/lib.rs
tomusdrw: Worth adding it to condition or asserting in case `compute_fee` ever changes?

pallets/asset-tx-payment/src/lib.rs
apopiak: So should we adjust this in `transaction-payment` and then just use the implementation here?

pallets/asset-tx-payment/src/lib.rs
shawntabrizi: I think we def should use a trait here, and not force us to have the pallets be tightly coupled.

We could even duplicate some code here if we should consider this pallet to just be different than the normal transaction pallet?

pallets/asset-tx-payment/src/payment.rs
apopiak: The comment above explains the scenario. It's a valid strategy to just burn the refund if the account stopped existing IMO.
Just wanted to flag this so it's reconsidered before merge. What do you think?

polkadot-parachains/statemint-common/src/impls.rs
apopiak: Not really. Just wanted to consider this again.
Will fix by having a more sophisticated (pot-based) fee dispersal later.

pallets/asset-tx-payment/src/lib.rs
apopiak: mmh, but they *are* tightly coupled in semantics as well as code. This is just a wrapper around the regular tx payment pallet that extends it. It's not meant to be used with any other pallet.

pallets/asset-tx-payment/src/lib.rs
apopiak: I'll create an issue

pallets/asset-tx-payment/src/tests.rs
apopiak: :heavy_check_mark:

pallets/asset-tx-payment/src/lib.rs
apopiak: https://github.com/paritytech/substrate/issues/9317

pallets/asset-tx-payment/src/lib.rs
apopiak: created issue here: https://github.com/paritytech/substrate/issues/9317

pallets/asset-tx-payment/src/lib.rs
apopiak: created issue here: https://github.com/paritytech/substrate/issues/9317

shawntabrizi (approved):
i think it is in a good spot

pallets/asset-tx-payment/src/tests.rs
```diff
@@ -0,0 +1,511 @@
+
```
bkchr, line 1: License header missing

pallets/asset-tx-payment/src/payment.rs
```diff
@@ -0,0 +1,141 @@
+///! Traits and default implementation for paying transaction fees in assets.
```
bkchr, line 1: Same.

pallets/asset-tx-payment/src/lib.rs
```diff
@@ -0,0 +1,277 @@
+// This file is part of Substrate.
```
bkchr, line 1: This is not Substrate :P

pallets/asset-tx-payment/Cargo.toml
```diff
@@ -0,0 +1,51 @@
+[package]
+name = "pallet-asset-tx-payment"
+version = "3.0.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2018"
+license = "Apache-2.0"
+homepage = "https://substrate.dev"
+repository = "https://github.com/paritytech/substrate/"
```
bkchr, line 8: :P

pallets/asset-tx-payment/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+description = "FRAME pallet to manage transaction payments"
```
bkchr, line 9: C&P

pallets/asset-tx-payment/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/// Require the transactor pay for themselves and maybe include a tip to gain additional priority
+/// in the queue. Allows paying via both `Currency` as well as `fungibles::Balanced`.
+///
+/// Wraps the transaction logic in `pallet_transaction_payment` and extends it with assets.
+/// An asset id of `None` falls back to the underlying transaction payment via the native currency.
+#[derive(Encode, Decode, Clone, Eq, PartialEq, TypeInfo)]
+#[scale_info(skip_type_params(T))]
+pub struct ChargeAssetTxPayment<T: Config>(#[codec(compact)] BalanceOf<T>, Option<ChargeAssetIdOf<T>>);
```
bkchr, line 120:
```suggestion
pub struct ChargeAssetTxPayment<T: Config> { #[codec(compact)] BalanceOf<T>, asset_id: Option<ChargeAssetIdOf<T>> };
```

And are we happy that the tx size increases by `size_of(AssetId)`?

pallets/asset-tx-payment/Cargo.toml
apopiak: :grimacing: good catch, thanks

pallets/asset-tx-payment/src/lib.rs
apopiak: For Statemint? Yes, I would say so, or do you mean it should be compact?

pallets/asset-tx-payment/src/payment.rs
apopiak: @tomusdrw @shawntabrizi I (finally!) ran a test via Polkadot-js and ran into a practical limitation: even if the conversion is mathematically correct it can end up rounding down the fee to 0 if the asset has low precision. I wonder whether we should
1. just not allow assets with low precision (i.e. low `min_balance`) to be `is_sufficient`
2. OR add `.max(1)` here

Opinions?

tomusdrw: Couldn't we reject the transaction (i.e. return it as invalid) in case it returns a `0` fee? I think that's the most sensible and would also protect us in the future in case the `to_asset_balance` method is implemented differently.

apopiak: I would say doing `.max(1)` ist strictly better because the transaction is not invalid and the fee is calculated "correctly" (the appropriate fee would be less than 1 token).

apopiak: Though, yes, we could reject the transaction in principle.

pallets/asset-tx-payment/src/lib.rs
bkchr: Yes for statemint. I just know that we try to minimize the size per tx as much as possible, but I also know that you can not do it differently. So, maybe at least use a compact as you proposed.

apopiak: I tried making it compact and that does not seem to be supported. I also saw @thiolliere mention as much recently for `#[pallet::compact]` (which seems the same issue?)

apopiak:
I would like to merge this PR.
Only open conversations are https://github.com/paritytech/cumulus/pull/488#discussion_r720174723 and https://github.com/paritytech/cumulus/pull/488#discussion_r716392717 and for both I tend towards staying with the status quo of the PR.

pallets/asset-tx-payment/src/payment.rs
```diff
@@ -0,0 +1,159 @@
...
+		paid: Self::LiquidityInfo,
+	) -> Result<(), TransactionValidityError> {
+		// Convert the corrected fee into the asset used for payment.
+		let converted_fee = CON::to_asset_balance(corrected_fee, paid.asset().into())
+			.map_err(|_| -> TransactionValidityError { InvalidTransaction::Payment.into() })?;
+		// Calculate how much refund we should return.
+		let (final_fee, refund) = paid.split(converted_fee);
+		// Refund to the account that paid the fees. If this fails, the account might have dropped
+		// below the existential balance. In that case we don't refund anything.
+		let _res = <T::Fungibles as Balanced<T::AccountId>>::resolve(who, refund);
```
bkchr, line 154:
```suggestion
		let _ = <T::Fungibles as Balanced<T::AccountId>>::resolve(who, refund);
```

pallets/asset-tx-payment/src/tests.rs
```diff
@@ -0,0 +1,556 @@
...
+	type EventHandler = ();
+}
+
+pub struct CreditToBlockAuthor;
+impl HandleCredit<AccountId, Assets> for CreditToBlockAuthor {
+	fn handle_credit(credit: CreditOf<AccountId, Assets>) {
+		let author = pallet_authorship::Pallet::<Runtime>::author();
+		// TODO: what to do in case paying the author fails (e.g. because `fee < min_balance`)
+		// default: drop the result which will trigger the `OnDrop` of the imbalance.
+		let _res = <Assets as Balanced<AccountId>>::resolve(&author, credit);
```
apopiak, line 205:
```suggestion
		let _ = <Assets as Balanced<AccountId>>::resolve(&author, credit);
```

PR #498, opened 2021-06-17
min collator check

JesseAbram:
This PR checks to make sure that there is a min amount of collators before a collator gets kicked, or tries to leave. This is a defence mechanism for better disaster recovery of a chain

### DIFF
--- a/pallets/collator-selection/src/lib.rs
+++ b/pallets/collator-selection/src/lib.rs
@@ -133,7 +133,7 @@ pub mod pallet {
 		/// Minimum number of candidates that we should have. This is used for disaster recovery.
 		///
 		/// This does not take into account the invulnerables.
-		type MinCandidates: Get<usize>;
+		type MinCandidates: Get<u32>;
 
 
 		/// Maximum number of invulnerables.
--- a/pallets/collator-selection/src/lib.rs
+++ b/pallets/collator-selection/src/lib.rs
@@ -375,7 +375,7 @@ pub mod pallet {
 			let new_candidates = candidates.into_iter().filter_map(|c| {
 				let last_block = <LastAuthoredBlock<T>>::get(c.who.clone());
 				let since_last = now.saturating_sub(last_block);
-				if since_last < kick_threshold || Self::candidates().len() <= T::MinCandidates::get() {
+				if since_last < kick_threshold || Self::candidates().len() as u32 <= T::MinCandidates::get() {
 					Some(c.who)
 				} else {
 					let outcome = Self::try_remove_candidate(&c.who);
--- a/pallets/collator-selection/src/lib.rs
+++ b/pallets/collator-selection/src/lib.rs
@@ -79,7 +79,7 @@ pub mod pallet {
 		pallet_prelude::*,
 		inherent::Vec,
 		traits::{
-			Currency, ReservableCurrency, EnsureOrigin, ExistenceRequirement::KeepAlive,
+			Currency, ReservableCurrency, EnsureOrigin, ExistenceRequirement::KeepAlive, ValidatorRegistration
 		},
 		PalletId,
 	};
@@ -92,6 +92,7 @@ pub mod pallet {
 		},
 		weights::DispatchClass,
 	};
+	use sp_runtime::traits::Convert;
 	use core::ops::Div;
 	use pallet_session::SessionManager;
 	use sp_staking::SessionIndex;
@@ -143,6 +144,18 @@ pub mod pallet {
 
 		// Will be kicked if block is not produced in threshold.
 		type KickThreshold: Get<Self::BlockNumber>;
+
+		/// A stable ID for a validator.
+		type ValidatorId: Member + Parameter;
+
+		/// A conversion from account ID to validator ID.
+		///
+		/// Its cost must be at most one storage read.
+		type ValidatorIdOf: Convert<Self::AccountId, Option<Self::ValidatorId>>;
+
+		/// Validate a user is registered
+		type ValidatorRegistration: ValidatorRegistration<Self::ValidatorId>;
+
 
 		/// The weight information of this pallet.
 		type WeightInfo: WeightInfo;
@@ -261,6 +274,10 @@ pub mod pallet {
 		NotCandidate,
 		/// User is already an Invulnerable
 		AlreadyInvulnerable,
+		/// Account has no associated validator ID
+		NoAssociatedValidatorId,
+		/// Validator ID is not yet registered
+		ValidatorNotRegistered
 	}
 
 	#[pallet::hooks]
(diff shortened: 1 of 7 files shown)


polkadot-parachains/statemine-runtime/src/lib.rs
```diff
@@ -638,6 +638,7 @@ impl pallet_aura::Config for Runtime {
 parameter_types! {
 	pub const PotId: PalletId = PalletId(*b"PotStake");
 	pub const MaxCandidates: u32 = 1000;
+	pub const MinCandidates: usize = 1;
```
bkchr, line 641: Maybe 5?

bkchr (approved):
Looks good, beside the minimum number of collators

polkadot-parachains/statemine-runtime/src/lib.rs
JesseAbram: ya, I had no idea, that sounds fine to me, although one small thing is that it won't allow people to exit with their deposits until others join in (if they are below the 5 limit), which seems fine to me I just want to bring it up

JesseAbram: actually gonna add it to the pallet documentation with the amount change

pallets/collator-selection/src/lib.rs
```diff
@@ -127,6 +130,12 @@ pub mod pallet {
 		/// This does not take into account the invulnerables.
 		type MaxCandidates: Get<u32>;
 
+		/// Minimum number of candidates that we should have. This is used for disaster recovery.
+		///
+		/// This does not take into account the invulnerables.
+		type MinCandidates: Get<usize>;
```
shawntabrizi, line 136:
```suggestion
		type MinCandidates: Get<u32>;
```

Should keep this as u32 for best practice

pallets/collator-selection/src/lib.rs
```diff
@@ -323,7 +333,7 @@ pub mod pallet {
 		#[pallet::weight(T::WeightInfo::leave_intent(T::MaxCandidates::get()))]
 		pub fn leave_intent(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
-
+			ensure!(Self::candidates().len() > T::MinCandidates::get(), Error::<T>::TooFewCandidates);
```
shawntabrizi, line 336:
```suggestion
			ensure!(Self::candidates().len() as u32 > T::MinCandidates::get(), Error::<T>::TooFewCandidates);
```

pallets/collator-selection/src/lib.rs
```diff
@@ -365,7 +375,7 @@ pub mod pallet {
 			let new_candidates = candidates.into_iter().filter_map(|c| {
 				let last_block = <LastAuthoredBlock<T>>::get(c.who.clone());
 				let since_last = now.saturating_sub(last_block);
-				if since_last < kick_threshold {
+				if since_last < kick_threshold || Self::candidates().len() <= T::MinCandidates::get() {
```
shawntabrizi, line 378:
```suggestion
				if since_last < kick_threshold || Self::candidates().len() <= T::MinCandidates::get() {
```

Why are you reading candidates here but also getting candidates from the input params of this function?

pallets/collator-selection/src/lib.rs
```diff
@@ -239,6 +248,7 @@ pub mod pallet {
 	#[pallet::error]
 	pub enum Error<T> {
 		TooManyCandidates,
+		TooFewCandidates,
 		Unknown,
 		Permission,
 		AlreadyCandidate,
```
shawntabrizi, line 254: its nice to have doc comments on these as they appear in the metadata and show up in the UI

polkadot-parachains/westmint-runtime/src/lib.rs
```diff
@@ -576,6 +576,7 @@ impl pallet_aura::Config for Runtime {
 parameter_types! {
 	pub const PotId: PalletId = PalletId(*b"PotStake");
 	pub const MaxCandidates: u32 = 1000;
+	pub const MinCandidates: usize = 1;
```
shawntabrizi, line 579:
```suggestion
	pub const MinCandidates: u32 = 1;
```

polkadot-parachains/statemint-runtime/src/lib.rs
```diff
@@ -579,6 +579,7 @@ impl pallet_aura::Config for Runtime {
 parameter_types! {
 	pub const PotId: PalletId = PalletId(*b"PotStake");
 	pub const MaxCandidates: u32 = 1000;
+	pub const MinCandidates: usize = 5;
```
shawntabrizi, line 582:
```suggestion
	pub const MinCandidates: u32 = 5;
```

polkadot-parachains/statemint-common/src/impls.rs
```diff
@@ -149,6 +149,7 @@ mod tests {
 		pub const PotId: PalletId = PalletId(*b"PotStake");
 		pub const MaxCandidates: u32 = 20;
 		pub const MaxInvulnerables: u32 = 20;
+		pub const MinInvulnerables: usize = 1;
```
shawntabrizi, line 152:
```suggestion
		pub const MinInvulnerables: u32 = 1;
```

polkadot-parachains/statemine-runtime/src/lib.rs
```diff
@@ -638,6 +638,7 @@ impl pallet_aura::Config for Runtime {
 parameter_types! {
 	pub const PotId: PalletId = PalletId(*b"PotStake");
 	pub const MaxCandidates: u32 = 1000;
+	pub const MinCandidates: usize = 5;
```
shawntabrizi, line 641:
```suggestion
	pub const MinCandidates: u32 = 5;
```

pallets/collator-selection/src/lib.rs
JesseAbram: so the candidates input does not get changed, it gets at the end collected into new_candidates, so if 5 people get removed, it does not know that, however the global candidates is tracking this, so it would for each candidate have the current knowledge of how many are left after the previous have been kicked

pallets/collator-selection/src/lib.rs
```diff
@@ -127,6 +130,12 @@ pub mod pallet {
 		/// This does not take into account the invulnerables.
 		type MaxCandidates: Get<u32>;
 
+		/// Minimum number of candidates that we should have. This is used for disaster recovery.
+		///
+		/// This does not take into account the invulnerables.
```
shawntabrizi, line 135: You are saying this count is the number of people + invulnerables?

I think that is over doing it. Invulnerables existing should satisfy the requirement for a minimum number of collators.

JesseAbram: I agree in theory but invulnerables can be removed all at once (through a permissioned call) I was thinking this would be more of a assume invulnerables would be removed soon thing would be an easier transition, but open to changing

xlc: Invulnerables should eventually be removed and by then minimum number of collators will be super useful

pallets/collator-selection/src/lib.rs
```diff
@@ -323,7 +333,7 @@ pub mod pallet {
 		#[pallet::weight(T::WeightInfo::leave_intent(T::MaxCandidates::get()))]
 		pub fn leave_intent(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
 			let who = ensure_signed(origin)?;
-
+			ensure!(Self::candidates().len() as u32 > T::MinCandidates::get(), Error::<T>::TooFewCandidates);
```
shawntabrizi, line 336: shouldn't this logic be in `try_remove_candidate`?

Else it may not be checked everywhere that this is happening correctly.

JesseAbram: I had originally looked at that, but try_remove_candidates is used in kick_candidates and that would track two items, the candidates array and the current candidates to be accumulated as collators. So if one failed while the other passed they would not track each other correctly. So it was either rewrite the function, or add this here, this was less code and I don't see how this would not cause it to track properly.

shawntabrizi (approved):

PR #504, opened 2021-06-23
Implement Asset Teleport for Statemint

apopiak:
This PR implements asset teleports from one instance of Statemint to another.

Open questions:
+ What should be the policy for sending and receiving teleports for Statemint?
+ When/Where does the conversion to the local asset schema happen? (This implementation just maps asset ids to one another directly, i.e. `asset(42)` on one instance is `asset(42)` on the other.)

depends on https://github.com/paritytech/polkadot/pull/3345


client/consensus/relay-chain/src/lib.rs
```diff
@@ -208,7 +208,7 @@ where
 
 		let mut block_import_params = BlockImportParams::new(BlockOrigin::Own, header);
 		block_import_params.body = Some(extrinsics);
-		block_import_params.storage_changes = Some(storage_changes);
+		// block_import_params.storage_changes = Some(storage_changes);
```
apopiak, line 211: just a quick fix to be able to use update Substrate
TODO: remove (probably by pulling in master)

xlc:
What's the use case for this? You can't teleport assets to relaychain. You can't teleport assets to other non system/common good parachains. I don't see any plans to have another system/common good parachain.

On the other hand, reserved based transfer is required to support assets transfer to other community parachains.

apopiak:
Mmmh, I would've thought that there will be support for transfer to community parachains (on the parachains' side).
You guys won't support teleport from Statemint?
You think that's only feasible via reserve based transfer?

xlc:
I am pretty sure @gavofyork said somewhere teleport is between trusted chains only and community parachain won’t be a trusted chain of system/common good/relaychain. So not we don’t want to support it, but for security purpose Statemine shouldn’t support it. Otherwise how could you prevent a community chain just mint extra assets and transfer back to Statemine?

gavofyork:
Precisely. There should be no changes to Statemine regarding teleports at present.

PR #507, opened 2021-06-24
Refactor paths and runtime names

chevdor:
This PR renames the rococo runtime to follow the pattern used by all other chains (in this repo and the pollkadot repo).

The PR also uses the same folder tree (= runtimes under a `runtime/<name>`) folder. That simplifies tooling and allows building all the runtimes the same way.

Once this PR is merged, I will be able to extend the CI to build statemint, shell and rococo using srtool (see #509)


bkchr (changes requested):
I would be fine to move this to `polkadot-parachains/runtime/*`, but not moving it to the top

chevdor:
That works for me. The most important is that:
- the rococo runtime is named `rococo-runtime`
- the runtime `<foobar>` in the `runtime_dir` (in our case `polkadot-parachains` or `polkadot-parachains/runtimes`) as a `foobar` folder (vs `foobar-runtime` currently).

It is easier if the `runtime_dir` is a folder called `runtime` as in Polkadot but having it a level down as you suggest is not a big issue. 

Let me adjust to your suggestion.

chevdor:
I don't see a benefit going from `polkadot-parachains/*` to `polkadot-parachains/runtime/*` so I will bring the runtimes back to `polkadot-parachains/*`

chevdor:
@bkchr would that work for you now ?

polkadot-parachains/Cargo.toml
```diff
@@ -23,11 +23,11 @@ hex-literal = "0.2.1"
 async-trait = "0.1.42"
 
 # Parachain runtimes
-rococo-parachain-runtime = { package = "cumulus-rococo-parachain-runtime", path = "rococo-runtime" }
-shell-runtime = { package = "cumulus-shell-runtime", path = "shell-runtime" }
-statemint-runtime = { path = "statemint-runtime" }
-statemine-runtime = { path = "statemine-runtime" }
-westmint-runtime = { path = "westmint-runtime" }
+rococo-parachain-runtime = { package = "rococo-runtime", path = "rococo" }
+shell-runtime = { package = "shell-runtime", path = "shell" }
```
bkchr, line 27:
```suggestion
shell-runtime = { path = "shell" }
```

bkchr (approved):

bkchr:
bot merge

PR #516, opened 2021-06-28
Propose bumping StringLimit to 128 bytes

Swader:
The StringLimit of 50 bytes is too short for useful hashes of different protocols:

- When using IPFS, the hash is 46 bytes. With a protocol prefix and type prefix (IPFS vs IPNS) it becomes: `ipfs://ipfs/QmPK1s3pNYLi9ERiq3BDxKa4XosgWwFRQUydHUtz4YgpqB` or 59 bytes. With CIDv1, the hashes are even longer, and with the prefixes they become 71 chars long: `ipfs://ipfs/bafybeierhgbz4zp2x2u67urqrgfnrnlukciupzenpqpipiz5nwtq7uxpx4`.
- On arweave, the hash is 43 bytes. Prefixed with just `arweave://` bumps it up to 53: `arweave://BNttzDav3jHVnNiV7nYbQv-GY0HQ-4XXsdkE5K9ylHQ`.
- On Sia, with 46 char hashes, adding `sia://` as the protocol to use will bump it past 50: `sia://GACjmEWXmYF1N3Rc-PyjN304-8M0zOXHYzAXY9222xkGhA` (52)

To allow for static URLs which are typically longer than 50 bytes, and also to cover all of the above hashes and protocols, we feel that 128 bytes is a reasonable minimum for the `StringLimit`.


joepetrowski:
I'm fine with this, `StringLimit` was added in the context of storing a string along with `pallet-assets`. Before merging, we can either:

- Keep this change and re-benchmark assets (cc @shawntabrizi ) or 
- Add a `StringLimit` specific to `pallet-uniques` to cover this issue.

shawntabrizi:
@Swader stringlimit is the wrong thing to change here. It has nothing to do with the metadata length

shawntabrizi:
> Add a StringLimit specific to pallet-uniques to cover this issue.

This is what you should do. Please make a new constant specific for this.

shawntabrizi (approved):

joepetrowski (approved):

PR #517, opened 2021-06-28
Look at the upgrade go-ahead and restriction signals

pepyakin:
This is a patch that changes when parachain applies the upgrade. Specifically, this implements the scheme introduced by this [patch](https://github.com/paritytech/polkadot/pull/3371) on the polka runtimes (specifically for rococo and kusama).

For now it's only a preview. This should land only after the patch above is landed and deployed.

# Migration

This is a breaking change.

As soon as we merge this patch, cumulus will be looking for a specific signal by the relay-chain in order to perform the last step in the code upgrade - assigning the `:code` property. This is in contrast to the previous approach where the switch happened after a fixed number of relay-chain blocks after signalling the upgrade.

Note that the old cumulus version should work just fine with the updated relay-chain runtime version.

This is good to proceed now, since Kusama was updated. Rococo runtime is still to be updated.


pallets/parachain-system/src/lib.rs
```diff
@@ -441,21 +434,24 @@ pub mod pallet {
 		Unauthorized,
 	}
 
-	/// We need to store the new validation function for the span between
-	/// setting it and applying it. If it has a
-	/// value, then [`PendingValidationCode`] must have a real value, and
-	/// together will coordinate the block number where the upgrade will happen.
-	#[pallet::storage]
-	pub(super) type PendingRelayChainBlockNumber<T: Config> =
-		StorageValue<_, RelayChainBlockNumber>;
```
pepyakin, line 450: Just a note about this.

If we remove a storage value, we typically need to remove it from the storage. Otherwise, this value will become unreachable and will stay as garbage on-chain. However, if I am reading the code right, right after the previous version enacts the upgrade it will remove `PendingRelayChainBlockNumber`. Therefore, no clean up is needed.

pallets/parachain-system/src/lib.rs
```diff
@@ -465,9 +461,16 @@ pub mod pallet {
...
-	/// The last relay parent block number at which we signalled the code upgrade.
+	/// An option which indicates if the relay-chain restricts signalling a validation code upgrade.
+	/// In other words, if this is `Some` and [`NewValidationCode`] is `Some` then the produced
+	/// candidate will be invalid.
+	///
+	/// This storage item is a mirror of the corresponding value for the current parachain from the
+	/// relay-chain. This value is emphemeral which means it doesn't hit the storage. This value is
+	/// set after the inherent.
 	#[pallet::storage]
-	pub(super) type LastUpgrade<T: Config> = StorageValue<_, relay_chain::BlockNumber, ValueQuery>;
```
pepyakin, line 470: Since this storage value is removed, we need to make sure this is removed. Otherwise, it will stay inaccessible in the storage as garbage.

pallets/parachain-system/src/lib.rs
```diff
@@ -321,25 +302,36 @@ pub mod pallet {
...
 			// sequencing problem.
-			if let Some(apply_block) = <PendingRelayChainBlockNumber<T>>::get() {
-				if vfp.relay_parent_number >= apply_block {
-					<PendingRelayChainBlockNumber<T>>::kill();
+			let upgrade_go_ahead_signal = relay_state_proof
+				.read_upgrade_go_ahead_signal()
+				.expect("Invalid upgrade go ahead signal");
+			match upgrade_go_ahead_signal {
+				Some(relay_chain::v1::UpgradeGoAhead::GoAhead) => {
 					let validation_code = <PendingValidationCode<T>>::take();
```
slumber, line 320: Shouldn't we double-check that the entry is present in the storage before taking it?

pepyakin: Like in the case the relay-chain flashed us with `GoAhead` when none was expected? Hm, theoretically that should never happen. Although I can see how a logic error be it on the relay-chain or the parachain can lead to that. If that happens the parachain is screwed most likely, but perhaps it won't hurt if we try to save the day here.

pepyakin:
✅  I looked through it and this looks fine. I cannot put a review of my own though because I am the author of the PR.

In anycase, I would feel more comfortable if we get another pair of eyes.

apopiak (approved):
Looks fine to me. Migration seems simple enough.

test/relay-sproof-builder/src/lib.rs
```diff
@@ -17,12 +17,13 @@
...
 	relay_chain, AbridgedHostConfiguration, AbridgedHrmpChannel, ParaId,
 };
+use polkadot_primitives::v1::UpgradeGoAhead;
 use sp_runtime::traits::HashFor;
 use sp_state_machine::MemoryDB;
 use sp_std::collections::btree_map::BTreeMap;
 
 /// Builds a sproof (portmanteau of 'spoof' and 'proof') of the relay chain state.
-#[derive(Clone)]
+// #[derive(Clone)]
```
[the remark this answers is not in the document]
slumber: This one is related to `TODO`, we cannot derive `Clone` since `UpgradeGoAhead` doesn't implement it.
To resolve it now, it's necessary to update polkadot repo and the PR as well.
It also can be done as a follow-up.

pallets/parachain-system/src/migration.rs
```diff
@@ -0,0 +1,45 @@
...
+
+/// The current storage version.
+pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+
+/// Call this during the next runtime upgrade for this module.
+pub fn on_runtime_upgrade<T: Config>() -> Weight {
+	let mut weight: Weight = 0;
+
+	if StorageVersion::get::<Pallet<T>>() == 0 {
+		weight = weight.saturating_add(v1::migrate::<T>());
```
bkchr, line 28:
```suggestion
		weight = weight.saturating_add(v1::migrate::<T>()).saturating_add(T::DbWeight::get().writes(1));
```

test/relay-sproof-builder/src/lib.rs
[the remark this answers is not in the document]
bkchr: Then remove it

pallets/parachain-system/src/lib.rs
```diff
@@ -441,22 +440,27 @@ pub mod pallet {
...
-	pub(super) type PendingRelayChainBlockNumber<T: Config> =
-		StorageValue<_, RelayChainBlockNumber>;
-
-	/// The new validation function we will upgrade to when the relay chain
-	/// reaches [`PendingRelayChainBlockNumber`]. A real validation function must
-	/// exist here as long as [`PendingRelayChainBlockNumber`] is set.
+	/// In case of a scheduled upgrade, this storage field contains the validation code to be applied.
+	///
+	/// As soon as the parachain gives us the go-ahead signal, we will overwrite the [`:code`][well_known_keys::CODE]
+	/// which will result the next block process with the new validation code. This concludes the upgrade process.
```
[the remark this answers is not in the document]
bkchr: Yes

pallets/parachain-system/src/lib.rs
```diff
@@ -441,22 +440,27 @@ pub mod pallet {
...
+	///
+	/// [well_known_keys::CODE]: [sp_core::storage::well_known_keys::CODE]
```
bkchr, line 448: There exists intra doc links

pallets/parachain-system/src/lib.rs
bkchr: This is not done yet?

bkchr (approved):

xlc:
This is a transaction version breaking change as `setUpgradeBlock` is removed.
It will cause issue if there are pending `authorizeUpgrade` in storage.
This also changes index of `setValidationData` which I am not sure if it is completely safe or not (looks like it should be safe).

bkchr:
> This also changes index of `setValidationData` which I am not sure if it is completely safe or not (looks like it should be safe).

This is fine.

PR #526, opened 2021-07-06
upgrade srtool-actions and subwasm

chevdor:
This PR brings:
- add builds for "statemint", "rococo", "shell"
- switch to `chevdor/srtool-actions@v0.3.0` which default the new srtool image from `paritytech/srtool`
- bumps up the version of `subwasm`
- cleanup
- ability to start builds on-demand

### DIFF
--- a/.github/workflows/srtool.yml
+++ b/.github/workflows/srtool.yml
@@ -14,59 +14,78 @@ on:
   schedule:
     - cron: "00 02 * * 1" # 2AM weekly on monday
 
+  workflow_dispatch:
+
 jobs:
   srtool:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


TriplEight (approved):
LGTM, but GitLab CI already runs `cargo test --all --release --locked`.
Can we do something to reuse the build caches for these jobs?

chevdor:
> Can we do something to reuse the build caches for these jobs?

That could be interesting to investigate the options but I admit I am very cautious when it comes to mixing up determinism (srtool) and caching. The goal is to produce the same output 100% of the time and not depend on external resources so I see a non neglictable risk using caching. Also, the frequency of those builds is rather low. Having this worfklow in place allows experimenting with caching though.

PR #527, opened 2021-07-06
CI: build a test artifact

TriplEight:
`build --release --bin polkadot-collator` takes 5 minutes after a test, `test --release --all` gets built in ~8 minutes without the cache.
So in the case of building after the test it saves a parallel/concurrent task, but certainly it doesn't make sense to run it every time on PRs, not even on every master pipeline.

I can run the build as it is, in the same job, but only when the pipeline's executed by schedule/trigger, @lovelaced WDYT?

Closes: https://github.com/paritytech/ci_cd/issues/141

### DIFF
--- a/scripts/ci/pre_cache.sh
+++ b/scripts/ci/pre_cache.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -u
+
+# if there is no directory for this $CI_COMMIT_REF_NAME/$CI_JOB_NAME
+# create such directory and
+# copy recursively all the files from the newest dir which has $CI_JOB_NAME, if it exists
+
+# cache lives in /ci-cache/${CI_PROJECT_NAME}/${2}/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}
+
+function prepopulate {
+  if [[ ! -d $1 ]]; then
+    mkdir -p "/ci-cache/$CI_PROJECT_NAME/$2/$CI_COMMIT_REF_NAME";
+    FRESH_CACHE=$(find "/ci-cache/$CI_PROJECT_NAME/$2" -mindepth 2 -maxdepth 2 \
+      -type d -name "$CI_JOB_NAME"  -exec stat --printf="%Y\t%n\n" {} \; |sort -n -r |head -1 |cut -f2);
+    if [[ -d $FRESH_CACHE ]]; then
+      echo "____Using" "$FRESH_CACHE" "to prepopulate the cache____";
+      time cp -r "$FRESH_CACHE" "$1";
+    else
+      echo "_____No such $2 dir, proceeding from scratch_____";
+    fi
+  else
+    echo "____No need to prepopulate $2 cache____";
+  fi
+}
+
+# CARGO_HOME cache is still broken so would be handled some other way.
+prepopulate "$CARGO_TARGET_DIR" targets
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -31,8 +31,6 @@ variables:                         &default-vars
     - if: $CI_PIPELINE_SOURCE == "web"
     - if: $CI_PIPELINE_SOURCE == "schedule"
     - if: $CI_COMMIT_REF_NAME =~ /^v[0-9]+\.[0-9]+.*$/              # i.e. v1.0, v2.1rc1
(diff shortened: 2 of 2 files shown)


scripts/ci/pre_cache.sh
```diff
@@ -0,0 +1,28 @@
...
+    else
+      echo "_____No such $2 dir, proceeding from scratch_____";
+    fi
+  else
+    echo "____No need to prepopulate $2 cache____";
+  fi
+}
+
+# CARGO_HOME cache is still broken so would be handled some other way.
+prepopulate "$CARGO_TARGET_DIR" targets
```
rcny, line 28: What does `prepopulate` command do? Or is it a typo?

scripts/ci/pre_cache.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -u
+
+# if there is no directory for this $CI_COMMIT_REF_NAME/$CI_JOB_NAME
+# create such directory and
+# copy recursively all the files from the newest dir which has $CI_JOB_NAME, if it exists
+
+# cache lives in /ci-cache/${CI_PROJECT_NAME}/${2}/${CI_COMMIT_REF_NAME}/${CI_JOB_NAME}
```
rcny, line 9: I'm still not sure how `ci-cache` directory correlates with the cache location for our bare metal runners. We have defined `ci-cache` as the volume for Docker executor [here](https://gitlab.parity.io/parity/devops/-/blob/master/ansible/roles/gitlab-runner/templates/config.toml.j2#L27), but we don't set `cache_dir` anywhere to use that volume as the location for caching (as per [docs](https://docs.gitlab.com/runner/configuration/advanced-configuration.html#the-runners-section) for `cache_dir` setting).

scripts/ci/pre_cache.sh
TriplEight: https://github.com/paritytech/cumulus/pull/527/files#diff-ec994b132c375339c6af70baf02771519b29ebfb102e2a49c62059652edf8517R11

scripts/ci/pre_cache.sh
TriplEight: we don't use GitLab's built-in cache, we attach our own as a [rw volume ](https://gitlab.parity.io/parity/devops/-/blob/master/ansible/roles/gitlab-runner/templates/config.toml.j2#L27) to each image.

scripts/ci/pre_cache.sh
rcny: Oh, sorry, dunno how I've missed that 😅

scripts/ci/pre_cache.sh
rcny: Right, thanks, that's something that we've talked through before, but what about the first volume, `/cache`? Is it used somewhere?

TriplEight: and that one is for GitLab's built-in, AFAIK it may misbehave when someone accidentally uses `cache:`.

rcny (approved):

TriplEight:
@bkchr @apopiak @lovelaced done, it's published automatically on tags to s3://releases.parity.io/cumulus/x86_64-debian:stretch/<tag name> AND **/latest** by default.
Also can set up a schedule if needed. It's just https://gitlab.parity.io/parity/cumulus/-/pipeline_schedules , choose cron, branch, add a non-empty `ARE_WE_RELEASING_YET` variable, after that binary will be found in **s3/.../branch** and **/latest**.

TriplEight:
Please raise a new issue with details on publishing the containerized version.

apopiak:
> @bkchr @apopiak @lovelaced done, it's published automatically on tags to s3://releases.parity.io/cumulus/x86_64-debian:stretch/ **[tag name]**/polkadot-collator AND **/latest**/polkadot-collator by default.
> Also can set up a schedule if needed. It's just https://gitlab.parity.io/parity/cumulus/-/pipeline_schedules , choose cron, branch, add a non-empty `ARE_WE_RELEASING_YET` variable, after that binary will be found in **s3/.../branch**/polkadot-collator and **/latest**/polkadot-collator.

DQ: Does that mean it's not published to github?

TriplEight:
@apopiak we didn't talk about publishing to GitHub yet, only testing release.

PR #552, opened 2021-07-27
Statemint Reserve Asset Transfer

apopiak:
This PR adds a `FungiblesAdapter` as an asset transactor to Statemint to allow for Reserve-based asset transfers of Statemint assets.

Todos
- [x] Add the same config to Statemine and Westmint


apopiak:
Config looks ok based on the xcm simulator test with a similar mock runtime here:
https://github.com/paritytech/polkadot/pull/3548/files#diff-0c1e73932c0f34b91c9dd40c479baa1f7550f4541ea15e74129f7fd94440da97R248

polkadot-parachains/statemine/src/lib.rs
```diff
@@ -498,6 +500,24 @@ pub type LocalAssetTransactor = CurrencyAdapter<
 	(),
 >;
 
+pub type FungiblesTransactor = FungiblesAdapter<
+	// Use this fungibles implementation:
+	Assets,
+	// Use this currency when it is a fungible asset matching the given location or name:
+	(
+		ConvertedConcreteAssetId<AssetId, Balance, AsPrefixedGeneralIndex<Local, AssetId, JustTry>, JustTry>,
```
KiChjang, line 508: Does this have to be in a tuple?

polkadot-parachains/statemine/src/lib.rs
```diff
@@ -498,6 +500,24 @@ pub type LocalAssetTransactor = CurrencyAdapter<
...
+	),
+	// Do a simple punn to convert an AccountId32 MultiLocation into a native chain account ID:
```
KiChjang, line 510: A simple "punn"?

polkadot-parachains/statemine/src/lib.rs
```diff
@@ -498,6 +500,24 @@ pub type LocalAssetTransactor = CurrencyAdapter<
...
+	LocationToAccountId,
+	// Our chain's account ID type (we can't get away without mentioning it explicitly):
+	AccountId,
+	// We only allow teleports of known assets.
+	NonZeroIssuance<AccountId, Assets>,
```
KiChjang, line 515: This doesn't seem related to teleportation? It's checking whether the assets in the account has a non-zero issuance.

KiChjang (commented):
3 comments, but they all apply to the statemint and westmint XCM configs.

On another note, the syntax for the XCM config is arcane, I wonder if we can use some macro magic to solve it...?

polkadot-parachains/statemine/src/lib.rs
apopiak: nope, good point

polkadot-parachains/statemine/src/lib.rs
apopiak: "unwrapping" the multilocation to get at the inner accountid, I'll remove it, it's copy-pasta

polkadot-parachains/statemine/src/lib.rs
apopiak: It is. The `CheckAsset` generic is used to determine whether we track the issuance of a teleport (e.g. [here](https://github.com/paritytech/polkadot/blob/616667f14a165951d56d574c6ea77b22dd4ed5ec/xcm/xcm-builder/src/fungibles_adapter.rs#L166)). I use non zero total issuance as a proxy here. I'll adjust the comment correspondingly.

KiChjang: Ah okay, you're trying to make the condition in `can_check_in` fail.

apopiak: For unknown assets, yes.

polkadot-parachains/westmint/src/lib.rs
```diff
@@ -435,20 +437,44 @@ pub type LocationToAccountId = (
...
+	>,
+	// Convert an XCM MultiLocation into a local account id:
+	LocationToAccountId,
+	// Our chain's account ID type (we can't get away without mentioning it explicitly):
+	AccountId,
+	// We only want to allow teleports of known assets. We use non-zero issuance as an indication
+	// that this asset is known.
+	NonZeroIssuance<AccountId, Assets>,
+	// The account to use for tracking teleports.
+	CheckingAccount,
```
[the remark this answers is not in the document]
apopiak: Yep, but if somehow teleports were happening (which would be misconfiguration), we would definitely want to track them, no?

shawntabrizi (approved):
this looks okay to me.

I assume we have tested this locally and all that?

Also an end to end test on westend when that is ready would be good

KiChjang (approved):

apopiak:
> this looks okay to me.
> 
> I assume we have tested this locally and all that?
> 
> Also an end to end test on westend when that is ready would be good

As mentioned it was tested via XCM sim. I've had issues getting it to test locally with Polkadot js apps due to types issues (with XCM) and was hoping an end-to-end test on Westend would be enough.
We also need to set up some integration tests in general.

apopiak:
Tested this sufficiently to merge and then experiment more with a deployment on Westend.

PR #562, opened 2021-08-05
Publish rustdocs to github pages

JoshOrndorff:
This PR adds a CI job that builds and published the rustdocs to github pages. Similar to crates.parity.io, https://paritytech.github.io/frontier/rustdocs/pallet_evm/, or https://purestake.github.io/moonbeam/moonbeam_runtime/index.html

This is useful when working on downstream code to have an API guide quickly available, especially since github will host it for free.

This will need a little bit of help from someone with some permissions on this repo to setup the publishing source as the `gh-pages` branch. [docs](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -12,20 +12,13 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v1
+        uses: actions/actions/checkout@v2.3.4
 
-      # With rustup's nice new toml format, we just need to run rustup show to install the toolchain
-      # https://github.com/actions-rs/toolchain/issues/126#issuecomment-782989659
-      - name: Setup Rust toolchain
+      - name: Rust show versions
         run: rustup show
 
-      - uses: actions/cache@v2
-        with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          key: ${{ runner.os }}-cargo-doc-${{ hashFiles('**/Cargo.lock') }}
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v1
 
       - name: Build rustdocs
         uses: actions-rs/cargo@v1
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,46 @@
+name: Publish Rust Docs
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  deploy-docs:
+    name: Deploy docs
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      # With rustup's nice new toml format, we just need to run rustup show to install the toolchain
+      # https://github.com/actions-rs/toolchain/issues/126#issuecomment-782989659
+      - name: Setup Rust toolchain
+        run: rustup show
+
+      - uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,46 @@
...
+      - master
+
+jobs:
+  deploy-docs:
+    name: Deploy docs
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
```
TriplEight, line 15:
```suggestion
        uses: actions/actions/checkout@v2.3.4
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Setup Rust toolchain
+        run: rustup show
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: ${{ runner.os }}-cargo-doc-${{ hashFiles('**/Cargo.lock') }}
```
TriplEight, line 28:
```suggestion
      - name: Rust cache
        uses: Swatinem/rust-cache@v1
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,46 @@
...
+    name: Deploy docs
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v1
+
+      # With rustup's nice new toml format, we just need to run rustup show to install the toolchain
+      # https://github.com/actions-rs/toolchain/issues/126#issuecomment-782989659
+      - name: Setup Rust toolchain
```
TriplEight, line 19:
```suggestion
      - name: Rust show versions
```

TriplEight (changes requested):

TriplEight (approved):

JoshOrndorff:
@TriplEight Have you adjusted the repo settings so that the published rustdocs will actually be available at paritytech.github.io/cumulus ? Or if not, could you?

TriplEight:
Yeah, I did. Not sure though where is it supposed to host the docs from, for now it's set to `/root`.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,39 @@
...
+      # Copied from https://github.com/substrate-developer-hub/rustdocs/blob/master/index.html
+      - name: Make index.html
+        run: echo "<meta http-equiv=refresh content=0;url=cumulus_client_collator/index.html>" > ./target/doc/index.html
+
+      - name: Deploy documentation
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_branch: gh-pages
+          publish_dir: ./target/doc
```
TriplEight, line 39: I'm not sure if `./target/doc` is correct, but let's try. GitHub has no other choice but `./root/ `or `./docs` and the Action's doc (which should be compatible) [says](https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-source-directory-publish_dir) it should be `./public`.

JoshOrndorff: This is the directory where `cargo doc` builds the docs to. `./public` is an example.

PR #615, opened 2021-09-16
ci: add quick-check with rustfmt

chevdor:
This PR brings a new **Quick Check** Github Workflow that will:
- trigger on any push to branches
- trigger for PRs
- check if the formatting is conform (`.rustfmt.toml`provided)

fix #614

PR #616, opened 2021-09-16
Add fmt and clippy

alvicsam:
Can I suggest using gitlab-ci instead of GHA?

Related to:
https://github.com/paritytech/cumulus/issues/614
https://github.com/paritytech/cumulus/pull/615

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -72,13 +72,13 @@ variables:                         &default-vars
 
 fmt:
   stage:                           lint
-  <<:                              *docker-env
+  <<:                              *kubernetes-env
   script:
     - cargo +nightly fmt --all -- --check
 
 clippy:
   stage:                           lint
-  <<:                              *docker-env
+  <<:                              *kubernetes-env
   script:
     # add clippy is temporary until tests in substrate are not fixed
     # https://github.com/paritytech/substrate/pull/9637
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -80,6 +80,9 @@ clippy:
   stage:                           lint
   <<:                              *docker-env
   script:
+    # add clippy is temporary until tests in substrate are not fixed
+    # https://github.com/paritytech/substrate/pull/9637
+    - rustup component add clippy
     - cargo clippy -- -D warnings
 
 #### stage:                        test
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -79,8 +79,7 @@ fmt:
 
 clippy:
   stage:                           lint
-  image:                           "${CI_IMAGE}"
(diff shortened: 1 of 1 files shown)


chevdor:
What would be the benefit in this case ?

alvicsam:
> What would be the benefit in this case ?

Almost all our projects use gitlab-ci and in future we want to have unified pipeline for all projects (at least the repeatable steps). So it's better to have one unified instrument for CI.

alvicsam:
Found out that `paritytech/ci-linux` doesn't have not fmt, nor clippy >_<
But I'll fix it.

chevdor:
For now, I am mainly working on fixing the issues related to those checks. 

Whether we do the checks in github or gitlab can be discussed. I would prioritize simplicity and speed, this is why I went for GHA since this is supposed to be a **quick** check and not a big mechanics. 

Ultimately, I don't mind but I find GHA more convenient for those, especially because we could have one Github Action that we use adn reused in most repos.

For big and heavy runs, I am with you on the choice of gitlab though.

NOTE: In your pipeline, you should run `fmt` first as it runs much faster and thus can fail much faster.

alvicsam:
`fmt` took 2 minutes, so it's relatively quick :) And it's just 5 lines in `.giltab-ci.yml`

>NOTE: In your pipeline, you should run fmt first as it runs much faster and thus can fail much faster.

Currently they are in one stage, so they run simultaneously. 

p.s. still need to add clippy to ci image

.gitlab-ci.yml
```diff
@@ -67,6 +68,23 @@ variables:                         &default-vars
     paths:
       - ./artifacts/
 
+#### stage:                        lint
+
+fmt:
+  stage:                           lint
+  <<:                              *docker-env
```
rcny, line 75: `&docker-env` refers to `linux-docker` runners - we use them for heavyweight jobs (e.g. compilation stuff). These runners have limited capacity. Please resort to `&kubernetes-env` (`kubernetes-parity-build`) instead.

.gitlab-ci.yml
```diff
@@ -67,6 +68,23 @@ variables:                         &default-vars
...
+  script:
+    - cargo +nightly fmt --all -- --check
+
+clippy:
+  stage:                           lint
+  <<:                              *docker-env
```
rcny, line 81: Same as above.

rcny (changes requested):

rcny:
`&kubernetes-env` lacks `image: "${CI_IMAGE}"` directive.

rcny (approved):

bkchr:
Already done. Ty anyway. 

I know that this runs stuff on gitlab,but I don't see any advantage in this anyway, so I'm going to close this pr.

Issue #588, opened 2021-08-25
Problem with the benchmark of pallet-collator-selection


andor0:
Tried to run the bechmark of pallet-collator-selection and got errors:
```
$ ./scripts/benchmarks.sh 
2021-08-25 09:35:21 assembling new collators for new session 0 at #0    
2021-08-25 09:35:21 assembling new collators for new session 1 at #0    
2021-08-25 09:35:22 panicked at 'called `Result::unwrap()` on an `Err` value: <wasm:stripped>', /home/ubuntu/cumulus/pallets/collator-selection/src/benchmarking.rs:59:86    
[...]

PR #639, opened 2021-10-03
Fix Benchmarks for Statemine-V4 release

NachoPal:
Related to: #627 

Issue was introduced in: #498 

Benchmarks stopped working because of this: https://github.com/paritytech/cumulus/pull/498/files#diff-a0359a0add814d66aac74493f7e17aefdcee21d3de1ae592e97059082b6feedfR336-R337 and the introduction of `MinCandidates`.

I had to fix `register_as_candidate`, `leave_intent` and `new_session` benchmarks.

In addition, I updated the `benchmarks.sh` script.

It fixes: #588

### DIFF
--- a/pallets/collator-selection/src/benchmarking.rs
+++ b/pallets/collator-selection/src/benchmarking.rs
@@ -20,14 +20,17 @@ use super::*;
 #[allow(unused)]
 use crate::Pallet as CollatorSelection;
 use sp_std::prelude::*;
+use sp_runtime::traits::Convert;
 use frame_benchmarking::{benchmarks, impl_benchmark_test_suite, whitelisted_caller, account};
 use frame_system::{RawOrigin, EventRecord};
 use frame_support::{
 	assert_ok,
+	assert_noop,
 	traits::{Currency, Get, EnsureOrigin},
+	codec::Decode
 };
 use pallet_authorship::EventHandler;
-use pallet_session::SessionManager;
+use pallet_session::{SessionManager, Pallet as Session};
 
 pub type BalanceOf<T> =
 	<<T as Config>::Currency as Currency<<T as frame_system::Config>::AccountId>>::Balance;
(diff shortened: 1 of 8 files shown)


pallets/collator-selection/src/benchmarking.rs
```diff
@@ -51,17 +52,60 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::Event) {
...
+	(create_funded_user::<T>("candidate", c, 1000), keys::<T>(c))
+}
+
+fn register_validators<T: Config + session::Config>(count: u32) {
+	let validators = (0..count).map(|c| validator::<T>(c)).collect::<Vec<_>>();
+
+	for (who, keys) in validators {
+			<session::Module<T>>::set_keys(
+			RawOrigin::Signed(who).into(), keys, vec![]
+		).unwrap();
```
apopiak, line 93: nitpick: formatting

pallets/collator-selection/src/benchmarking.rs
```diff
@@ -107,22 +151,32 @@ benchmarks! {
...
+
 	}: _(RawOrigin::Signed(caller.clone()))
 	verify {
 		assert_last_event::<T>(Event::CandidateAdded(caller, bond / 2u32.into()).into());
 	}
 
 	// worse case is the last candidate leaving.
 	leave_intent {
-		let c in 1 .. T::MaxCandidates::get();
+		let c in (T::MinCandidates::get() + 1) .. T::MaxCandidates::get();
```
apopiak, line 175:
```suggestion
		let c in T::MinCandidates::get().max(1) .. T::MaxCandidates::get();
```
Could it be that you want to address the possibility of `MinCandidates` being 0?

apopiak (approved):

pallets/collator-selection/src/benchmarking.rs
NachoPal: No, what I want is to make sure that there is at least an extra candidate which is able to leave. If the number of candidates is not bigger than `MinCandidates` it will panic.

PR #648, opened 2021-10-07
Version bump

### LINKED ISSUE #646: bump polkadot-collator version on release
We should bump the polkadot-collator on release.

chevdor:
fix #646


bkchr:
> I think I would've just bumped the binary?

Yes. No reason to bump everything

PR #669, opened 2021-10-15
disable permissionless asset creation for Statemint

apopiak:
We want to launch Statemint in stages similar to Statemine.
We will thus limit the extrinsics for permissionless asset creation on launch and only enable them once we are ready.


Swader:
What is the reasoning behind this change?

apopiak:
> What is the reasoning behind this change?

See, updated description: We want to launch slowly in stages. Permissionless asset creation will be enabled eventually.

PR #671, opened 2021-10-16
Make sure comments and code for local origins match

gavofyork:
- On Rococo and Westmint, local origns are allowed to send/execute XCM.
- In the template, Statemine and Statemint, local origin sending/execution is disabled.


KiChjang (approved):
Everything looks good, but should we start leaving 🔒 symbols as comments on code that needs CTO/CEO approval?

gavofyork:
Yes, though probably worth doing in a separate PR.

PR #673, opened 2021-10-17
Update weights from v4 to v5

NachoPal:
Merge needed before #672 to be able to compare new weights for V5 with old ones in V4


polkadot-parachains/statemine/src/weights/pallet_collator_selection.rs
```diff
@@ -26,46 +45,66 @@ use sp_std::marker::PhantomData;
...
+		(67_925_000 as Weight)
 			.saturating_add(T::DbWeight::get().reads(3 as Weight))
 			.saturating_add(T::DbWeight::get().writes(4 as Weight))
 	}
-	fn new_session(r: u32, c: u32) -> Weight {
+	// Storage: CollatorSelection Candidates (r:1 w:1)
+	// Storage: CollatorSelection LastAuthoredBlock (r:1000 w:1)
+	// Storage: System Account (r:1 w:1)
+	// Storage: CollatorSelection Invulnerables (r:1 w:0)
+	// Storage: System BlockWeight (r:1 w:1)
```
apopiak, line 99: It's a bit odd that System BlockWeight is not whitelisted for the benchmarks, but not sure it's worth it to dive deep into this :shrug:

PR #678, opened 2021-10-19
Companion for substrate#9732

cheme:
companion for https://github.com/paritytech/substrate/pull/9732


client/service/src/genesis.rs
```diff
@@ -21,21 +21,26 @@ use sp_runtime::traits::{Block as BlockT, Hash as HashT, Header as HeaderT, Zero
 /// Generate the genesis block from a given ChainSpec.
 pub fn generate_genesis_block<Block: BlockT>(
 	chain_spec: &Box<dyn ChainSpec>,
+	genesis_state_version: sp_runtime::StateVersion,
```
bkchr, line 24: Can we not fetch the runtime version from the code in the genesis state? As we do this in Substrate?

polkadot-parachains/seedling/src/lib.rs
```diff
@@ -60,6 +60,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
+	state_version: 1,
```
bkchr, line 63:
```suggestion
	state_version: 0,
```

Probably too late.

client/service/src/genesis.rs
cheme: That was my first idea, but current code need an executor (for compatibility I think), so ends up being really easier to use this way.
You think it is a must have, or can we keep it this way (seems only use from commands at this point)?

bkchr: Hmm, this is not really future proof.

bkchr (approved):

cheme:
bot merge

Issue #683, opened 2021-10-20
Template to have *safe* defaults to use in production networks


nuke-web3:
https://github.com/paritytech/cumulus/blob/2ff5f7c443e046dc983615c0dd11053d6f630a59/parachain-template/runtime/src/lib.rs#L446-L448

See also the review comments: https://github.com/substrate-developer-hub/substrate-parachain-template/pull/80

And key motivation: https://medium.com/kusama-network/kusamas-governance-thwarts-would-be-attacker-9023180f6fb

What else needs to be modified so that if the template is used on a value bearing network as a sovereign chain, it would by as safe as possible from misconfiguration? 

Should also add in-line comment the runtime with respect to the most important and easy to misconfigure things. And a section on the devhub to discuss them

PR #690, opened 2021-10-20
Use safer template defaults for `polkadot-v0.9.12`, move to `recognize`, comment on Override for AdvertisedXcmVersion default

nuke-web3:
Based on discussion & review here: https://github.com/substrate-developer-hub/substrate-parachain-template/pull/80

A start (but not a full fix) for #683 targeting `polkadot-v0.9.12` release, should also be used in `master` if merged here. 
Please commit directly anything that you (Cumulus core team) feels should be added/amended :pray:

### DIFF
--- a/parachain-template/runtime/src/lib.rs
+++ b/parachain-template/runtime/src/lib.rs
@@ -518,7 +518,7 @@ impl pallet_xcm::Config for Runtime {
 	type Origin = Origin;
 	type Call = Call;
 
-	const VERSION_DISCOVERY_QUEUE_SIZE: u32 = 100; 
+	const VERSION_DISCOVERY_QUEUE_SIZE: u32 = 100;
 	// Override for AdvertisedXcmVersion default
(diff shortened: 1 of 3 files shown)


parachain-template/runtime/src/lib.rs
```diff
@@ -516,7 +512,7 @@ impl pallet_xcm::Config for Runtime {
 	type XcmExecuteFilter = Everything;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
 	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Everything;
+	type XcmReserveTransferFilter = ();
```
bkchr, line 515:
```suggestion
	type XcmReserveTransferFilter = Nothing;
```

Wasn't there a `Nothing`?

bkchr (approved):

NukeManDan:
https://writingexplained.org/recognise-or-recognize-difference I think the `z` is now common enough? :sweat_smile:

PR #715, opened 2021-10-31
Companion for `dispatch_as`

shawntabrizi:
companion for https://github.com/paritytech/substrate/pull/9934

polkadot companion: paritytech/polkadot#4075

### DIFF
--- a/polkadot-parachains/statemine/src/weights/pallet_utility.rs
+++ b/polkadot-parachains/statemine/src/weights/pallet_utility.rs
@@ -39,4 +39,7 @@ impl<T: frame_system::Config> pallet_utility::WeightInfo for WeightInfo<T> {
 			// Standard Error: 0
 			.saturating_add((4_578_000 as Weight).saturating_mul(c as Weight))
 	}
+	fn dispatch_as() -> Weight {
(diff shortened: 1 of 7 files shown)


polkadot-parachains/statemine/src/weights/pallet_utility.rs
```diff
@@ -39,4 +39,7 @@ impl<T: frame_system::Config> pallet_utility::WeightInfo for WeightInfo<T> {
 			// Standard Error: 0
 			.saturating_add((4_578_000 as Weight).saturating_mul(c as Weight))
 	}
+	fn dispatch_as() -> Weight {
+		(14_340_000 as Weight)
```
thiolliere, line 43: should we run benchmark on the PR ?
Anyhow I guess we'll run them before release

thiolliere (approved):

polkadot-parachains/statemine/src/weights/pallet_utility.rs
shawntabrizi: also this kind of benchmark is unaffected by which runtime it is inside

shawntabrizi:
bot merge

PR #724, opened 2021-11-07
Add councillors

brenzi:
Adding keys from 5 Encointer association members for genesis council. 
Prime will be elected once launched


brenzi:
wrong repo

Issue #681, opened 2021-10-19
Remove RandomnessCollectiveFlip


apopiak:
RandomnessCollectiveFlip was removed from Polkadot (https://github.com/paritytech/polkadot/pull/3347) because it is not secure. We should remove it from Statemint (etc.) as well.

PR #726, opened 2021-11-08
Remove randomness collective flip new

NachoPal:
Fixes: #681


polkadot-parachains/statemint/src/lib.rs
```diff
@@ -789,8 +786,19 @@ pub type Executive = frame_executive::Executive<
...
 
+pub struct RemoveCollectiveFlip;
+impl frame_support::traits::OnRuntimeUpgrade for RemoveCollectiveFlip {
+	fn on_runtime_upgrade() -> Weight {
+		use frame_support::storage::migration;
+		// Remove the storage value `RandomMaterial` from removed pallet `RandomnessCollectiveFlip`
+		migration::remove_storage_prefix(b"RandomnessCollectiveFlip", b"RandomMaterial", b"");
+		<Runtime as frame_system::Config>::DbWeight::get().writes(1)
+	}
+}
```
apopiak, line 800: I don't think we need this as Statemint is not yet live

polkadot-parachains/westmint/src/lib.rs
```diff
@@ -702,7 +700,6 @@ construct_runtime!(
 		ParachainSystem: cumulus_pallet_parachain_system::{
 			Pallet, Call, Config, Storage, Inherent, Event<T>, ValidateUnsigned,
 		},
-		RandomnessCollectiveFlip: pallet_randomness_collective_flip::{Pallet, Storage},
```
apopiak, line 705: we cannot just remove this, it would change the indices. You'll need to add an explicit index at least for the timestamp pallet directly after.

apopiak: Talked with Joe about it and we want to align the indices between Westmint and Statemint (https://github.com/paritytech/cumulus/issues/729) so this is less relevant I guess. We'll just need to bump the tx_version at some point

NachoPal: Can I align them in this PR?

apopiak: I'm opening a separate PR

NachoPal: If you bump the tx_version in #731 I think this one can be merged

NachoPal: Or I bump it here?

apopiak (approved):

apopiak:
Needs an update to the `Cargo.lock`

PR #731, opened 2021-11-09
Align Call Indices between Westmint and Statemint

### LINKED ISSUE #729: Make Westmint indices explicit
we want to align construct_runtime indices between westmint and statemint

apopiak:
We would like the Call indices to be the same between the different Statemint flavors.
This PR adjusts Westmint to be in line with Statemint and Statemine (except for adding `pallet-sudo` as index 255).

merge after #726 
closes #729


NachoPal (approved):

NachoPal:
Can merge now

PR #740, opened 2021-11-10
polkadot collator builder

TriplEight:
- add `polkadot-collator_builder.Containerfile` to simplify/unify the build process for the users and docs on how to use it
- not quite sure if you use `docker/test-parachain-collator.dockerfile` but I've optimized it a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,13 +41,13 @@ Statemint is a common good parachain providing an asset store for the Polkadot e
 To run a Statemine or Westmint node (Statemint is not deployed, yet) you will need to compile the
 `polkadot-collator` binary:
 
-```sh
+```bash
 cargo build --release --locked -p polkadot-collator
 ```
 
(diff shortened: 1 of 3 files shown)


alvicsam:
Why `polkadot-collator_builder.Containerfile` not `polkadot-collator_builder.Dockerfile`?

TriplEight:
The syntax is the same, `docker` consumes both extensions likewise, as well as `podman` and `buildah`. We use the latter ones. Just want to quit this `Xerox` thing, nothing more than that.

alvicsam (approved):

alvicsam:
> The syntax is the same, `docker` consumes both extensions likewise, as well as `podman` and `buildah`. We use the latter ones. Just want to quit this `Xerox` thing, nothing more than that.

I think it worth adding to the container naming policy

bkchr (approved):

PR #743, opened 2021-11-10
Add safe_xcm_version to Genesis Specs

apopiak:
We want to configure `pallet-xcm`'s `safe_xcm_version` on genesis.
This PR adds it to the chain spec.


polkadot-parachains/src/chain_spec.rs
```diff
@@ -30,6 +30,9 @@ pub type ChainSpec =
 /// Specialized `ChainSpec` for the shell parachain runtime.
 pub type ShellChainSpec = sc_service::GenericChainSpec<shell_runtime::GenesisConfig, Extensions>;
 
+/// The default XCM version to set in genesis config.
+const SAFE_XCM_VERSION: u32 = 2;
```
KiChjang, line 34: Do you want to instead set it to the latest version of the XCM pallet whenever it gets updated, i.e. `pallet_xcm::XCM_VERSION`?

apopiak: I don't think so? This is supposed to be the fallback version, right? Version negotiation is meant to go on top of this.
And the docs in the pallet say:
```
/// An implementation of `Get<u32>` which just returns the latest XCM version which we can support.
```

apopiak: https://github.com/paritytech/polkadot/blob/55872676fae63399eb111a60ba27cb735cbaead6/xcm/pallet-xcm/src/lib.rs#L66-L70

KiChjang: Yeah, fallback and default, just in case the destination is yet to be discovered but an XCM version still needs to be selected in order to be able to be sent over the wire. I guess it makes sense to set it to be as low as possible in order for everything to work flawlessly.

KiChjang (approved):

chevdor (approved):

PR #748, opened 2021-11-11
bump transaction_version due to parachain_system change

apopiak:
https://github.com/paritytech/cumulus/pull/517 changed the calls of the  `parachain_system` pallet.
--> Bump transaction version.


chevdor:
Westmint could use a bump as well ([details](https://github.com/paritytech/cumulus/issues/744#issuecomment-966431768))

apopiak:
> Westmint could use a bump as well ([details](https://github.com/paritytech/cumulus/issues/744#issuecomment-966431768))

It is already bumped to 4. I was aware that we're reshuffling all the pallets :-D

chevdor:
What about rococo ?

apopiak:
It's not live anywhere, is it?
We don't really have to bump spec or transaction versions for chains that are not live

chevdor (approved):
LGTM

PR #759, opened 2021-11-14
move to rust 2021

NukeManDan:
closes #758 
Confirmed to build locally on Ubuntu 20.04.3 LTS:
```
rustc 1.56.1 (59eed8a2a 2021-11-01)
rustc 1.58.0-nightly (e90c5fbbc 2021-11-12)
```

PR #760, opened 2021-11-15
Specify edition 2021 in .rustfmt.yml

liuchengxu:
This PR is mainly to fix the error `[E0670]: async fn is not permitted in the 2015 edition` when
using Vim along with rust-analyzer, also removes an unused dependency from parachain-template.

Ref: https://github.com/rust-analyzer/rust-analyzer/issues/1959

Can be also merged into #759 if you prefer.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,4 +1,5 @@
 # Basic
+edition = "2021"
 hard_tabs = true
 max_width = 100
 use_small_heuristics = "Max"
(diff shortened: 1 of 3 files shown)


chevdor:
[srtool build](https://github.com/paritytech/srtool/runs/4209420301?check_suite_focus=true) ✅

chevdor (approved):

chevdor:
This should be merged after #759

bkchr (approved):

PR #771, opened 2021-11-19
Mock XCM in parachain inherent

JoshOrndorff:
This PR extends the `MockValidationDataInherentDataProvider` to support injecting downward and horizontal XCM messages into the runtime. This allows development nodes that use this mock inherent data provider to test their XCM functionality without the overhead of a backing relay chain.

For an example of this code in action, see https://github.com/PureStake/moonbeam/pull/916 where we wire it up to an RPC endpoint called `manual-xcm-rpc` that works similar to manual seal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1796,6 +1796,7 @@ dependencies = [
  "sp-runtime",
  "sp-state-machine",
  "sp-std",
+ "sp-storage",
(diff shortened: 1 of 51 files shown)


JoshOrndorff:
Oops, wrong branch. Let me try again.

PR #782, opened 2021-11-22
Add check_labels script and workflow

chevdor:
This PR adds the same `check_labels` check for PRs.

### DIFF
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -1,0 +1,23 @@
+name: Check labels
+
+on:
+  pull_request:
+    types: [labeled, opened, synchronize, unlabeled]
+
+jobs:
+  check-labels:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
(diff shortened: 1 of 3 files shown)


TriplEight (changes requested):
I suggest it's better to avoid doubling the maintenance and move these files to some common source like https://github.com/paritytech/pipeline-scripts

chevdor:
Yes, that would be ideal. I would love however to have something in place asap because right now, we have nothing in place. Would that be ok to merge this version and think about an upgrade ?
Also keep in mind that that repos are different so the rules may be(come) different.

I have ideas to improve the topic but I would like to first have *something* in place.

TriplEight:
@chevdor I'm OK with merging this, but please create a follow-up issue, mentioning where the instances of the script should be substituted with an importable version. And certainly with your ideas, too! I think the right place for the issue should be here https://github.com/paritytech/opstooling/issues/

TriplEight (approved):

PR #794, opened 2021-11-24
Configure Statemint and Statemine to Allow Teleports

apopiak:
This PR adjust the config for pallet-xcm for Statemint and Statemine to allow teleports (and reserve transfers), but disallow generic sends.

Also bumps the `spec_version`.


apopiak:
Feel free to review and check, already in the v6.0.1 release branch, though.

PR #799, opened 2021-11-25
Support overweight messages in XCMP queue

KiChjang:
Implementation follows closely to the UMP overweight message handling, as seen [here in this PR](https://github.com/paritytech/polkadot/pull/3575).

Fixes paritytech/srlabs_findings#145.

### DIFF
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -25,6 +25,8 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+pub mod migration;
+
 #[cfg(test)]
 mod mock;
 
@@ -51,6 +53,8 @@ pub use pallet::*;
 /// Index used to identify overweight XCMs.
 pub type OverweightIndex = u64;
 
+const LOG_TARGET: &str = "xcmp_queue";
+
 #[frame_support::pallet]
 pub mod pallet {
 	use super::*;
@@ -59,6 +63,7 @@ pub mod pallet {
 
 	#[pallet::pallet]
 	#[pallet::generate_store(pub(super) trait Store)]
+	#[pallet::storage_version(migration::STORAGE_VERSION)]
 	pub struct Pallet<T>(_);
 
 	#[pallet::config]
@@ -80,6 +85,10 @@ pub mod pallet {
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_runtime_upgrade() -> Weight {
+			migration::migrate_to_latest::<T>()
+		}
+
 		fn on_idle(_now: T::BlockNumber, max_weight: Weight) -> Weight {
 			// on_idle processes additional messages with any remaining block weight.
 			Self::service_xcmp_queue(max_weight)
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -36,7 +36,7 @@ use cumulus_primitives_core::{
 	relay_chain::BlockNumber as RelayBlockNumber, ChannelStatus, GetChannelInfo, MessageSendError,
 	ParaId, XcmpMessageFormat, XcmpMessageHandler, XcmpMessageSource,
 };
-use frame_support::weights::Weight;
+use frame_support::weights::{constants::WEIGHT_PER_MILLIS, Weight};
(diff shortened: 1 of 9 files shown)


pallets/xcmp-queue/src/lib.rs
```diff
@@ -390,6 +455,20 @@ impl<T: Config> Pallet<T> {
 								remaining_fragments = last_remaining_fragments;
 								break
 							},
+							Err(XcmError::WeightLimitReached(required))
+								if required > max_individual_weight =>
```
apopiak, line 459: AFAICT This is only ever triggered if `required > max_weight`, thus `max_individual_weight` is ignored unless we run out of weight for processing XCMs.
The UMP has the two conditions switched: the `required > max_individual_weight` is checked first: 
https://github.com/paritytech/polkadot/pull/3575/files#diff-2893612305e6b9c3b5c48fe27288f735b343e5ce4cca69459fe89f75e282d10bR489

pallets/xcmp-queue/src/lib.rs
```diff
@@ -196,6 +244,22 @@ pub struct QueueConfigData {
 	/// The speed to which the available weight approaches the maximum weight. A lower number
 	/// results in a faster progression. A value of 1 makes the entire weight available initially.
 	weight_restrict_decay: Weight,
+	/// The maximum amount of weight any individual message may consume. Messages above this weight
+	/// go into the overweight queue and may only be serviced explicitly.
+	xcmp_max_individual_weight: Weight,
```
apopiak, line 249: `QueueConfigData` is held in storage --> you will need to introduce a storage migration.

apopiak (commented):
I would like to see a test that tries to process an overweight XCM and then calls the service_overweight functionality.

pallets/xcmp-queue/src/migration.rs
```diff
@@ -0,0 +1,128 @@
...
+		super::QueueConfigData {
+			suspend_threshold: pre.suspend_threshold,
+			drop_threshold: pre.drop_threshold,
+			resume_threshold: pre.resume_threshold,
+			threshold_weight: pre.threshold_weight,
+			weight_restrict_decay: pre.weight_restrict_decay,
+			xcmp_max_individual_weight: super::QueueConfigData::default()
+				.xcmp_max_individual_weight,
+		}
+	};
```
apopiak, line 82: Would it make sense to parameterize the migration function to take a `xcmp_max_individual_weight` so that people can migrate to whatever config value they want?
Alternatively we'd want to expose some knob somewhere to update the queue config, no?
Maybe a candidate for a follow-up issue?

apopiak (approved):
LGTM, I'll let Gav address the question in code

pallets/xcmp-queue/src/migration.rs
apopiak: added an issue: https://github.com/paritytech/cumulus/issues/801

pallets/xcmp-queue/src/lib.rs
```diff
@@ -339,7 +412,7 @@ impl<T: Config> Pallet<T> {
 
 	fn handle_xcm_message(
 		sender: ParaId,
-		_sent_at: RelayBlockNumber,
+		_sent_at: RelayBlockNumber, // Review Q: why is this arg still here?
```
KiChjang, line 415: This here is important because if we can remove this arg, then I can store one less thing in storage, potentially saving quite a bit of storage space.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -430,6 +504,20 @@ impl<T: Config> Pallet<T> {
...
+								// overweight - add to overweight queue and continue with message
+								// execution consuming the message.
+								let msg_len = last_remaining_fragments
+									.len()
+									.saturating_sub(remaining_fragments.len());
+								let overweight_xcm = last_remaining_fragments[..msg_len].to_vec();
+								let index = Self::stash_overweight(sender, sent_at, overweight_xcm);
+								Self::deposit_event(Event::OverweightEnqueued(
+									sender, sent_at, index, required,
+								));
```
gavofyork, line 519:
```suggestion
								let e = Event::OverweightEnqueued(sender, sent_at, index, required);
								Self::deposit_event(e);
```

gavofyork (approved):

gavofyork:
Looks reasonable.

shawntabrizi (approved):

PR #806, opened 2021-11-29
support asset transfer from/to sibling chain

yrong:
This PR adds the `AsPrefixedGeneralIndexFromLocalOrRemote` conversion type to allow accepting assets that were moved to another chain via reserve based transfers back onto Statemint (and friends).

### DIFF
--- a/polkadot-parachains/parachains-common/src/impls.rs
+++ b/polkadot-parachains/parachains-common/src/impls.rs
@@ -22,9 +22,10 @@ use frame_support::traits::{
 };
 use pallet_asset_tx_payment::HandleCredit;
 use sp_runtime::traits::Zero;
-use sp_std::marker::PhantomData;
-use xcm::latest::{AssetId, Fungibility::Fungible, MultiAsset, MultiLocation};
-use xcm_executor::traits::FilterAssetLocation;
+use sp_std::{borrow::Borrow, marker::PhantomData, result};
+use xcm::latest::{AssetId, Fungibility::Fungible, Junction, MultiAsset, MultiLocation};
+
+use xcm_executor::traits::{Convert, FilterAssetLocation};
 
 /// Type alias to conveniently refer to the `Currency::NegativeImbalance` associated type.
 pub type NegativeImbalance<T> = <pallet_balances::Pallet<T> as Currency<
@@ -113,6 +114,44 @@ impl<T: Get<MultiLocation>> FilterAssetLocation for AssetsFrom<T> {
 	}
 }
 
+/// Converter struct with two main usage scenarios:
+/// 1. Transfer asset from local to other para-chain
+/// 	- with `reverse_ref` to convert a numeric asset ID (must be `TryFrom/TryInto<u128>`) into a `GeneralIndex` junction
+/// 2. Transfer asset back from other para-chain to local
+/// 	- with `convert_ref` to convert multilocation struct `(1,X2(ParaChain(para_id),GeneralIndex(general_index))` to a numeric asset ID
+pub struct AsPrefixedGeneralIndexFromLocalOrRemote<
+	LocalPrefix,
+	RemotePrefix,
+	AssetId,
+	ConvertAssetId,
+>(PhantomData<(LocalPrefix, RemotePrefix, AssetId, ConvertAssetId)>);
+impl<
+		LocalPrefix: Get<MultiLocation>,
+		RemotePrefix: Get<MultiLocation>,
+		AssetId: Clone,
+		ConvertAssetId: Convert<u128, AssetId>,
+	> Convert<MultiLocation, AssetId>
+	for AsPrefixedGeneralIndexFromLocalOrRemote<LocalPrefix, RemotePrefix, AssetId, ConvertAssetId>
+{
+	fn convert_ref(id: impl Borrow<MultiLocation>) -> result::Result<AssetId, ()> {
+		let id = id.borrow();
+
+		match id.match_and_split(&LocalPrefix::get()) {
+			Some(Junction::GeneralIndex(id)) => ConvertAssetId::convert_ref(id),
+			_ => match id.match_and_split(&RemotePrefix::get()) {
+				Some(Junction::GeneralIndex(id)) => ConvertAssetId::convert_ref(id),
+				_ => Err(()),
+			},
+		}
+	}
+	fn reverse_ref(what: impl Borrow<AssetId>) -> result::Result<MultiLocation, ()> {
+		let mut location = LocalPrefix::get();
+		let id = ConvertAssetId::reverse_ref(what)?;
+		location.push_interior(Junction::GeneralIndex(id)).map_err(|_| ())?;
+		Ok(location)
+	}
+}
+
 #[cfg(test)]
 mod tests {
 	use super::*;
@@ -130,7 +169,9 @@ mod tests {
 		traits::{BlakeTwo256, IdentityLookup},
 		Perbill,
 	};
+
 	use xcm::prelude::*;
+	use xcm_executor::traits::JustTry;
 
 	type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<Test>;
 	type Block = frame_system::mocking::MockBlock<Test>;
@@ -280,5 +321,60 @@ mod tests {
 			),
 			"AssetsFrom should allow assets from any of its interior locations"
 		);
+	}
+
+	#[test]
+	fn assets_from_convert_correctly() {
+		parameter_types! {
+			pub const Local: MultiLocation = Here.into();
+			pub Remote: MultiLocation = MultiLocation::new(1, X1(Parachain(1000)));
+		}
+		let local_asset_location = Local::get().pushed_with_interior(GeneralIndex(42)).unwrap();
+		let remote_asset_location = Remote::get().pushed_with_interior(GeneralIndex(42)).unwrap();
+		let mut asset_id =
+			AsPrefixedGeneralIndexFromLocalOrRemote::<Local, Remote, u32, JustTry>::convert_ref(
+				&local_asset_location,
+			)
+			.unwrap_or(u32::default());
+		assert_eq!(asset_id, 42);
+		asset_id =
+			AsPrefixedGeneralIndexFromLocalOrRemote::<Local, Remote, u32, JustTry>::convert_ref(
+				&remote_asset_location,
+			)
+			.unwrap_or(u32::default());
+		assert_eq!(asset_id, 42);
+	}
+
+	#[test]
+	fn assets_from_convert_error_with_different_parent() {
+		parameter_types! {
+			pub const Local: MultiLocation = Here.into();
+			pub Remote: MultiLocation = MultiLocation::new(2, Junctions::Here);
+		}
+		let remote_asset_location = MultiLocation::new(1, X1(GeneralIndex(42)));
+		let asset_id =
+			AsPrefixedGeneralIndexFromLocalOrRemote::<Local, Remote, u32, JustTry>::convert_ref(
+				&remote_asset_location,
+			)
+			.unwrap_or(u32::default());
+		assert_eq!(asset_id, u32::default())
+	}
+
+	#[test]
+	fn assets_from_convert_error_with_different_interiors() {
+		parameter_types! {
+			pub const Local: MultiLocation = Here.into();
+			pub Remote: MultiLocation = MultiLocation::new(1, X1(Parachain(1000)));
+		}
+		let mut remote_asset_location =
+			Remote::get().pushed_with_interior(PalletInstance(10)).unwrap();
+		remote_asset_location =
+			remote_asset_location.pushed_with_interior(GeneralIndex(42)).unwrap();
+		let asset_id =
+			AsPrefixedGeneralIndexFromLocalOrRemote::<Local, Remote, u32, JustTry>::convert_ref(
+				&remote_asset_location,
+			)
+			.unwrap_or(u32::default());
+		assert_eq!(asset_id, u32::default())
 	}
 }
--- a/polkadot-parachains/statemint/src/lib.rs
+++ b/polkadot-parachains/statemint/src/lib.rs
@@ -65,17 +65,18 @@ use parachains_common::{
 pub use sp_runtime::BuildStorage;
 
 // Polkadot imports
+use crate::common::impls::AsPrefixedGeneralIndexFromLocalOrRemote;
 use pallet_xcm::{EnsureXcm, IsMajorityOfBody, XcmPassthrough};
 use polkadot_parachain::primitives::Sibling;
 use polkadot_runtime_common::{BlockHashCount, RocksDbWeight, SlowAdjustingFeeUpdate};
 use xcm::latest::prelude::*;
 use xcm_builder::{
 	AccountId32Aliases, AllowKnownQueryResponses, AllowSubscriptionsFrom,
-	AllowTopLevelPaidExecutionFrom, AllowUnpaidExecutionFrom, AsPrefixedGeneralIndex,
-	ConvertedConcreteAssetId, CurrencyAdapter, EnsureXcmOrigin, FixedWeightBounds,
-	FungiblesAdapter, IsConcrete, LocationInverter, NativeAsset, ParentAsSuperuser,
-	ParentIsDefault, RelayChainAsNative, SiblingParachainAsNative, SiblingParachainConvertsVia,
-	SignedAccountId32AsNative, SovereignSignedViaLocation, TakeWeightCredit, UsingComponents,
+	AllowTopLevelPaidExecutionFrom, AllowUnpaidExecutionFrom, ConvertedConcreteAssetId,
+	CurrencyAdapter, EnsureXcmOrigin, FixedWeightBounds, FungiblesAdapter, IsConcrete,
+	LocationInverter, NativeAsset, ParentAsSuperuser, ParentIsDefault, RelayChainAsNative,
+	SiblingParachainAsNative, SiblingParachainConvertsVia, SignedAccountId32AsNative,
+	SovereignSignedViaLocation, TakeWeightCredit, UsingComponents,
 };
 use xcm_executor::{traits::JustTry, Config, XcmExecutor};
 
@@ -445,6 +446,7 @@ parameter_types! {
 	pub Ancestry: MultiLocation = Parachain(ParachainInfo::parachain_id().into()).into();
 	pub const Local: MultiLocation = Here.into();
 	pub CheckingAccount: AccountId = PolkadotXcm::check_account();
+	pub Remote: MultiLocation = MultiLocation::new(1, X1(Parachain(ParachainInfo::parachain_id().into())));
 }
 
 /// Type for specifying how a `MultiLocation` can be converted into an `AccountId`. This is used
@@ -481,7 +483,7 @@ pub type FungiblesTransactor = FungiblesAdapter<
 	ConvertedConcreteAssetId<
 		AssetId,
 		Balance,
-		AsPrefixedGeneralIndex<Local, AssetId, JustTry>,
+		AsPrefixedGeneralIndexFromLocalOrRemote<Local, Remote, AssetId, JustTry>,
 		JustTry,
 	>,
 	// Convert an XCM MultiLocation into a local account id:
--- a/polkadot-parachains/westmint/src/lib.rs
+++ b/polkadot-parachains/westmint/src/lib.rs
@@ -65,18 +65,18 @@ use parachains_common::{
 pub use sp_runtime::BuildStorage;
 
 // Polkadot imports
+use crate::common::impls::AsPrefixedGeneralIndexFromLocalOrRemote;
 use pallet_xcm::XcmPassthrough;
 use polkadot_parachain::primitives::Sibling;
 use polkadot_runtime_common::{BlockHashCount, RocksDbWeight, SlowAdjustingFeeUpdate};
 use xcm::latest::prelude::*;
 use xcm_builder::{
 	AccountId32Aliases, AllowKnownQueryResponses, AllowSubscriptionsFrom,
-	AllowTopLevelPaidExecutionFrom, AllowUnpaidExecutionFrom, AsPrefixedGeneralIndex,
-	ConvertedConcreteAssetId, CurrencyAdapter, EnsureXcmOrigin, FixedWeightBounds,
-	FungiblesAdapter, IsConcrete, LocationInverter, NativeAsset, ParentAsSuperuser,
-	ParentIsDefault, RelayChainAsNative, SiblingParachainAsNative, SiblingParachainConvertsVia,
-	SignedAccountId32AsNative, SignedToAccountId32, SovereignSignedViaLocation, TakeWeightCredit,
-	UsingComponents,
+	AllowTopLevelPaidExecutionFrom, AllowUnpaidExecutionFrom, ConvertedConcreteAssetId,
+	CurrencyAdapter, EnsureXcmOrigin, FixedWeightBounds, FungiblesAdapter, IsConcrete,
+	LocationInverter, NativeAsset, ParentAsSuperuser, ParentIsDefault, RelayChainAsNative,
+	SiblingParachainAsNative, SiblingParachainConvertsVia, SignedAccountId32AsNative,
+	SignedToAccountId32, SovereignSignedViaLocation, TakeWeightCredit, UsingComponents,
 };
 use xcm_executor::{traits::JustTry, Config, XcmExecutor};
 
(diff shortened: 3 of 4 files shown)


yrong:
@apopiak @Byron 

 not sure it is the proper way, AFAIK the current asset-multilocation adaptor/convert only allows to transfering statemine asset from local and this pr is for allow transfering from other parachain remote with location like `X2(parachain:1000,GeneralIndex:asset_id)`

KiChjang:
Thanks for your PR! Unfortunately, this is not the right place to add a new item to configure XCM. I believe the right place would be to add it in Polkadot's repository, under xcm/xcm-builder.

Aside from that, I'm not quite sure what this PR is trying to solve. Can you please elaborate the background of your problem and why your solution would solve it? I don't understand anything about this need for a `GeneralIndex` for example.

yrong:
@KiChjang The requirement is simple, we would like to transfer asset issued in statemine (e.g:usdt) to our chain and then transfer it back. With this pr and [some adaptor logic](https://github.com/bifrost-finance/bifrost/blob/c7eb1bef7abfdeb6562aeda29ab76805308b77f5/runtime/common/src/xcm_impl.rs#L153-L158) I can achive that. The `GeneralIndex` is to encode asset_id as multiLocation now adopted in statemine runtime.

KiChjang:
So if I'm hearing you correctly, you're saying that the `GeneralIndex` is used to specify which one of the many assets in statemine that you want to transfer to the destination chain via reserve transfers? In other words, your flow is exactly the same as `InititateReserveWithdraw`, but with the ability to specify an asset that is not KSM, correct?

If so, then there may need to be a bit more thought into how the MultiLocation of a reserve chain asset should be represented. Right now you have `Parachain(para_id)/GeneralIndex(index)`, which is rather imprecise -- a `GeneralIndex` can be interpreted in many different ways. To reduce the ambiguity, I propose inserting a `PalletIndex(assets_pallet_index)` in between, so that the whole path becomes `Parachain(para_id)/PalletIndex(assets_pallet_index)/GeneralIndex(index)`, so that there's even less of a chance to misinterpret what a `GeneralIndex` means.

yrong:
> So if I'm hearing you correctly, you're saying that the `GeneralIndex` is used to specify which one of the many assets in statemine that you want to transfer to the destination chain via reserve transfers? In other words, your flow is exactly the same as `InititateReserveWithdraw`, but with the ability to specify an asset that is not KSM, correct?
Exactly

> If so, then there may need to be a bit more thought into how the MultiLocation of a reserve chain asset should be represented. Right now you have `Parachain(para_id)/GeneralIndex(index)`, which is rather imprecise -- a `GeneralIndex` can be interpreted in many different ways. To reduce the ambiguity, I propose inserting a `PalletIndex(assets_pallet_index)` in between, so that the whole path becomes `Parachain(para_id)/PalletIndex(assets_pallet_index)/GeneralIndex(index)`, so that there's even less of a chance to misinterpret what a `GeneralIndex` means.

seems you want to add `assets_pallet_index' in between, I'am not sure fully understand your concern here. AFAIK statemine is a common good chain focus in asset management and `GeneralIndex` is the only place we use in statemine to reprensent asset. IMHO we may not need another level for pallet index here.

KiChjang:
@yrong If for whatever reason, `1000` is suddenly _not_ the parachain ID for statemine, due to it being reorganized or that you're in fact on a relay chain that has a different architecture (or maybe simply a typo), and `Parachain(1000)` also happens to interpret `GeneralIndex(index)` in a way that's different from what you'd expect, then you're in trouble because now you don't actually know what the XCM is doing.

apopiak:
When I set up the asset XCM integration I figured that we need a unique asset identifier within Statemint, so I used the `GeneralIndex`. When receiving an asset identified by a `GeneralIndex` Statemint will interpret it as a local asset that it has sovereignty/control over.
Now, I agree that we might want to allow the transaction of other assets in the future, so we will need to define how to identify them and then a `PalletIndex` or originating `Parachain` id might come in handy.

@yrong My question to you is: What keeps you from just sending a `GeneralIndex` as the asset's id? Left a corresponding comment here: https://github.com/bifrost-finance/bifrost/pull/420/files#r759127339

yrong:
> When I set up the asset XCM integration I figured that we need a unique asset identifier within Statemint, so I used the `GeneralIndex`. When receiving an asset identified by a `GeneralIndex` Statemint will interpret it as a local asset that it has sovereignty/control over. Now, I agree that we might want to allow the transaction of other assets in the future, so we will need to define how to identify them and then a `PalletIndex` or originating `Parachain` id might come in handy.
> 
> @yrong My question to you is: What keeps you from just sending a `GeneralIndex` as the asset's id? Left a corresponding comment here: https://github.com/bifrost-finance/bifrost/pull/420/files#r759127339

> Edit: Context is that the `AsPrefixedGeneralIndex` is already live in Statemine so I would want a good argument why it needs to be changed.

@apopiak some clarification here https://github.com/bifrost-finance/bifrost/pull/420#discussion_r759428910

apopiak:
Alright, seems like the reanchoring logic in the executor means we should merge (something like) this.

polkadot-parachains/westmint/src/lib.rs
```diff
@@ -456,6 +461,40 @@ pub type CurrencyTransactor = CurrencyAdapter<
...
+			},
+		}
+	}
+	fn reverse_ref(what: impl Borrow<AssetId>) -> result::Result<MultiLocation, ()> {
+		let mut location = LocalPrefix::get();
+		let id = ConvertAssetId::reverse_ref(what)?;
+		location.push_interior(Junction::GeneralIndex(id)).map_err(|_| ())?;
+		Ok(location)
+	}
+}
```
apopiak, line 496: Thank you for the contribution from my side as well, btw. :pray: 

Here are the adjustments I'd like to see:
+ [ ] Please move this into parachains-common. (Statemint and Statemine will have to use the same.)
+ [ ] This also should have a few tests:
  + matching the local prefix
  + matching the remote prefix
  + the failure cases
+ [ ] Please add docs
+ [ ] Also you skipped the prefix checking from the original here:
https://github.com/paritytech/polkadot/blob/55872676fae63399eb111a60ba27cb735cbaead6/xcm/xcm-builder/src/fungibles_adapter.rs#L41-L46
You should be able to use `match_and_split` instead, though: https://github.com/paritytech/polkadot/blob/cc24fb87b702db6510d08c689e442b6be384d798/xcm/src/v1/multilocation.rs#L249

+ We should probably also add (something like) this to the xcm-builder as @KiChjang suggests. (Would be great if you could open that PR as well to upstream this.)

apopiak (changes requested):
Thank you for the contribution from my side as well, btw. :pray: 

Here are the adjustments I'd like to see:
+ [ ] Please move this into parachains-common. (Statemint and Statemine will have to use the same.)
+ [ ] This also should have a few tests:
  + matching the local prefix
  + matching the remote prefix
  + the failure cases
+ [ ] Please add docs
+ [ ] Also you skipped the prefix checking from the original here:
https://github.com/paritytech/polkadot/blob/55872676fae63399eb111a60ba27cb735cbaead6/xcm/xcm-builder/src/fungibles_adapter.rs#L41-L46
You should be able to use `match_and_split` instead, though: https://github.com/paritytech/polkadot/blob/cc24fb87b702db6510d08c689e442b6be384d798/xcm/src/v1/multilocation.rs#L249

+ We should probably also add (something like) this to the xcm-builder as @KiChjang suggests. (Would be great if you could open that PR as well to upstream this.)

yrong:
@apopiak some refact per your request list. For the xcm-builder upstream refact I would like to do later and we're looking forward this pr merged into `polkadot-v0.9.13`  release

polkadot-parachains/parachains-common/src/impls.rs
```diff
@@ -113,6 +114,59 @@ impl<T: Get<MultiLocation>> FilterAssetLocation for AssetsFrom<T> {
...
+			.iter()
+			.enumerate()
+			.any(|(index, junction)| id.interior().at(index) != Some(junction))
+		{
+			return Err(())
+		}
+
+		match id.interior().at(LocalPrefix::get().interior().len()) {
+			Some(Junction::GeneralIndex(id)) => ConvertAssetId::convert_ref(id),
+			_ => match id.interior().at(RemotePrefix::get().interior().len()) {
```
apopiak, line 156: I would strongly suggest that you use `match_and_split` instead. I have a sense that this is implementation is error prone in edge cases.
e.g. in this case:
```Rust
let id = (1, GeneralIndex(42)).into();
let local_prefix = Here.into();
let remote_prefix = (2, Here).into();
```
I would expect your implementation to to accept the `id` while it should fail (because it matches neither of the parent counts).

yrong: @apopiak you are absolutely right, and since `match_and_split` already do the check interior logic internally I think we can remove some check logic with more clean codes.

apopiak (approved):

apopiak:
Added a basic description. Feel free to expand it.

apopiak:
I'd like someone else to also have a look before merging this.

gavofyork:
Why is this against 0.9.13? Surely it should be going into master?

polkadot-parachains/parachains-common/src/impls.rs
```diff
@@ -113,6 +114,44 @@ impl<T: Get<MultiLocation>> FilterAssetLocation for AssetsFrom<T> {
 	}
 }
 
+/// Converter struct with two main usage scenarios:
```
gavofyork, line 117: Documentation needs improvement. This is incomprehensible at present.

polkadot-parachains/parachains-common/src/impls.rs
```diff
@@ -113,6 +114,44 @@ impl<T: Get<MultiLocation>> FilterAssetLocation for AssetsFrom<T> {
...
+/// 1. Transfer asset from local to other para-chain
+/// 	- with `reverse_ref` to convert a numeric asset ID (must be `TryFrom/TryInto<u128>`) into a `GeneralIndex` junction
+/// 2. Transfer asset back from other para-chain to local
+/// 	- with `convert_ref` to convert multilocation struct `(1,X2(ParaChain(para_id),GeneralIndex(general_index))` to a numeric asset ID
+pub struct AsPrefixedGeneralIndexFromLocalOrRemote<
```
gavofyork, line 122: Wrong place for this - it should be in xcm-builder.

gavofyork (changes requested):
I remain yet to be convinced that this is needed or even a good idea.

polkadot-parachains/statemint/src/lib.rs
```diff
@@ -481,7 +483,7 @@ pub type FungiblesTransactor = FungiblesAdapter<
 	ConvertedConcreteAssetId<
 		AssetId,
 		Balance,
-		AsPrefixedGeneralIndex<Local, AssetId, JustTry>,
+		AsPrefixedGeneralIndexFromLocalOrRemote<Local, Remote, AssetId, JustTry>,
```
gavofyork, line 486: No alterations to Statemint's XCM right now. Stuff like this is EXTREMELY dangerous to change and must a) be completely understood by me; and b) be in place in Kusama for some time prior.

polkadot-parachains/parachains-common/src/impls.rs
```diff
@@ -281,4 +322,59 @@ mod tests {
...
+				&local_asset_location,
+			)
+			.unwrap_or(u32::default());
+		assert_eq!(asset_id, 42);
+		asset_id =
+			AsPrefixedGeneralIndexFromLocalOrRemote::<Local, Remote, u32, JustTry>::convert_ref(
+				&remote_asset_location,
+			)
+			.unwrap_or(u32::default());
+		assert_eq!(asset_id, 42);
```
KiChjang, line 345: Previously I have tried to summarize your problem as the following:

> if I'm hearing you correctly, you're saying that the `GeneralIndex` is used to specify which one of the many assets in statemine that you want to transfer to the destination chain via reserve transfers? In other words, your flow is exactly the same as `InititateReserveWithdraw`, but with the ability to specify an asset that is not KSM, correct?

But this unit test seems to indicate that you're doing something more -- you're trying to represent a remote asset containing a `GeneralIndex` locally as a `GeneralIndex` as well, e.g. a `../Parachain(1000)/GeneralIndex(42)` gets converted and represented as `./GeneralIndex(42)`. Is this what you are trying to do?

If so, then this is absolutely not the right thing to do. Assets should always have __ONE__ canonical `MultiLocation` to represent it, and as such, an asset located at `../Parachain(1000)/GeneralIndex(42)` __MUST__ be interpreted as a different asset from an asset located at `./GeneralIndex(42)`. If the intention is to mint a derivative asset of the one located at `../Parachain(1000)/GeneralIndex(42)`, then it is the `AssetTransactor` logic of your chain that needs to change, and not the XCM configuration.

KiChjang: Pondering a bit more into this, I think what you are trying to solve is the reanchoring issue, in which case, this is an operation that should have been done by the XCM sender, and not the recipient, i.e. when sending an asset containing a `MultiLocation` of `../Parachain(1000)/GeneralIndex(42)` to `../Parachain(1000)`, the XCM executor [should have reanchored this `AssetId` to `./GeneralIndex(42)`](https://github.com/paritytech/polkadot/blob/2d84ad65fb3ac54acab7486426267b3ea5b72681/xcm/xcm-executor/src/lib.rs#L312). Did this not happen when you're testing?

yrong: @KiChjang @apopiak I do not think we can achive that without hack in cumulus, let us put more details here:
  
- transfer asset from statemine to our chain with `TransferReserveAsset`

since [`Local`](https://github.com/paritytech/cumulus/blob/e70f7fd0162fef6ca63bc31466bb2f66469310fa/polkadot-parachains/westmint/src/lib.rs#L467) defined here original asset sent from statemine should be represented as `./GeneralIndex(42)`  and after the [reanchor logic]( https://github.com/paritytech/polkadot/blob/2d84ad65fb3ac54acab7486426267b3ea5b72681/xcm/xcm-executor/src/lib.rs#L312) you mentioned above asset will be converted to `../Parachain(1000)/GeneralIndex(42)` and in our chain we should add  some [adaptor logic](https://github.com/bifrost-finance/bifrost/blob/c7eb1bef7abfdeb6562aeda29ab76805308b77f5/runtime/common/src/xcm_impl.rs#L153-L158) to handle this

- transfer asset from our chain to statemine with `InitiateReserveWithdraw`

 I agree `Assets should always have ONE canonical MultiLocation to represent` so original asset sent from our chain should still be `../Parachain(1000)/GeneralIndex(42)` but after the [reanchor logic here](https://github.com/paritytech/polkadot/blob/24758c5ab5e6c7f99e9af08e583f6084f5d087a6/xcm/xcm-executor/src/lib.rs#L410) will still be `../Parachain(1000)/GeneralIndex(42)` and we can not convert it back to `./GeneralIndex(42)` so statemine will not understand without hack

KiChjang: A couple of observations:

1. What you interact on your local chain should _always_ be `./GeneralIndex(42)` and not `../Parachain(1000)/GeneralIndex(42)`, i.e. you interact with the derivative asset and never the reserve asset. Therefore when you do a reserve asset transfer, you will first execute a `WithdrawAsset(GeneralIndex(42))` and you will get some `GeneralIndex(42)` stored in your chain's holding register.
2. The next step in the process is `InitiateReserveWithdraw`, which takes the derivative asset in the holding register and sends it back to Statemine. Before that happens though, what should happen on your location chain is that it should burn and convert the derivative asset (i.e. `GeneralIndex(42)`) to the reserve asset (i.e. `../Parachain(1000)/GeneralIndex(42)`).
3. The receiver should never be responsible for converting between derivative asset IDs and reserve asset IDs. Doing so would mean that all reserve assets should "register" their derivatives with the reserve chain, which is absolutely not the direction we want to head towards.

I've looked into the code for the XCM executor now and it does seem like there's no way to burn and convert a derivative asset back to its reserve asset during the execution of `InitiateReserveWithdraw`, and I think that is the crux of the issue here.

yrong: @KiChjang 
> What you interact on your local chain should always be `./GeneralIndex(42)` and not `../Parachain(1000)/GeneralIndex(42)` i.e. you interact with the derivative asset and never the reserve asset.

No, I do think we should use reserve asset rather than derivative asset here. except for statemine we need to talk to other parachains like Acala and as you mentioned before only the GeneralIndex is too abstract. How to differ the same derivateive asset `./GeneralIndex(42)` in acala and statemine?

KiChjang: Sorry, but your opinion is literally not how XCM works with reserve assets by definition. If you want to use XCM to hold, receive and transfer assets that are not reserve-backed, then you will use asset teleportation instead with your destination chain. Now obviously that would require you to have an agreement with the destination chain to configure each other as trusted teleport locations, and more generally speaking, we will not support a use case where we break the reserve asset model to solve a particular problem.

If you are still inclined to work with this model that you have in mind, then my suggestion would be that you formulate a complete conceptual model, explain why the current reserve asset model does not work and the deficiencies of it, and propose that change in https://github.com/paritytech/xcm-format.

yrong: @KiChjang Ok, let me put this way, do you think it is possible to support this very basic requirement without introducing new xcm primitives or any hack in cumulus? 
 no big deal for the forward direction from statemine to us, we can do any logic to recognize whatever format defined in statemine. but for the reverse direction, no matter reserved based or teleport based, In current xcm implementation I'am afraid assets will already contains some parachain prefix after the reanchor logic so statemine can not recognize without hack. Correct me if I'am wrong.

KiChjang: It seems to me that there's a misunderstanding of what it means when your chain receives a `ReserveAssetDeposited` instruction. `ReserveAssetDeposited` is a _notification_ XCM instruction, meaning it does _not_ actually contain any assets in its parameters; instead, it _notifies_ the recipient that some assets were deposited in their sovereign account on the reserve chain.

Thus, when your chain receives `ReserveAssetDeposited` from the reserve chain, it should then perform the necessary operations to mint a derivative asset that represents the amount of reserve assets that has been transferred to its sovereign account on the reserve chain, and should __not__ be directly minting the reserve asset on your chain, as it is __not__ an asset teleportation.

I suspect this is why you thought it is necessary to use the reserve asset on your chain, and that it is a "very basic" operation. It is not. We do not provide any opinionated approaches as to how you should process and mint derivative assets; it is up to your chain to investigate and invent solutions on how exactly a chain should perform bookkeeping on the assets that they own on the reserve chain. I don't believe this is a new problem however -- there should already been successful examples of how exactly it is done in the ecosystem.

yrong: @KiChjang I can use the derivative reprensentation(`./GeneralIndex(42)`) in our chain when sent back with `InitiateReserveWithdraw` and actually in a previous [commit](https://github.com/bifrost-finance/bifrost/commit/97476948afda895313ad399048d0ba9471f3fcd9) I did try this way, though which means there are two diffrent repensentations for the same asset when received and sent back which seems weird and actually breaks the principle you mentioned `Assets should always have ONE canonical MultiLocation to represent` I can bear with it.  You may check discussion with @apopiak [here](https://github.com/bifrost-finance/bifrost/pull/420#discussion_r761678588), 

The problem is also what you mentiond  like `there's no way to convert a derivative asset back...during the execution of InitiateReserveWithdraw, and I think that is the crux of the issue here`

KiChjang: That could easily be solved if you prepend an `ExchangeAsset` instruction before you execute `InitiateReserveWithdraw`, and in `ExchangeAsset`, your custom XCM executor will convert your derivative asset back to the reserve asset.

This is only one of the ways in which you can solve this; there are many others and I really don't think this is a novel issue -- some other parachains in the ecosystem should have solved this already, and it would be really surprising to me if you're the only chain currently that's running into this problem.

yrong: > That could easily be solved if you prepend an ExchangeAsset instruction before you execute InitiateReserveWithdraw, and in ExchangeAsset, your custom XCM executor will convert your derivative asset back to the reserve asset.

 @KiChjang I really do not know this primitive and will try, thanks for your suggestion

yrong: @KiChjang 
seems [ExchangeAsset](https://github.com/paritytech/polkadot/blob/4468927b770eb1515799a9d4aba37b8d1645d787/xcm/xcm-executor/src/lib.rs#L493) you mentioned not implemented yet, is it in some feature branch? And I would like to know if there is any other solution or implementation other parachain team already solved as you mentioned?

KiChjang: It is purposefully not implemented; as I said, we do not provide any opinionated approaches as to how a chain should manage their bookkeeping of their reserve assets. I'm not as familiar as to how other parachain teams in the ecosystem solves this problem, however, but I would think that any parachain that has onboarded to Kusama and has dealt with reserve assets would have come up with a solution dealing with it.

yrong: Understood. we can implemente this to do some asset convert logic from sender side. But I would assume at last we still need `InitiateReserveWithdraw` primitive to send asset back to statemine,right? From what I've tested after the renchor logic in InitiateReserveWithdraw [here](https://github.com/paritytech/polkadot/blob/4468927b770eb1515799a9d4aba37b8d1645d787/xcm/xcm-executor/src/lib.rs#L410) asset will still be converted to `../Parachain(1000)/GeneralIndex(42)` no matter what it previously is and which can not be recognized by statemine which only knows `./GeneralIndex(42)`

KiChjang: Correct, the reanchoring logic is what we're in the middle of investigating as well. However, in the context of this PR, your changes will unfortunately not be the correct fix. We will provide an update in #827 as soon as we have any more details to share.

gavofyork:
https://github.com/paritytech/cumulus/issues/827#issuecomment-986074970

As I mention here, the primitives in the XCM builder expect that the derivative asset (on your chain, which represented ownership of the reserve assets on Statemint) should generally be represented in XCM as the _same_ conceptual location as the reserve asset.

So, generally this will mean it is `GeneralIndex(42)` on Statemint and `(Parent, Parachain(1000), GeneralIndex(42))` on your parachain.

Reanchoring should ensure that the former `MultiLocation` value (used on Statemint) gets changed into the latter `MultiLocation` value (used on your chain) and back again.

If you believe there is an issuer with the reanchoring, then I'll just need a failing test to demonstrate this. So far I have not seen one.

You will likely want to have an instance of the `Assets` pallet which represents all Statemint assets. You might also write your own `Convert<MultiLocation, AssetId>` to feed into the `FungiblesAdapter` which would effectively "namespace" the `Statemint` assets within your general assets pallet; to do this you'd need to ensure that your `AssetId` type is big enough to contain all Statemint `AssetId` values as well as any other `AssetId` values specific to your chain which you might want.

gavofyork:
Ok, I believe I figured it out. There is indeed a subtle issue with the reanchoring logic, specifically it gives a correct non-canonical destination value. Basically, it sends a `MultiLocation` to Statemint of `(Parent, Statemint, 42)` rather than just `(42)`. From Statemint's point of view, they're equivalent, only the former includes a superfluous `(Parent, Statemint)` prefix.

I'll put in something to simplify it and stuff should work.

gavofyork:
https://github.com/paritytech/polkadot/pull/4470 should fix things

yrong:
>

yrong:
@KiChjang I also see this [pr](https://github.com/paritytech/cumulus/pull/833) which seems we decide to add pallet index in between to repensentate asset in statemine, right? if so we may also revamp adaptor logic from our side accordingly.

gavofyork:
> @gavofyork @KiChjang seems this fix will break current behavior how we bifrost(2001) communicate with acala(2000) and other parachains, breaking test cases as following:
> 
> ```
> #[test]
> fn reanchor_1_works() {
>   let mut id: MultiLocation = (Parent, Parachain(2000), GeneralKey(42.encode())).into();
>   let ancestry = Parachain(2001).into();
>   let target = (Parent, Parachain(2000)).into();
>   let expected = (Parent, Parachain(2001), GeneralKey(42.encode())).into();
>   id.reanchor(&target, &ancestry).unwrap();
>   assert_eq!(id, expected);
> }
> ```
> 
> maybe @xlc can add some comments here also

Yes. This test is very much incorrect. The `GeneralKey(42)` exists on parachain #2000, not #2001.

From parachain #2001's point of view (where we begin):
1. ancestry: `Parachain(2001)`
2. parachain #2000: `(Parent, Parachain(2000))`
3. parachain #2001: `Here`
4. the asset: `(Parent, Parachain(2000), GeneralKey(42))`

From parachain #2000's point of view (where we reanchor to):
5. ancestry: `Parachain(2000)`
6. parachain #2000: `Here`
7. parachain #2001: `(Parent, Parachain(2001))`
8. the asset: `GeneralKey(42)`

So the reanchoring should take the ancestry of the location where we begin (1), the destination (relative to where we begin) (2) and the asset (also relatively to where we begin) (4).

From these three pieces of information, it is able to tell us how the destination would refer to the same asset (8).

gavofyork:
Closing since this is definitely not the right fix.

Issue #807, opened 2021-11-29
Statemine balances are too low


NachoPal:
Statemine balances are too low for running some integration test. Some extriniscs such as `assets:create` are very costly and the default initial balance is not enough. Balance is also too low to chain multiple tests and pay the fees for all of them.

PR #808, opened 2021-11-29
Increase statemine balance

NachoPal:
Fixes: #807

### DIFF
--- a/polkadot-parachains/src/chain_spec.rs
+++ b/polkadot-parachains/src/chain_spec.rs
@@ -498,7 +498,7 @@ fn statemine_genesis(
 				.to_vec(),
 		},
 		balances: statemine_runtime::BalancesConfig {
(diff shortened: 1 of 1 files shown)


apopiak (approved):

bkchr (approved):

apopiak:
needs formatting

apopiak:
We should probably do the same for Statemint and Westmint?

NachoPal:
> We should probably do the same for Statemint and Westmint?

Thanks for the formatting @chevdor 

Westmint & Statemint don't need it, the balance depends on the ED.

Statemine -> `pub const EXISTENTIAL_DEPOSIT: Balance = CENTS / 10;`
Statemint ->  `pub const EXISTENTIAL_DEPOSIT: Balance = 10 * CENTS;`
Westmint ->  `pub const EXISTENTIAL_DEPOSIT: Balance = 100 * MILLICENTS;`

PR #812, opened 2021-11-29
Ensure invulnerables have associated validator keys before they are set

omadoyeabraham:
- Added check to ensure invulnerables have associated validator keys before they are set


bkchr (approved):
A test would be nice, otherwise looks good.

KiChjang:
Needs a test, and a benchmark run to update the weights for this dispatchable.

pallets/collator-selection/src/lib.rs
```diff
@@ -292,6 +292,17 @@ pub mod pallet {
 					"invulnerables > T::MaxInvulnerables; you might need to run benchmarks again"
 				);
 			}
+
+			// check if the invulnerables have associated validator keys before they are set
+			for account_id in new {
```
KiChjang, line 297: CI is failing due to ownership issues on this vec, use a reference to avoid:

```suggestion
			for account_id in &new {
```

riusricardo (commented):
Looks good to me. Thanks @omadoyeabraham for the PR. The check is a good idea. As Basti and Keith said, a test and benchmarks will help here.

omadoyeabraham:
@riusricardo @KiChjang @bkchr 

Thanks for the feedback, I'm working on the tests and benchmarking and will update the PR once done

gilescope:
Did you get success with the tests? One thing to note on the mock runtime for those tests is that I found that 4 is always picked as the collator. It is fixed and does not rotate.

bkchr:
> Needs a test, and a benchmark run to update the weights for this dispatchable.

@KiChjang in what way does this require changing the weight of the call? How does this change the worst case of the tx?

bkchr:
@omadoyeabraham you need to accept the CLA.

pallets/collator-selection/src/lib.rs
```diff
@@ -292,6 +292,17 @@ pub mod pallet {
...
 				);
 			}
+
+			// check if the invulnerables have associated validator keys before they are set
+			for account_id in &new {
+				let validator_key = T::ValidatorIdOf::convert(account_id.clone())
+					.ok_or(Error::<T>::NoAssociatedValidatorId)?;
+				ensure!(
+					T::ValidatorRegistration::is_registered(&validator_key),
+					Error::<T>::ValidatorNotRegistered
```
bkchr, line 302: I think actually changing the event to `Error::<T>::ValidatorNotRegistered(account_id)` would be nice.

pallets/collator-selection/src/lib.rs
```diff
@@ -292,6 +292,17 @@ pub mod pallet {
 					"invulnerables > T::MaxInvulnerables; you might need to run benchmarks again"
 				);
 			}
+
+			// check if the invulnerables have associated validator keys before they are set
+			for account_id in &new {
```
KiChjang, line 297: @bkchr There's now a new loop here that's dependent on the number of elements inside of this `new` vector. I see that the weight function already takes in the length of `new` as part of the weight formula, but at the very least, the constant that it needs to multiply by has surely increased due to this loop here.

bkchr: But shouldn't that be taken care of by the benchmarking? Aka we run this function with different number of inputs and that should give us the weight per input size.

KiChjang: Yeah, I suppose we just need to run the benchmark again using benchbot and that would suffice.

bkchr: Yeah, but that isn't a requirement for this pr. This needs to be done for a release.

KiChjang: Right ok, if it's part of the release process, then we don't need to update the weights in this PR.

bkchr:
Ping @omadoyeabraham

omadoyeabraham:
Hi @bkchr 

I'm unable to sign the CLA, I keep getting the following error message

```
ERROR: signature_token is missing or is invalid.
```

bkchr:
@omadoyeabraham now you fucked up the cla again :D

bkchr:
bot merge

omadoyeabraham:
@bkchr Apologies, I recently switched to a new PC and had a Github account mixup :)

seunlanlege:
this pr broke benchmarks for collator-selection 😭

PR #813, opened 2021-11-30
Refactor some types in XCMP queue

KiChjang:
Because seeing `.1` and `.2` irks me to no end.

Mostly renames, but also contains some helper methods to construct a new type using the builder pattern.

### DIFF
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -198,7 +198,7 @@ pub struct OutboundChannelDetails {
 	recipient: ParaId,
 	/// The state of the channel.
 	state: OutboundState,
-	/// Whether or not any signals exists in this channel.
+	/// Whether or not any signals exist in this channel.
 	signals_exist: bool,
 	/// The index of the first outbound message.
 	first_index: u16,
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -124,9 +124,9 @@ pub mod pallet {
 
 	/// Status of the inbound XCMP channels.
 	#[pallet::storage]
-	pub(super) type InboundXcmpStatus<T: Config> = StorageValue<
+	pub(super) type InboundChannelStatus<T: Config> = StorageValue<
 		_,
-		Vec<(ParaId, InboundStatus, Vec<(RelayBlockNumber, XcmpMessageFormat)>)>,
+		Vec<InboundChannelDetails>,
 		ValueQuery,
 	>;
 
@@ -169,7 +169,7 @@ pub mod pallet {
 }
 
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, RuntimeDebug, TypeInfo)]
-pub enum InboundStatus {
+pub enum InboundState {
 	Ok,
 	Suspended,
 }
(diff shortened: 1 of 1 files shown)


pallets/xcmp-queue/src/lib.rs
```diff
@@ -149,8 +146,8 @@ pub mod pallet {
 	/// case of the need to send a high-priority signal message this block.
 	/// The bool is true if there is a signal message waiting to be sent.
 	#[pallet::storage]
-	pub(super) type OutboundXcmpStatus<T: Config> =
-		StorageValue<_, Vec<(ParaId, OutboundStatus, bool, u16, u16)>, ValueQuery>;
+	pub(super) type OutboundChannelStatus<T: Config> =
```
shawntabrizi, line 149: How important is this storage name change?

Will need migrations for this right?

shawntabrizi (commented):
Tuple to struct is 👍 

I see no issues there since they will encode the same.

However, the subtle name change to the storage will change the derived key. Probably not worth, or at least make a second PR which isolates these breaking changes from things which are not breaking.

I will approve when either a migration is created, or the names are changed back.

pallets/xcmp-queue/src/lib.rs
KiChjang: Ah right, it does require a migration change... maybe it's time to use the `#[pallet::rename_as]` attribute that I added a while ago?

KiChjang: I also didn't check whether a struct is SCALE-encoded the same way as a tuple, so I may have to do a migration anyway.

KiChjang: Well ok, I am more confused after reading https://docs.substrate.io/v3/advanced/scale-codec/#data-structures, but since you said that structs encode the same as tuples, I guess it's fine?

shawntabrizi: > I also didn't check whether a struct is SCALE-encoded the same way as a tuple, so I may have to do a migration anyway.

It is the same, but you can write a simple test to confirm this using `encode()` on the old tuple and new struct.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -146,6 +147,7 @@ pub mod pallet {
 	/// case of the need to send a high-priority signal message this block.
 	/// The bool is true if there is a signal message waiting to be sent.
 	#[pallet::storage]
+	#[pallet::storage_prefix = "OutboundXcmpStatus"]
```
shawntabrizi, line 154: is the name here really that important to change it and have this code?

I really think we should separate this change from the rest, and then make a separate decision.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -169,17 +166,68 @@ pub mod pallet {
...
+}
+
+/// Struct containing detailed information about the outbound channel.
+#[derive(Clone, Eq, PartialEq, Encode, Decode, TypeInfo)]
+pub struct OutboundChannelDetails {
+	/// The `ParaId` of the parachain that this channel is connected with.
+	recipient: ParaId,
+	/// The state of the channel.
+	state: OutboundState,
+	/// Whether or not any signals exists in this channel.
```
apopiak, line 201:
```suggestion
	/// Whether or not any signals exist in this channel.
```

apopiak (approved):
LGTM

shawntabrizi (approved):

KiChjang:
bot merge

PR #757, opened 2021-11-12
WIP Statemint Genesis Values

apopiak:
This PR will add the genesis values that will be injected into shell to prepare the upgrade to Statemint.

based on keys from https://github.com/paritytech/devops/issues/1194
part of https://github.com/paritytech/cumulus/issues/744

PR #817, opened 2021-12-01
Add Statemint Genesis Values

apopiak:
Adds the genesis values for the upcoming Statemint release and adjusts the scripts to generate them.

+ `statemint_genesis.json` is a chainspec for reference only. It is the basis for:
  + `statemint_genesis_values.json`, the values that will be set on Shell to prepare for the upgrade
  + `statemint_genesis_values.scale`, the values encoded as a `System::set_storage` call
+ `statemint.json` is the spec that will allow syncing the Statemint chain after the upgrade.

based on keys from https://github.com/paritytech/devops/issues/1194
part of https://github.com/paritytech/cumulus/issues/744


apopiak:
#757 was somewhat reviewed already, but would appreciate another check

NachoPal (approved):

polkadot-parachains/src/chain_spec.rs
```diff
@@ -315,6 +315,64 @@ pub fn statemint_local_config(id: ParaId) -> StatemintChainSpec {
...
+				vec![],
+				id,
+			)
+		},
+		vec![
+			"/ip4/34.65.251.121/tcp/30334/p2p/12D3KooWG3GrM6XKMM4gp3cvemdwUvu96ziYoJmqmetLZBXE8bSa".parse().unwrap(),
+			"/ip4/34.65.35.228/tcp/30334/p2p/12D3KooWMRyTLrCEPcAQD6c4EnudL3vVzg9zji3whvsMYPUYevpq".parse().unwrap(),
+			"/ip4/34.83.247.146/tcp/30334/p2p/12D3KooWE4jFh5FpJDkWVZhnWtFnbSqRhdjvC7Dp9b8b3FTuubQC".parse().unwrap(),
+			"/ip4/104.199.117.230/tcp/30334/p2p/12D3KooWG9R8pVXKumVo2rdkeVD4j5PVhRTqmYgLHY3a4yPYgLqM".parse().unwrap(),
+		],
```
riusricardo, line 368: Interesting that you added the bootnodes here.

riusricardo (commented):
The chain spec and genesis values look good to me.

polkadot-parachains/src/chain_spec.rs
apopiak: I figured why not. If we do update them frequently I might have to remove them again :shrug:

PR #824, opened 2021-12-02
Seedling Runtime + Node

NachoPal:
Basic implementation of a upgradable Seedling runtime/node.

While the PR is reviewed, I'll start adding the new pallet for `set_custom_validation_head_data` implementation.


polkadot-parachains/seedling/src/lib.rs
```diff
@@ -0,0 +1,325 @@
...
+			Call::ParachainSystem(cumulus_pallet_parachain_system::Call::set_validation_data {
+				..
+			}) => true,
+			Call::Sudo(pallet_sudo::Call::sudo_unchecked_weight { call: ref x, .. }) => {
+				matches!(x.as_ref(), &Call::System(frame_system::Call::set_code { .. }))
+			},
+			_ => false,
+		}
+	}
+}
```
riusricardo, line 120: I like the filter, it is a good idea. We also need to consider if we need to allow other calls in case of any emergency situation.

riusricardo (approved):
Everything is working fine. It is ready to get merged.

PR #825, opened 2021-12-02
Adds support for returning a custom header from `validate_block`

bkchr:
This adds support for returning a custom header from `validate_block`. Before this, we always
returned the header of the block that was validated (and still do it by default). However, after
this pr it is possible to set a custom header or better custom head data that will be returned
instead from `validate_block`. This can be for example when a chain wants to fork.


skunert (approved):

riusricardo (approved):
LGTM. The usage is very straightforward :+1:

PR #830, opened 2021-12-04
Update Statemint prefix

riusricardo:


polkadot-parachains/src/chain_spec.rs
```diff
@@ -318,6 +320,7 @@ pub fn statemint_local_config(id: ParaId) -> StatemintChainSpec {
 // Not used for syncing, but just to determine the genesis values set for the upgrade from shell.
 pub fn statemint_config(id: ParaId) -> StatemintChainSpec {
 	let mut properties = sc_chain_spec::Properties::new();
+	properties.insert("ss58Format".into(), 0.into());
```
hbulgarini, line 323: Great! should we add the equivalent on here: https://github.com/paritytech/cumulus/blob/master/polkadot-parachains/src/chain_spec.rs#L355

I suspect for statemine we should add something like:

`properties.insert("ss58Format".into(), 2.into());`

hbulgarini: Maybe in this PR we might solve the same issue for both chains. What do you think?

hbulgarini:
@chevdor , just to confirm ... these are the kind of PR that you don't want to `Squash and merge` right?

chevdor:
Any significant PR landing in a `release-*` branch should show up and not be squashed indeed.

Be aware that committing to such a branch will [create a new RC](https://github.com/paritytech/cumulus/blob/master/.github/workflows/release-01_rc-automation.yml#L2-L6) for each commit will be pushed.

So the best is to prepare the set of changes in another branch (with a name **not** starting with `release`) and then have a PR bringing the new set of changes to the release branch **without** squashing. This will trigger the next RC.

It is cleaner to have those important changes proposed against master (master is always there and there is a single one) and then backport them in the release branch as required.

riusricardo:
@chevdor In this case there is no way good way to do all the changes against master. This is why I asked about the release branch merge to master. There are files that only exist here.

polkadot-parachains/src/chain_spec.rs
joepetrowski: That's the dev config, we usually just use the default `42` for that

hbulgarini: Ok! this is fine, i would suggest to keep the same config for everything ( with the  ss58Format on the properties or not)

riusricardo:
That doesn't makes sense. This will require 2 PRs instead of one because master is missing 2 places where the changes need to be made. You will end up on the same situation unless the changes from this PR get into master first.

joepetrowski (approved):

PR #831, opened 2021-12-04
Update XCM MultiLocation prefixes for assets

KiChjang:
Ensure that any assets under the Assets pallet is prefixed by a `PalletInstance(assets_pallet_index)` junction.


gavofyork (approved):

shawntabrizi (approved):
This is to allow multiple different kinds of assets pallets on the parachain?

KiChjang:
Yes, currently `GeneralIndex` is only used to describe fungibles via the assets pallet; in the future, we may want to include the `uniques` pallet as well, or maybe some other pallet describing other kinds of assets, and it's best that we future-proof it now.

KiChjang:
bot merge

PR #832, opened 2021-12-04
Update XCM MultiLocation prefixes for assets in Statemine

KiChjang:


joepetrowski:
I think we should close this in favor of #831 and just include this change in the v7* release of Statemint/e

KiChjang:
Fair enough.

PR #836, opened 2021-12-06
Add CI to build docker images on releases

chevdor:
This PR bring a new workflow to build and publish injected docker images. 
The workflow only allows and consider binaries that have been GPG signed.

```
# This workflow listens to pubished releases.
# It includes releases and pre-releases.
# It fetches the binaries, checks sha256 and GPG
# signatures, then builds an injected docker
# image and publishes it.
```

This PR requires the following secrets:
- [ ] `DOCKERHUB_USERNAME `: `parity`
- [ ] `DOCKERHUB_TOKEN `: `<docker hub token>`


s3krit (approved):
Looks awesome!

Issue #834, opened 2021-12-06
Disable Teleports in Statemint at Launch


joepetrowski:
Based on https://github.com/paritytech/polkadot/issues/4462, the Relay Chain will not process upwards messages when parachains launch. We will need to disable teleports in both directions between Statemint and the Relay Chain.

PR #837, opened 2021-12-06
Disable asset transfers on Statemint (against Statemint release branch v6.0.1)

KiChjang:
#834

### DIFF
--- a/polkadot-parachains/statemine/src/lib.rs
+++ b/polkadot-parachains/statemine/src/lib.rs
@@ -584,8 +584,8 @@ impl pallet_xcm::Config for Runtime {
 	// ... but disallow generic XCM execution. As a result only teleports and reserve transfers are allowed.
 	type XcmExecuteFilter = Nothing;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Everything;
+	type XcmTeleportFilter = Nothing;
+	type XcmReserveTransferFilter = Nothing;
 	type Weigher = FixedWeightBounds<UnitWeightCost, Call, MaxInstructions>;
 	type LocationInverter = LocationInverter<Ancestry>;
 	type Origin = Origin;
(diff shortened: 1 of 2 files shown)


polkadot-parachains/statemine/src/lib.rs
```diff
@@ -584,8 +584,8 @@ impl pallet_xcm::Config for Runtime {
 	// ... but disallow generic XCM execution. As a result only teleports and reserve transfers are allowed.
 	type XcmExecuteFilter = Nothing;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Everything;
+	type XcmTeleportFilter = Nothing;
+	type XcmReserveTransferFilter = Nothing;
```
apopiak, line 588: This changes Statemine, should be Statemint. (same for #838

gavofyork (approved):

gavofyork:
This is safe, but I don't know if we want to disable the reserve asset transfer or just the teleporting?

KiChjang:
Hmm ok, if we disable reserve asset transfers, that means no one would be able to transfer reserve assets via the `reserve_transfer_assets` extrinsic on the XCM pallet. However, this does not mean `InitiateReserveWithdraw` XCMs are going to be disabled, since they're processed on the source chain first, and then subsequently a `WithdrawAsset` instruction is sent to Statemint and then executed, and the `XcmReserveTransferFilter` is not used.

All this is to say that disabling `XcmReserveTransferFilter` only affects whether or not users are able to dispatch `reserve_transfer_assets` on the XCM pallet on Statemint, and other XCM operations dealing with reserve assets are not affected. I can see that it can be a prudent step for us to take to set `XcmReserveTransferFilter` to `Nothing` if we are being conservative here and do not want to have any funny business going on with people attempting to exploit `reserve_transfer_assets`.

joepetrowski:
Unless we change https://github.com/paritytech/polkadot/pull/4475 to allow reserve transfers to Statemint, I don't think it matters here as no DOT would arrive on Statemint (and thus asset creation isn't possible either because nobody could pay the deposit). I would rather restrict them so that opening them on the Relay Chain doesn't have the effect of making them bi-directional. That is, we allow teleports/reserve asset transfer on Statemint only after allowing them from the Relay Chain.

PR #842, opened 2021-12-08
Companion for #10403: Remove Default for AccountId

gavofyork:
paritytech/substrate#10403


parachain-template/runtime/src/lib.rs
```diff
@@ -763,6 +763,7 @@ impl_runtime_apis! {
 
 			list_benchmark!(list, extra, frame_system, SystemBench::<Runtime>);
 			list_benchmark!(list, extra, pallet_balances, Balances);
+			list_benchmark!(list, extra, pallet_session, SessionBench::<Runtime>);
```
gavofyork, line 766: I presume this should be in here @bkchr ?

bkchr: Probably. The template wasn't created by me.

shawntabrizi (approved):

pallets/collator-selection/src/mock.rs
```diff
@@ -225,21 +225,19 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
 	sp_tracing::try_init_simple();
 	let mut t = frame_system::GenesisConfig::default().build_storage::<Test>().unwrap();
 	let invulnerables = vec![1, 2];
-	let keys = invulnerables
+
+	let balances = vec![(1, 100), (2, 100), (3, 100), (4, 100), (5, 100)];
+	let keys = balances
 		.iter()
-		.map(|i| (*i, *i, MockSessionKeys { aura: UintAuthorityId(*i) }))
+		.map(|&(i, _)| (i, i, MockSessionKeys { aura: UintAuthorityId(i) }))
```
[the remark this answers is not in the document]
gavofyork: Yes - other keys were used in tests as validators, but they won't show up in the validator list unless they have a session key defined (this is the change).

They always should have had a session key defined, but we got away without it previously because we used a very lazy and not very correct `.unwrap_or_default()` when retrieving the validators' session keys.

shawntabrizi:
bot merge

PR #845, opened 2021-12-08
Seedling runtime fix

NachoPal:
`system.SetCode` does not allow upgrades for runtimes with different `spec_name`: https://github.com/paritytech/substrate/blob/master/frame/system/src/lib.rs#L1523-L1525 and we expect Parachains to have their own `spec_name` (different to `seedling`)

With this fix the runtime upgrade will be done as follows:
1. `sudo.SudoUncheckedWeight(parachainSystem.AuthorizeUpgrade(code_hash)))`
2. `parachainSystem.enactAuthorizedUpgrade(code)` Unsigned Tx


bkchr (changes requested):
As @riusricardo already said in element, we don't need the filter. The runtime has no balances module and there are no one with any balance, meaning that all signed calls are not callable. Only unsigned calls are doable and that shows us some problems here, we need to create more unsigned transactions to make this work.

NachoPal:
> As @riusricardo already said in element, we don't need the filter. The runtime has no balances module and there are no one with any balance, meaning that all signed calls are not callable. Only unsigned calls are doable and that shows us some problems here, we need to create more unsigned transactions to make this work.

`BaseCallFilter` removed. I added instead a `CheckSudo` `SignedExtension` in the bare-bones of the `SoloToPara` pallet that will need to be updated by this: #846 I couldn't ban unsigned txs because we need `ParachainSystem::set_validation_data` to go through.

riusricardo:
Fixes were merged into #846

PR #857, opened 2021-12-13
Decouple protocol level from application level for XCM

mustermeiszer:
**NOTE:** This PR is not meant to be seen as "the code that is meant to be included" but rather I am using the PR mechanism as it makes explaining the changes and where they could apply in the codebase more easily understandable than if I wrote it down in an issue.

## Receiving Side
Currently, the protocol level of XCM on the receiving side (namely `cumulus-pallet-xcmp-queue`, `cumulus-pallet-dmp-queue`) passes the received XCM directly into the executor provided by the pallet. 

Instead of doing this, I was curious to see if there are any chance to bring in a trait of the following definition: 

```rust
/// A handler that processes XCMs provided from the protocol layer. The message handler is allowed to consume at most
/// `free_weight` Weight.
pub trait XcmMessageHandler<Call, Weight> {
	/// Handles the message and returns the Weight consumed by this call.
	fn handle_message(origin: VersionedMultiLocation, msg: VersionedXcm<Call>, free_weight: Weight) -> Weight;
}
```

I see the following benefits here (arguably might not apply as I am not aware of future changes and the reasonings behind the current approach): 
* Allows parachains to process XCMs of multiple versions as we are passing the `Versioned*` types here.
* Allows parachains to have full control over their executor's errors and how to handle them.
* Decouple the protocol layer from the application layer (My personal opinion is that the protocol queue should not decide how to handle `WeightLimitReached` but rather take care of protocol stuff, like the `Signals`, errors on closed/congested channels).

## Sending Side
Currently, the trait definition of the sending side of the application level is defined as (using the version aware `SendXcm` trait of latest)

```rust
pub trait SendXcm {
	/// Send an XCM `message` to a given `destination`.
	///
	/// If it is not a destination which can be reached with this type but possibly could by others, then it *MUST*
	/// return `CannotReachDestination`. Any other error will cause the tuple implementation to exit early without
	/// trying other type fields.
	fn send_xcm(destination: impl Into<MultiLocation>, message: Xcm<()>) -> SendResult;
}
```

and implemented by the protocol layer (namely `struct ParentAsUmP` and `cumulus-pallet-xcmp-queue`. 
Would it be possible to rathe have the protocol level implement the following trait 

```rust
/// A sink that takes `VersionedXcm`'s and sends them to the specified `VersionedMultilocation`
pub trait XcmSink<Call> {
        // Not sure about the return type though, as the current error is version aware...
	fn send(recv: VersionedMultiLocation, msg: VersionedXcm<Call>) -> Result<(), ()>;
}
```

I see the following benefits here (arguably might not apply as I am not aware of future changes and the reasonings behind the current approach): 
* Protocol layer is able to send different versions of XCM (I understand that the executor must be version aware and the `SendXcm` represents this restriction for the executor, but I think that the protocol level should be able to send older and newer versions of XCMs)

### DIFF
--- a/pallets/dmp-queue/src/lib.rs
+++ b/pallets/dmp-queue/src/lib.rs
@@ -85,7 +85,8 @@ pub mod pallet {
 		/// The overarching event type.
 		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 
-		type XcmExecutor: ExecuteXcm<Self::Call>;
+		/// The actual handler for xcm messages
+		type XcmHandler: XcmMessageHandler<Self::Call, Weight>;
 
 		/// Origin which is allowed to execute overweight messages.
 		type ExecuteOverweightOrigin: EnsureOrigin<Self::Origin>;
@@ -240,16 +241,8 @@ pub mod pallet {
 					Ok(0)
 				},
 				Ok(Ok(x)) => {
-					let outcome = T::XcmExecutor::execute_xcm(Parent, x, limit);
-					match outcome {
-						Outcome::Error(XcmError::WeightLimitReached(required)) =>
-							Err((id, required)),
-						outcome => {
-							let weight_used = outcome.weight_used();
-							Self::deposit_event(Event::ExecutedDownward(id, outcome));
-							Ok(weight_used)
-						},
-					}
+					let consumed = T::XcmHandler::handle_message(Parent, x, limit);
+                    Ok(consumed)
 				},
 			}
 		}
--- a/pallets/xcm/src/lib.rs
+++ b/pallets/xcm/src/lib.rs
@@ -191,3 +191,15 @@ where
 		_ => Err(BadOrigin),
 	}
 }
+
+/// A sink that takes `VersionedXcm`'s and sends them to the specified `VersionedMultilocation`
+pub trait XcmSink<Call> {
+	fn send(recv: VersionedMultiLocation, msg: VersionedXcm<Call>) -> Result<(), ()>;
+}
+
+/// A handler that processes XCMs provided from the protocol layer. The message handler is allowed to consume at most
+/// `free_weight` Weight.
+pub trait XcmMessageHandler<Call, Weight> {
+	/// Handles the message and returns the Weight consumed by this call.
+	fn handle_message(origin: VersionedMultiLocation, msg: VersionedXcm<Call>, free_weight: Weight) -> Weight;
+}
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -70,8 +70,8 @@ pub mod pallet {
 	pub trait Config: frame_system::Config {
 		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
 
-		/// Something to execute an XCM message. We need this to service the XCMoXCMP queue.
-		type XcmExecutor: ExecuteXcm<Self::Call>;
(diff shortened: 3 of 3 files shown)


gavofyork:
Why would sending an old version of XCM be advantageous?

mustermeiszer:
I thought XCM was supposed to have overlapping compatibility between versions and hence chain A could be using XCM V4 while I am already on XCM V5. Otherwise, chains will be dependent on always being in sync with respect to the XCM version they are using when they want to communicate. If this is not the case, the point is of course negligible. 

And what do you think about the removing the executor from this layer? 

Again, I am totally not in the full picture here, so sorry if I am missing something (or a lot)

KiChjang:
Such a proposal requires a strict definition of what "application-level" and "protocol-level" means. Currently I don't think we use any of these terms in the XCM codebase, and so it is really difficult to understand and see the benefits of a change like this.

In addition, the queues themselves already accepts and parses `VersionedXcm`s when handling messages, and the XCM executor is completely configurable by the implementer, so I don't understand why there's suddenly an impetus for fixing something that is already working.

mustermeiszer:
Correct me if I am wrong, but the W3F names XCMP as ["[...] Polkadot’s cross-chain message-passing protocol" ](https://research.web3.foundation/en/latest/polkadot/XCMP/index.html). And the [polkadot wikis](https://wiki.polkadot.network/docs/learn-crosschain) stack overview shows it also in the transport protocol section, alongside VMP of course. 
And the names of the pallets (`cumulus-pallet-xcmp-queue`, `cumulus-pallet-dmp-queue`) indicate that this is also what the code does, hence my wording in the title. 
So for me, this part of the discussion mostly results in the point of whether the `XcmExcutor` belongs to this layer and if `XCM` can be abstracted over it, which currently is not possible as the queues are aware of the version (`ExecuteXcm`, error handling in [dmp-queue](https://github.com/paritytech/cumulus/blob/ccf18114326287aedb6906b42c74ed4829c99022/pallets/dmp-queue/src/lib.rs#L243-L251) and `SendXcm`). 

> In addition, the queues themselves already accepts and parses `VersionedXcm`s when handling messages, and the XCM executor is completely configurable by the implementer, [...]

Using `VersionedXcm`s does not help the parachain here, as the `ExecuteXcm` is version aware and the parachain never "sees" the versioned XCM. 

If this is intended, and everybody just must use the latest version of XCM, this is of course fine for me. 

But if not (citing myself from above to make my use-case clear)
> I thought XCM was supposed to have overlapping compatibility between versions and hence chain A could be using XCM V4 while I am already on XCM V5. Otherwise, chains will be dependent on always being in sync with respect to the XCM version they are using when they want to communicate. If this is not the case, the point is of course negligible.

then I would appreciate a hint on how this is possible with the current approach?

mustermeiszer:
@KiChjang and @gavofyork any more info on the above?

gavofyork:
I see little sense in exposing `VersionedXcm` any more than it already is. No runtime is ever expected to include multiple XCM executors, since they are tightly bound to the APIs in their Polkadot version. It would also be a huge pain to integrate multiple versions into the same runtime, with many of the structures and types needing duplicates.

Without have multiple XCM executors, the only things you could reasonably handle multiple XCM versions is by either rejecting them or translating them into the native version of your runtime's XCM executor. This is already what happens.

mustermeiszer:
Ah, thanks. This was enlightening. TBH I have overlooked the `TryFrom` implementation for every XCM version. Makes sense to translate the instructions from version to version. 

I will close this here. Thanks for your patience and time!

PR #868, opened 2021-12-18
Use Polkadot Constants instead of Hardcoding

### LINKED ISSUE #860: Use Polkadot Constants instead of Hardcoding
Now that https://github.com/paritytech/polkadot/pull/4456 is in, we could use the constants crates instead of hard coding the values e.g. [here](https://github.com/paritytech/cumulus/blob/ccf18114326287aedb6906b42c74ed4829c99022/polkadot-parachains/statemine/src/constants.rs#L19-L30).

Doordashcon:
This PR resolves #860
polkadot companion: [paritytech/polkadot#4561](https://github.com/paritytech/polkadot/pull/4561)

polkadot address: 12zsKEDVcHpKEWb99iFt3xrTCQQXZMu477nJQsTBBrof5k2h


Doordashcon:
@apopiak please review

apopiak (commented):
Thanks for your contribution!

I don't agree with your particular approach here:
We want to keep the `constants.rs` modules in the runtimes, we just want to define the values by using the relay chain constants instead of them being hard-coded.
So I would ask you to re-add the removed modules.
Neither the client code (`chain_spec.rs`) nor the runtime `lib.rs` files should need to change.

polkadot-parachains/statemine/src/constants.rs
```diff
@@ -14,15 +14,16 @@
...
-	pub const CENTS: Balance = UNITS / 30_000;
-	pub const GRAND: Balance = CENTS * 100_000;
-	pub const MILLICENTS: Balance = CENTS / 1_000;
+	pub const UNITS: Balance = constants::currency::UNITS;
+	pub const CENTS: Balance = constants::currency::CENTS;
+	pub const GRAND: Balance = constants::currency::GRAND;
+	pub const MILLICENTS: Balance = constants::currency::MILLICENTS;
 
 	pub const fn deposit(items: u32, bytes: u32) -> Balance {
 		// map to 1/10 of what the kusama relay chain charges (v9020)
```
apopiak, line 29: would be good to use the kusama `deposit` function here (and divide by ten)
(and use the corresponding `deposit` functions in other runtimes respectively)

apopiak (commented):
Looking good, just the deposit missing :+1:

polkadot-parachains/statemine/src/constants.rs
Doordashcon: That should do it :)

apopiak:
Cool, ready to be merged, I'd say.
@bkchr @KiChjang can we get a review?

polkadot-parachains/statemine/src/constants.rs
```diff
@@ -14,19 +14,20 @@
 // limitations under the License.
 
 pub mod currency {
+	use kusama_runtime_constants as constants;
 	use node_primitives::Balance;
 
 	/// The existential deposit. Set to 1/10 of its parent Relay Chain (v9020).
-	pub const EXISTENTIAL_DEPOSIT: Balance = CENTS / 10;
+	pub const EXISTENTIAL_DEPOSIT: Balance = constants::currency::EXISTENTIAL_DEPOSIT / 10;
```
KiChjang, line 21: The `EXISTENTIAL_DEPOSIT` is not directly equal to `constants::currency::EXISTENTIAL_DEPOSIT`, but a tenth of it? How does this make any sense, especially given that the previous code has `CENTS / 10`?

polkadot-parachains/statemint/src/constants.rs
```diff
@@ -15,18 +15,19 @@
 
 pub mod currency {
 	use node_primitives::Balance;
+	use polkadot_runtime_constants as constants;
 
 	/// The existential deposit. Set to 1/10 of its parent Relay Chain (v9010).
-	pub const EXISTENTIAL_DEPOSIT: Balance = 10 * CENTS;
+	pub const EXISTENTIAL_DEPOSIT: Balance = constants::currency::EXISTENTIAL_DEPOSIT / 10;
```
KiChjang, line 21: Same comment here, why is the `EXISTENTIAL_DEPOSIT` divided by 10?

polkadot-parachains/westmint/src/constants.rs
```diff
@@ -15,18 +15,19 @@
 
 pub mod currency {
 	use node_primitives::Balance;
+	use westend_runtime_constants as constants;
 
 	/// The existential deposit. Set to 1/10 of its parent Relay Chain.
-	pub const EXISTENTIAL_DEPOSIT: Balance = 100 * MILLICENTS;
+	pub const EXISTENTIAL_DEPOSIT: Balance = constants::currency::EXISTENTIAL_DEPOSIT / 10;
```
KiChjang, line 21: Same here.

polkadot-parachains/statemine/src/constants.rs
apopiak: If you check the [existential deposit in Kusama](https://github.com/paritytech/polkadot/blob/12ffb5683b05f9fe7ae0b4c0cd1c329e2e125942/runtime/kusama/constants/src/lib.rs#L23) it is `1 CENT`. As the comment above the code explains, we want the ED on Statemine to be 1/10th of the relay chain deposit, so that's why it originally was hard-coded to `CENTS / 10` and now uses the `kusama_runtime_constants` divided by 10.

The reason for wanting it to be 1/10th is that storage on the relay is more expensive to the overall network than on a parachain. (Which is part of the motivation for introducing Statemint/e as an assets parachain.)

polkadot-parachains/statemine/src/constants.rs
```diff
@@ -14,19 +14,20 @@
 // limitations under the License.
 
 pub mod currency {
+	use kusama_runtime_constants as constants;
 	use node_primitives::Balance;
 
 	/// The existential deposit. Set to 1/10 of its parent Relay Chain (v9020).
```
apopiak, line 20: @Doordashcon Just noticed that we'll want to remove the reference to a specific runtime version.
```suggestion
	/// The existential deposit. Set to 1/10 of its parent Relay Chain.
```

shawntabrizi:
@Doordashcon i think you need to update your rust compiler to make the rustfmt match up here

gilescope (approved):
LGTM

bkchr (approved):

gilescope:
bot merge

Doordashcon:
Got a notification by @gilescope about a tip

Issue #881, opened 2021-12-30
Support returning custom header on the client side as well


bkchr:
https://github.com/paritytech/cumulus/pull/825 brought support for returning a custom header as part of the validation phase on the relay chain. This can be used for example to switch to a different chain. The problem now is that we also send the expected header as part of the candidate descriptor when building the collation: https://github.com/paritytech/cumulus/blob/master/client/collator/src/lib.rs#L178
The relay chain checks that the candidate descriptor header matches the one being returned by the validation function. This means we now also need to ensure that if a custom header should be set, we also use this header when announcing the collation to the relay chain validators. This should be done in the following way:

1. Bump the version of the `CollationInfo` runtime api to `2`.
2. Change the `collect_collation_info` to also take the header of the current block and change the return value to also return the head data.
3. In the runtime we then check if a custom header should be set and if that is true, we return this custom header or we return the header that we got passed in through the runtime api function.
4. Ensure that the code on the client side is backwards compatible. This can be done by checking the current runtime api version and calling the appropriate "old" function. 

CC @NachoPal

PR #882, opened 2022-01-02
Fix solo to para custom header return

bkchr:
https://github.com/paritytech/cumulus/pull/825 brought support for returning a custom header as part of the validation phase. This was tested only in isolation, aka not in a integration test. While testing it, we found out that we missed to also handle the client side correctly when there is a custom header to be returned. To be more precise, we need to return the new custom header also as part of the collation because otherwise the validation fails on the relay chain. This pr fixes this by adding an integration test and then it follows on what was outlined in https://github.com/paritytech/cumulus/pull/881

Fixes: https://github.com/paritytech/cumulus/issues/881

### DIFF
--- a/client/collator/src/lib.rs
+++ b/client/collator/src/lib.rs
@@ -18,11 +18,12 @@
 
 use cumulus_client_network::WaitToAnnounce;
 use cumulus_primitives_core::{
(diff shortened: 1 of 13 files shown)


skunert (approved):
Looks Good!

PR #835, opened 2021-12-06
Introduce interface for relay chain interaction

skunert:
In this PR, I introduce an interface that exposes methods for interaction with the relay chain, as discussed in #545.
These changes are only refactoring and have no impact on functionality.
Outline of the changes:
- Introduce trait `RelayChainInterface`
- Implement `RelayChainInterface` for `RelayChainLocal`, a status-quo implementation that uses a full node in the same process
- replace all occurrences of polkadot client and backend with `RelayChainLocal`

PR #887, opened 2022-01-05
Fix `rococo-parachain-runtime` build

skunert:
Command `cargo build -p rococo-parachain-runtime` broke after #835.

I found that we are missing some entries in `polkadot-parachains/rococo-parachain/Cargo.toml`, I am not sure why it worked before however.

Additionally, I add `client/relay-chain-local` to the workspace.


bkchr (approved):

bkchr:
> I am not sure why it worked before however.

It worked because you compiled the workspace. If you enable a feature in the workspace, it is enabled everywhere automatically.

PR #888, opened 2022-01-05
Ensure bad datastream cannot cause problems2

gilescope:
Is `decode_and_advance_with_depth_limit` guaranteed to always consume some of the input? I feel like we should do the same as #701 and add a guard and fail if the buffer was not reduced in size to be on the safe side (or failing that make it clear in the doc of that function that it must uphold that invariant).

(`if let` has been turned into a `match` and guard `if remaining_fragments.len() < last_remaining_fragments.len()` added.)

### DIFF
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -520,7 +520,7 @@ impl<T: Config> Pallet<T> {
 									let e =
 										Event::OverweightEnqueued(sender, sent_at, index, required);
 									Self::deposit_event(e);
-								}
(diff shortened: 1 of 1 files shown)


bkchr:
Yes it is guaranteed to always consume the input, otherwise it would not have decoded anything.

bkchr:
Maybe you could just read the docs of scale codec?

bkchr:
#701 wasn't correct, see https://github.com/paritytech/cumulus/pull/889

This pr is also not required, because we consume the input here and if there is an error we handle that properly.

PR #889, opened 2022-01-05
xcmp-queue: Fix handling of encoded blobs

bkchr:
With #701 we tried to fix some infinite loop related to encoded blobs, however that lead actually to
not being able to process encoded blobs at all. The reason for this is that `decode_all` doesn't
consume the given input. The point of this function is that it returns an error if the data couldn't
be decoded or there is still data left. However, this means that the check
`remaining_fragments.len() < last_remaining_fragments.len()` would always fail.

We remove the while loop, because we decode the entire fragment anyway or it fails. Aka, we don't
need to loop here. Next we remove the broken check and we directly reset the
`remaining_fragments` (because `decode_all` doesn't consume anything).

### DIFF
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -33,7 +33,7 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
-use codec::{Decode, DecodeAll, DecodeLimit, Encode};
+use codec::{Decode, DecodeLimit, Encode};
 use cumulus_primitives_core::{
 	relay_chain::BlockNumber as RelayBlockNumber, ChannelStatus, GetChannelInfo, MessageSendError,
 	ParaId, XcmpMessageFormat, XcmpMessageHandler, XcmpMessageSource,
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -536,28 +536,28 @@ impl<T: Config> Pallet<T> {
 				}
 			},
 			XcmpMessageFormat::ConcatenatedEncodedBlob => {
-				while !remaining_fragments.is_empty() {
+				if !remaining_fragments.is_empty() {
+					// `decode_all` doesn't consume the input. Thus, we directly reset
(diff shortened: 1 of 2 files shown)


gilescope:
Can we do a similar thing with the XcmpMessageFormat::ConcatenatedVersionedXcm case?

bkchr:
> Can we do a similar thing with the XcmpMessageFormat::ConcatenatedVersionedXcm case?

This is not required.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -538,26 +538,24 @@ impl<T: Config> Pallet<T> {
...
-									// Message invalid; don't attempt to retry
-								},
-							}
-						},
-						_ => {
-							debug_assert!(false, "Invalid incoming blob message data");
-							remaining_fragments = &b""[..];
-						},
+
+					if let Ok(blob) = <Vec<u8>>::decode(&mut remaining_fragments) {
```
KiChjang, line 542: Do we not check for `remaining_fragments.len() < last_remaining_fragments.len()` anymore here?

bkchr: No. We will for sure consume something and that will then either return `Ok()` or `Err()`. If there is an error we abort.

`remaining_fragments.len() < last_remaining_fragments.len()` this check was also completely wrong in the old code and was always returning `false.

KiChjang: Hold on, the code was always returning false previously because we used decode_all, which doesn't advance the cursor in remaining_fragments. If we do check it now, then we should expect it to return true most of the time.

The only time where I can see that it would not return true is if remaining_fragments is empty, and we keep on trying to decode it. I think we could instead make a debug_assert here to ensure that we always decode something.

bkchr: > not return true is if remaining_fragments is empty

`decode` would return an error when `remaining_fragments` is empty, because that is no valid encoded vec.

KiChjang (approved):

PR #893, opened 2022-01-11
Enable asset teleportation on Statemint

KiChjang:


gilescope:
https://github.com/paritytech/cumulus/pull/912 seems related

KiChjang:
@gilescope You're right, we'll need both #912 and this PR in order for the `teleport_asset` extrinsic to work properly.

joepetrowski (approved):
Given that this has no effect until Relay Chain governance decides it's OK to send DOT to Statemint, then it's OK for Statemint to send it back as well.

joepetrowski:
Actually, we probably need https://github.com/paritytech/polkadot/issues/4462 in the Relay Chain first.

hbulgarini:
@KiChjang can we close this PR? it seems to be doing the same as #1057 which contains related work for enabling XCM transfers and lift asset creation restrictions.

KiChjang:
Yeah, let's close this.

PR #895, opened 2022-01-12
Companion for substrate#10632

koushiro:
Companion for paritytech/substrate#10632


polkadot-parachains/src/cli.rs
```diff
@@ -52,69 +52,70 @@ pub enum Subcommand {
...
 	/// The custom benchmark subcommmand benchmarking runtime pallets.
-	#[structopt(name = "benchmark", about = "Benchmark runtime pallets.")]
+	#[clap(name = "benchmark", about = "Benchmark runtime pallets.")]
 	Benchmark(frame_benchmarking_cli::BenchmarkCmd),
 
 	/// Try some testing command against a specified runtime state.
 	TryRuntime(try_runtime_cli::TryRuntimeCmd),
 
 	/// Key management CLI utilities
+	#[clap(subcommand)]
```
bkchr, line 62: Still don't get why we need this here and not for the others.

bkchr: What is the error when we don't do this? Does it fail at compile time?

koushiro: yeah, it will be fail at compile time

bkchr (approved):

KiChjang:
bot merge

PR #896, opened 2022-01-12
Add the ability to suspend or resume XCM execution on the XCMP queue

### LINKED ISSUE #891: Suspend xcmp/dmp queue
Sometimes a multi-block migration is required and we need to suspend all the operations for some pallets depends on the migration. We can easily reject all user transactions by updating the base call filter, however we will also need a way to suspend executions delivered via XCM.

We are building a pallet for that but encountered an issue that we have no way do disable execution within on `on_idle` hook.

https://github.com/AcalaNetwork/Acala/pull/1737#issuecomment-1006609731

It will be good if there is public API to suspend & resume the queue, and extrinsic to send suspend/resume signal to other chains.

KiChjang:
Fixes #891.


shawntabrizi (approved):
looks super straight forward

shawntabrizi:
probably should be "needs audit"

zqhxuyuan:
maybe dmp-queue also need this feature? would that another PR or just here?

pallets/xcmp-queue/src/lib.rs
```diff
@@ -220,6 +249,10 @@ pub mod pallet {
 	/// available free overweight index.
 	#[pallet::storage]
 	pub(super) type OverweightCount<T: Config> = StorageValue<_, OverweightIndex, ValueQuery>;
+
+	/// Whether or not the XCMP queue is active in executing incoming XCMs or not.
+	#[pallet::storage]
+	pub(super) type QueueActive<T: Config> = StorageValue<_, bool, ValueQuery>;
```
xlc, line 255: this will be default to false so we either need a migration + genesis config to make this true, or make it `PauseQueue` or something that default value means standard behaviour.

KiChjang: That was what I was wondering as well, glad that you caught it!

pallets/dmp-queue/src/lib.rs
```diff
@@ -154,6 +161,32 @@ pub mod pallet {
 			Self::deposit_event(Event::OverweightServiced(index, used));
 			Ok(Some(used.saturating_add(1_000_000)).into())
 		}
+
+		/// Suspends all XCM executions for the XCMP queue, regardless of the sender's origin.
```
apopiak, line 165:
```suggestion
		/// Suspends all XCM executions for the DMP queue, regardless of the sender's origin.
```

pallets/dmp-queue/src/lib.rs
```diff
@@ -109,6 +112,10 @@ pub mod pallet {
 	pub(super) type Overweight<T> =
 		StorageMap<_, Blake2_128Concat, OverweightIndex, (RelayBlockNumber, Vec<u8>), OptionQuery>;
 
+	/// Whether or not the XCMP queue is suspended from executing incoming XCMs or not.
```
apopiak, line 115:
```suggestion
	/// Whether or not the DMP queue is suspended from executing incoming XCMs or not.
```

pallets/dmp-queue/src/lib.rs
```diff
@@ -154,6 +161,32 @@ pub mod pallet {
...
+		#[pallet::weight(T::DbWeight::get().writes(1))]
+		pub fn suspend_xcm_execution(origin: OriginFor<T>) -> DispatchResult {
+			T::ControllerOrigin::ensure_origin(origin)?;
+
+			QueueSuspended::<T>::put(true);
+
+			Ok(())
+		}
+
+		/// Resumes all XCM executions for the XCMP queue.
```
apopiak, line 177:
```suggestion
		/// Resumes all XCM executions for the DMP queue.
```

apopiak (commented):
copy-pasta fixes

polkadot-parachains/statemine/src/lib.rs
```diff
@@ -604,12 +604,14 @@ impl cumulus_pallet_xcmp_queue::Config for Runtime {
 	type ChannelInfo = ParachainSystem;
 	type VersionWrapper = PolkadotXcm;
 	type ExecuteOverweightOrigin = EnsureRoot<AccountId>;
+	type ControllerOrigin = EnsureRoot<AccountId>;
```
joepetrowski, line 607: The problem with this is that `Root` is actually on another chain (Kusama). So in order to`resume_xcm_execution`, it would need to ... process an XCM.

So I think `service_{xcmp,dmp}_queue` needs a bypass, where messages with `ControllerOrigin` can still be executed.

I also think this should be `EnsureOneOf<EnsureRoot<AccountId>, EnsureXcm<IsMajorityOfBody<KsmLocation, ExecutiveBody>>>` but open to discussion on this.

xlc: For chains without their own governance body (Statemine/Statemint), this feature is actually a bit dangerous because if XCM is disabled, it will not be able to process the XCM to resume it...

KiChjang: Oh wow ok, if that's the case then we have a bit of a problem in our hands, because I briefly considered whether it made sense to just suspend ALL inbound channels by looping through them and modifying their state, instead of suspending the entire queue in the way that the PR currently handles it.

With such a requirement, it does look to me that the best way to handle it is to suspend all inbound channels, which could be quite computationally heavy if there are a lot of them. The other thing is, when we try to resume all inbound channels afterwards, we would also accidentally resume a channel that was suspended for other reasons, but I think this can be solved by creating more channel states.

KiChjang: So I figured that this is not as important on the XCMP queue, since for our use case, it would be the relay chain that's sending us an XCM, and it would use the DMP queue. The problem here is that we don't really get to know whether the XCM came from an executive body; all we know is that it came from the relay chain. This would mean that without the ability to discern the origin more precisely, we should never suspend the DMP queue as it may contain important root calls.

KiChjang: A corollary for not being able to know whether or not the XCM came from an executive body is that adding an `EnsureXcm<IsMajorityOfBody<...>>` isn't that useful, as the origin is always set to the sibling parachain. Currently the bypass works by having the `OriginConverter` convert _all_ incoming XCMs originating from a specified parachain to be the `Root` origin.

Note that the root conversion is only used to determine whether or not we bypass the suspended queue; the `Transact` instruction is unaffected. Perhaps I should rename `OriginConverter` to `ControllerConverter` to make the intention clearer.

gavofyork (approved):

polkadot-parachains/statemine/src/lib.rs
gavofyork: While I reckon that this is not a bad feature to have, we could also just do all of this with a barrier.

PR #900, opened 2022-01-13
Add collator selection tool tips

gilescope:
Document calls to collator-selection

(First line of doc comments show as tool tips in polkadot js.)


pallets/collator-selection/src/lib.rs
```diff
@@ -281,6 +281,7 @@ pub mod pallet {
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		#[pallet::weight(T::WeightInfo::set_invulnerables(new.len() as u32))]
+		/// Sets the list of invulnerable (fixed) collators
```
bkchr, line 284:
```suggestion
		/// Sets the list of invulnerable (fixed) collators
		#[pallet::weight(T::WeightInfo::set_invulnerables(new.len() as u32))]
```

pallets/collator-selection/src/lib.rs
```diff
@@ -280,6 +280,7 @@ pub mod pallet {
 
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
+		/// Sets the list of invulnerable (fixed) collators
```
joepetrowski, line 283:
```suggestion
		/// Set the list of invulnerable (fixed) collators.
```

pallets/collator-selection/src/lib.rs
```diff
@@ -312,6 +316,8 @@ pub mod pallet {
 			Ok(().into())
 		}
 
+		/// Set the amount to deposit for becoming a collator.
+		/// When a collator calls `leave_intent` they receive the deposit back.
```
joepetrowski, line 320: I think this should be on the `CandidacyBond` storage item, not the function that sets it.

pallets/collator-selection/src/lib.rs
```diff
@@ -323,6 +329,10 @@ pub mod pallet {
 			Ok(().into())
 		}
 
+		/// Ask for this account to become a collator candidate.
+		/// The account must already have called session set_keys (linking the account to a running validator).
+		/// The candidacy bond will be reserved from this account.
+		/// (This call is not available to invulnerable collators.)
```
joepetrowski, line 335:
```suggestion
		/// Register this account as a collator candidate. The account must (a) already have
		/// registered session keys and (b) be able to reserve the `CandidacyBond`.
		///
		/// This call is not available to `Invulnerable` collators.
```

pallets/collator-selection/src/lib.rs
```diff
@@ -362,6 +372,11 @@ pub mod pallet {
 			Ok(Some(T::WeightInfo::register_as_candidate(current_count as u32)).into())
 		}
 
+		/// Notifies that the collator attached to this account will be leaving
+		/// once the session has finished. The candidacy bond is refunded to the account immediately.
+		/// This call will fail if the number of candidate collators would drop below
+		/// the minimun candidates.
+		/// (This call is not available to invulnerable collators.)
```
joepetrowski, line 379:
```suggestion
		/// Deregister `origin` as a collator candidate. Note that the collator can only leave on
		/// session change. The `CandidacyBond` will be unreserved immediately.
		///
		/// This call will fail if the total number of candidates would drop below `MinCandidates`.
		///
		/// This call is not available to `Invulnerable` collators.
```

joepetrowski (approved):

PR #901, opened 2022-01-14
Companion for paritytech/polkadot#4712

KiChjang:
Companion for paritytech/polkadot#4712.


polkadot-parachains/shell/src/lib.rs
```diff
@@ -177,6 +177,7 @@ parameter_types! {
 	pub const RococoLocation: MultiLocation = MultiLocation::parent();
 	pub const RococoNetwork: NetworkId = NetworkId::Polkadot;
 	pub Ancestry: MultiLocation = Parachain(ParachainInfo::parachain_id().into()).into();
+	pub ParentAccount: AccountId = PalletId(*b"pa/shell").into_account();
```
shawntabrizi, line 180: what is the diff between `p-` and `pa/` as a prefix here?

polkadot-parachains/westmint/src/lib.rs
```diff
@@ -434,14 +434,15 @@ parameter_types! {
 	pub AssetsPalletLocation: MultiLocation =
 		PalletInstance(<Assets as PalletInfoAccess>::index() as u8).into();
 	pub CheckingAccount: AccountId = PolkadotXcm::check_account();
+	pub ParentAccount: AccountId = PalletId(*b"westend/").into_account();
```
shawntabrizi, line 437: and this one has no prefix?

I think you should be following a standard here

KiChjang: I need to have exactly 8 characters in `PalletId`, otherwise compilation fails and complains, so I have to make up some prefixes and suffixes.

Is there another ID type that I can use where I don't have the strict 8-character limit?

shawntabrizi: look at other uses of `PalletId`. We prefix anything with parity as `py/` and then use some unique suffix.

`py` stands for parity, whereas probably some of these networks we should not directly connect with Parity since they are not owned or controlled by Parity.

As mentioned in the other chat, probably this is important enough to engage a consensus, and not just willy-nilly fill in something here that compiles.

This will be a permanent part of the network after you set it.

shawntabrizi: Even using `PalletId` here might not be the right thing to do. Maybe you should even create your own struct which is `ParachainId` and allow more bytes to be used. All of these questions are not exactly clear, and so you need to engage the rest of the team.

KiChjang: I've asked Gav before and he said that the hash of anything sensible is good enough. I think what would be good is to not use `PalletId` and instead use something that's more permissive here, and put in the name of the parent network to indicate the parent account, e.g. on Westmint, we'll have `Id("Westend")` and on Statemine, we'd have `Id("Kusama")`.

polkadot-parachains/statemint/src/lib.rs
```diff
@@ -449,14 +449,17 @@ parameter_types! {
 	pub AssetsPalletLocation: MultiLocation =
 		PalletInstance(<Assets as PalletInfoAccess>::index() as u8).into();
 	pub CheckingAccount: AccountId = PolkadotXcm::check_account();
+	pub ParentAccount: AccountId = b"Parent"
+		.using_encoded(|b| AccountId::decode(&mut TrailingZeroInput(b)))
+		.expect("infinite length input; no invalid inputs for type; qed");
```
KiChjang, line 454: Wonder if we can put this in cumulus or polkadot parachain primitives, make it a function and call it here instead.

polkadot-parachains/statemint/src/lib.rs
```diff
@@ -449,14 +449,17 @@ parameter_types! {
...
+	pub ParentAccount: AccountId = b"Parent"
+		.using_encoded(|b| AccountId::decode(&mut TrailingZeroInput::new(b)))
+		.expect("infinite length input; no invalid inputs for type; qed");
 }
 
 /// Type for specifying how a `MultiLocation` can be converted into an `AccountId`. This is used
 /// when determining ownership of accounts for asset transacting and when attempting to use XCM
 /// `Transact` in order to determine the dispatch Origin.
 pub type LocationToAccountId = (
 	// The parent (Relay-chain) origin converts to the default `AccountId`.
```
bkchr, line 461: Docs are outdated.

polkadot-parachains/statemint/src/lib.rs
bkchr: Why not just having `ParentIs` having doing this internally? (The type should then be renamed) But if someone needs a special parent account they can just create their own type that implements `Convert`

KiChjang: Yeah, makes more sense that way.

bkchr (approved):

Issue #902, opened 2022-01-14
Ensure everything is merged back to master from statemint-v6.0.1 release


gilescope:
https://github.com/paritytech/cumulus/commits/release-statemint-v6.0.1 (after 24th nov should already have been upstreamed)

There's some changes that we want to make sure get back to master:

- [x] weight updates
- [x] https://github.com/paritytech/cumulus/pull/817/files
- [x] https://github.com/paritytech/cumulus/pull/794
- [x] check is in master https://github.com/paritytech/cumulus/pull/743
- [x] check the transaction_version: https://github.com/paritytech/cumulus/pull/748
- [x] https://github.com/paritytech/cumulus/pull/753

(the open PRs referenced here also include all the ones from the https://github.com/paritytech/cumulus/commits/release-statemine-v6.0.1 as well)

gilescope:
I think we can close this issue now that everything has been merged back.

PR #904, opened 2022-01-14
Add safe_xcm_version to Genesis Specs (#743)

gilescope:
This merges https://github.com/paritytech/cumulus/pull/743 back to master.
( part of https://github.com/paritytech/cumulus/issues/902 - merging the released changes back to master)


gilescope:
Should I extend this change to include the rococo and the parachain template? Feels like the right thing to do to me...

polkadot-parachains/src/chain_spec.rs
```diff
@@ -34,6 +34,9 @@ pub type ShellChainSpec = sc_service::GenericChainSpec<shell_runtime::GenesisCon
 pub type SeedlingChainSpec =
 	sc_service::GenericChainSpec<seedling_runtime::GenesisConfig, Extensions>;
 
+/// The default XCM version to set in genesis config.
+const SAFE_XCM_VERSION: u32 = 2;
```
bkchr, line 38: This makes no real sense. The xcm crate already exposes the latest `XCM_VERSION` and that should be used here.

apopiak: It exposes the *latest* version, which means that the xcm version in the chain spec would change when updating Polkadot/XCM. I don't think that's what we want. I think we want to use the known-good version and then update conservatively/in our own time. (i.e. when v3 is published, we want to default to v2 for the safe fallback version until we decide to upgrade).

bkchr: Why? These are development chain specs and why do you not want to use the latest version there? Anything else doesn't make really sense, because for development you don't really care what is the latest "commonly" deployed version.

apopiak: Mmmh, I would've assumed on a fresh checkout you would want the chain specs to be fairly close to what's deployed.
Will impact any integration tests for example.

gilescope: Where do / should the prod chain specs live? I would like to think that the prod chain specs would still be stored in master and then the release branch have the smallest delta possible.

apopiak: the prod chain specs are in `polkadot-parachains/res`, so e.g. `polkadot-parachains/res/statemint.json`

bkchr: > Mmmh, I would've assumed on a fresh checkout you would want the chain specs to be fairly close to what's deployed. Will impact any integration tests for example. Edit: Could see this going either way shrug

How should that work? If you add some new pallet or whatever, the chain spec may changes. There is tons of reasons why they could differ. 

For tests you probably want to send explicit XCM versions anyway. So, having cross chain transfers using V2 and V3. Maybe later with V4. Etc.

bkchr (approved):

gilescope:
bot merge

PR #906, opened 2022-01-14
Make it clear what happens when bond increases

gilescope:
Currently existing collator bond holders don't get grandfathered (unless removing them would take them below the min threshold).

Additional test to make this edge case clear.


gilescope:
hmm ignore this - the validator is set to 4 always so this isn't testing what I had hoped - I think 3 is just getting kicked due to not producing blocks.

Issue #910, opened 2022-01-18
Remove Stale `on_runtime_upgrade` Companion


joepetrowski:
See: https://github.com/paritytech/substrate/pull/10650#discussion_r786932021

PR #911, opened 2022-01-18
Add the uniques migration for statemine

kianenigma:
See: https://github.com/paritytech/substrate/pull/10650#discussion_r786932021

fixes https://github.com/paritytech/cumulus/issues/910

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,20 +14,11 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.16.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e61f2b7f93d2c7d2b08263acaa4a363b3e276806c68af6134c44f523bf1aacd"
-dependencies = [
- "gimli 0.25.0",
-]
-
-[[package]]
-name = "addr2line"
 version = "0.17.0"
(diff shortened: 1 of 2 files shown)


shawntabrizi (approved):

kianenigma:
Thinking out loud: you might say this is now extra effort that we've created by removing migrations from inside the pallet, but to me this is actually the point: I want more people (perhaps forcefully) thinking about what migrations they need to write, and in which chains they need to be applied, and how to label their PRs so that everyone is notified etc.

hamidra:
> Thinking out loud: you might say this is now extra effort that we've created by removing migrations from inside the pallet, but to me this is actually the point: I want more people (perhaps forcefully) thinking about what migrations they need to write, and in which chains they need to be applied, and how to label their PRs so that everyone is notified etc.

Also might be worth to update this [docs](https://docs.substrate.io/how-to-guides/v3/storage-migrations/basics/) to mention the best practices. I guess many people will start from there when writing migrations!

kianenigma:
> 

Good idea: https://github.com/substrate-developer-hub/substrate-docs/issues/780

kianenigma:
Should be green now. @bkchr feel free to merge when ready. Note how I updated substrate and polkadot in the commit messages as well.

PR #922, opened 2022-01-19
Change Rococo Para id from 1000 to 2000

NachoPal:
Right now there is not an easy way of running a Community Parachain collator type (id >= 2000 and not 1000) without generating the `chain-spec-raw.json`.

Rococo parachain is set to 1000, but we don't really have a "Rockmine" Common Good parachain (and we could use Westmint for that purpose if needed).

I'd like to change the Rococo Parachain id to 2000, so then I could easily tests scenarios such as opening HRMP channels between Common Good parachain & Community parachain or any kind of interaction between them.

The Rococo Parachain collator could be run just with the `--chain=""` command attribute

### DIFF
--- a/polkadot-parachains/src/chain_spec.rs
+++ b/polkadot-parachains/src/chain_spec.rs
@@ -91,7 +91,7 @@ pub fn get_chain_spec() -> ChainSpec {
 					get_account_id_from_seed::<sr25519::Public>("Eve//stash"),
 					get_account_id_from_seed::<sr25519::Public>("Ferdie//stash"),
 				],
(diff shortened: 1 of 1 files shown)


NachoPal:
I think we can forget about changing the `rococo-parachain` `paraId`. I realised that we'll need more than one Community Parachain for testing new scenarios. For example, a Reserved Transfer between two Community Parachains via Statemint. Having only one extra Parachain to launch with `paraId != 1000` does not really solve the issue. It seems unavoidable the need of building the chain_specs and change their `paraId` manually.

Closing the PR

PR #927, opened 2022-01-21
pov-recovery: Enable pov-recovery as well on full nodes

bkchr:
Pov recovery was before only enabled for collators. The reason behind this was prevention of spam of
the relay chain for block recovery. However, recent events has shown that this wasn't a good idea.
So, this pr enables pov-recover for normal full nodes as well, but with a much higher delay before
trying to recover a block. This means that full nodes will wait in minimum 2.5 minutes and in
maximum 5 minutes before recovering a block. This should give collators in "normal mode" enough time
to recover a block (they wait in maximum 6 seconds after they have seen a new candidate in the relay
chain) before recovering a block. So, we should hopefully not spam the relay chain.


skunert:
Looks good. Only thing is that the comment here should be adjusted to reflect this change: https://github.com/paritytech/cumulus/blob/master/client/pov-recovery/src/lib.rs#L27

skunert (approved):

PR #933, opened 2022-01-25
fix: remove duplicate event of pallet_balances

wangminqi:
https://github.com/paritytech/substrate/blob/master/frame/support/src/traits/tokens/currency.rs#L158
https://github.com/paritytech/substrate/blob/master/frame/balances/src/lib.rs#L1647

Since 0.9.12 of substrate, there is no need for a separate of balances event in runtime.


apopiak (commented):
Thank you for your contribution!

Looks correct to me :+1:  

related to https://github.com/paritytech/substrate/pull/9425 and https://github.com/paritytech/polkadot/pull/4180

shawntabrizi (approved):

bkchr (approved):

PR #934, opened 2022-01-25
Add production profile

chevdor:
See also https://github.com/paritytech/polkadot/pull/4778

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,4 +40,11 @@ members = [
 ]
 
 [profile.release]
+# Polkadot runtime requires unwinding.
 panic = "unwind"
+
+[profile.production]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -40,4 +40,11 @@ members = [
 ]
 
 [profile.release]
+# Polkadot runtime requires unwinding.
```
bkchr, line 43:
```suggestion
```

bkchr (approved):

Cargo.toml
chevdor: I kept it so we know **why** but it does not matter much.

chevdor:
bot merge

PR #936, opened 2022-01-25
Rococo Parachain XCM Adjustments

apopiak:
This PR:
+ adjusts the `AssetsFrom` filter to allow for suffixes longer than 1 `Junction` to be compatible with the new `PalletInstance` prefix introduced by https://github.com/paritytech/cumulus/pull/831
  + and adds some XCM tracing.
+ sets Statemint (Parachain 1000) as a trusted reserve for relay chain tokens.

Questions coming out of this:
+ Do we want to upstream this kind of matching (allowing more than 1 Junction suffix) into the XCM modules? (e.g. a `fn match_prefix(self, prefix: MultiLocation)` on `MultiLocation`? --> Yes: https://github.com/paritytech/polkadot/pull/4835


polkadot-parachains/parachains-common/src/impls.rs
```diff
@@ -103,9 +103,16 @@ pub struct AssetsFrom<T>(PhantomData<T>);
...
 		let loc = T::get();
+		log::trace!(target: "xcm::AssetsFrom", "loc: {:?}, origin: {:?}", loc, origin);
 		&loc == origin &&
-			matches!(asset, MultiAsset { id: AssetId::Concrete(asset_loc), fun: Fungible(_a) }
-			if asset_loc.match_and_split(&loc).is_some())
+			match asset {
+				MultiAsset { id: AssetId::Concrete(asset_loc), fun: Fungible(_a) } =>
+					loc.parent_count() == asset_loc.parent_count() &&
+						loc.interior().iter().zip(asset_loc.interior().iter()).all(
+							|(prefix_junction, asset_junction)| prefix_junction == asset_junction,
```
gilescope, line 112: if asset_loc.interior() is longer, then I think std iter zip will ignore any extra length. Is this intended?

apopiak: Should be fixed now.

polkadot-parachains/parachains-common/src/impls.rs
```diff
@@ -98,14 +98,41 @@ where
...
+/// ```
+/// # use xcm::prelude::*;
+/// # use parachains_common::impls::match_prefix;
+/// let prefix =  MultiLocation::new(1, X1(Parachain(1234)));
+/// let loc =  MultiLocation::new(1, X2(Parachain(1234), GeneralIndex(5)));
+/// assert!(match_prefix(&prefix, &loc));
+/// let failing =  MultiLocation::new(1, Here);
+/// assert!(!match_prefix(&prefix, &failing));
+/// ```
+pub fn match_prefix(prefix: &MultiLocation, loc: &MultiLocation) -> bool {
```
KiChjang, line 114: This looks like it's a function that should be upstreamed to Polkadot.

apopiak: yep, agree, as mentioned in the description :-D

apopiak: https://github.com/paritytech/polkadot/pull/4827

polkadot-parachains/parachains-common/src/impls.rs
```diff
@@ -98,14 +98,33 @@ where
 	}
 }
 
+/// Tests `loc` whether it starts with `prefix`.
+/// Similar to `MultiLocation::match_and_split` but allows arbitrary suffixes (instead of just one
+/// junction).
+// TODO: Replace with version implemented on MultiLocation here:
+// https://github.com/paritytech/polkadot/pull/4827
```
apopiak, line 105: XCM v3 might take a while to be merged. I can just create an issue to track this.

KiChjang (approved):

joepetrowski:
@KiChjang @NachoPal : Is this still a change we want/need? Let's merge master or close it.

KiChjang:
IIRC, the discussion we had before was that Statemint/e shouldn't view the relay chain as a reserve for DOT/KSM as it can cause problems with reserves drainage via the mixed usage of asset teleportation and reserve transfers. Given so, we probably do not need to allow Statemint to recognize the relay chain as a reserve.

However, the `AssetsFrom` filter change still makes sense to me.

Issue #938, opened 2022-01-31
v7 release branch does't build


gilescope:
While the 9.16 branch builds the v7 does not. ( https://github.com/paritytech/cumulus/runs/5006506401?check_suite_focus=true )

I think we can get past the first build failure by a change like 
```
fn host_storage_root() -> Vec<u8> {
	vec[]
	with_externalities(|ext| ext.storage_root(StateVersion::V0))
}
```
in pallets/parachain-system/src/validate_block/implementation.rs

reverting b03e8bcf151bbf4c20a7d7faa782366fc3554d58 should fix this.

but then there seem to be some clap issues also:

```
error[E0277]: the trait bound `sc_cli::RunCmd: FromArgMatches` is not satisfied
   --> client/cli/src/lib.rs:126:9
    |
126 |     #[clap(flatten)]
    |            ^^^^^^^ the trait `FromArgMatches` is not implemented for `sc_cli::RunCmd`
```

(maybe the latest cut has brought in a new version of clap?)

UPDATE:
reverting
b03e8bcf151bbf4c20a7d7faa782366fc3554d58 ,
9c977d66a3a8eb9412004d1a75c1da76c96adb80 
and fd145766d4c2ecb7297c5d2a4a389b4c5b9b174e should fix the build. The latter two are reverted already in the cumulus 9.16 release branch, but not the first one - let me check if something was brought in to fix that.

I also think we need to include: 86f76c5619c64d1300315612695ad4b4fcd0f562 (as this was on the release branch of cumulus)

gilescope:
dep of https://github.com/paritytech/cumulus/issues/937

gilescope:
ah looks like master has gone from structopt to clap.

gilescope:
also if we cherry pick this commit: ec07b4e532369fff4548241c30371aece9b472e4 to revert https://github.com/paritytech/cumulus/pull/901 then we get something that builds again. Yay.

gilescope:
Fixed.

PR #939, opened 2022-02-01
Merge all commits from cumulus release branch

gilescope:
It turns out that all the commits in the `polkadot-v0.9.16` branch are needed. This PR adds them in to the parachains release branch also.

Fixes: https://github.com/paritytech/cumulus/issues/938


gilescope:
( 90e3eb6 may not be strictly necessary but it seems right to include it.)

gilescope:
will close this and redo with taking the same polkadot hash choice as basti's release branch.

gilescope:
(for the record we seem to be ok excluding 901)

PR #953, opened 2022-02-03
set collator to statemin version

gilescope:
This appears here I believe:

Screenshot(19)

So would be nice for statemine/statemint users to not get confused.


NachoPal:
Is it expected to be changed also in cumulus or only in our releases?

I can only say that the collator is not specific for the parachain release (v700) so I would follow whatever is used in cumulus.

chevdor:
It is not in sync actually on purpose as we want to disconnect to 2. There is no reason why the evolution of the runtime and the collator remains in sync. I would actually prefer NOT to have it "in sync" to avoid people assuming that the 2 are linked.

So I would prefer not to merge this change.

chevdor (commented):
For me this is a no 🙅

gilescope:
Ok, what if I change the base to merge into to [polkadot-v0.9.16](https://github.com/paritytech/cumulus/tree/polkadot-v0.9.16)

We have updated the collator so surely it should have a version bump if we are doing semver on it?

gilescope:
changing base isn't the right way to do this - will close this.

PR #957, opened 2022-02-04
ci: fix publish-runtimes for parachains containing - in name

coderobe:
rust outputs these with underscores, so `rococo-parachain` ends up as `rococo_parachain`, this breaks assumptions in the ci publish-runtimes step of release-02_create-draft, so we move it back to where it's expected


coderobe:
see https://github.com/paritytech/cumulus/actions/runs/1795113358 for an example of the failure this is addressing

chevdor (approved):

chevdor:
bot merge

PR #959, opened 2022-02-04
ci: fix publish-runtimes for parachains containing - in name (v2)

coderobe:
https://github.com/paritytech/cumulus/pull/957 did not work. let's try that again :sweat_smile:


coderobe:
i've tested the workflow locally with this change and we have a success :sparkles: 

this should be all good to merge :heavy_check_mark:

EgorPopelyaev (approved):

coderobe:
bot merge

Issue #545, opened 2021-07-21
Support running the relay chain node as an extra process


bkchr:
Currently every parachain node includes a relay chain node. This relay chain node is required to get certain information about the relay chain that are important for the parachain, for example to query what is the best parachain block currently etc. While this makes it relative easy to run a parachain node, it also brings some problems like the high compilation time as we need to compile the parachain node and the relay chain node. Another, a more bigger problem is that parachain developers are required to update their collators when there is a new relay chain release that requires a timely update, because a new host function is added or something in the client code of the relay chain is fixed. So, it would be nice to have the relay chain node running as an extra process. Collator operators would just run an extra relay chain node (that could maybe even shared between multiple parachain nodes, but this is no real initial requirement!) and can freely update the relay chain node. The relay chain node itself could maybe directly bring the functionality required by Cumulus to connect to it or we provide some sort of wrapper (probably the best way for the first implementation).

This feature would be implemented in the following order:
1. Refactor all usages of the polkadot client to have them behind some common trait or maybe multiple traits. So, we should not have any reference to `polkadot-service` or `polkadot-client` in any of the "low-level" functionality of Cumulus. It should only use these interfaces to talk to the relay chain.
2.  Write an implementation of these traits for the "in-node relay chain" so that we are back to on par with the current implementation.
3. Research what is the best way to implement the inter-process communication. Maybe some sort of json rpc over https://crates.io/crates/parity-tokio-ipc or whatever.
4. Implement the wrapper and make it work. Running the relay chain as an external process should always be some sort of optional way of doing it. So, if the feature is compiled it should be enabled via some cli flag or something. 
5. ...
6. Profit :P

xlc:
This should also make it possible to share a relaychain node between multiple parachain nodes? 

Will this make it possible to allow relaychain part and parachain part using different versions of Substrate?

NukeManDan:
Would you be able to have remote relay-nodes that collators could connect to? If so, what requirements (latency, bandwidth, etc.) should be outlined?

bkchr:
> This should also make it possible to share a relaychain node between multiple parachain nodes?

Yes, as written above. However, I don't see this as an initial requirement for this issue.

> Will this make it possible to allow relaychain part and parachain part using different versions of Substrate?

This is the whole point or better, make parachains collators not required to update when the relay chain requires a node update.

bkchr:
[...]

PR #963, opened 2022-02-07
Introduce rpc client for relay chain full node

skunert:
In this PR I continue the work towards #545. After this is merged, we support running a parachain full node via RPC. Running a collator is currently not supported.

This PR
- introduces an RPCClient that simply maps the required full node RPC methods
- introduces a second implementation of the `RelayChainInterface` Trait that fetches relay chain info from a full node rpc api via websockets
- modifies integration tests to include nodes that have this feature enabled
- adds command line option `relay-chain-rpc-url` to specify which relay chain node to talk to. Can not be used together with the relay-chain options that usually follow after `--`

I would consider this an MVP, there are some things that could be improved in the future, like making subscriptions more fault tolerant and caching of some local blocks that were already seen.


client/cli/src/lib.rs
```diff
@@ -119,6 +119,16 @@ impl sc_cli::CliConfiguration for PurgeChainCmd {
 	}
 }
 
+fn validate_relay_chain_url(arg: &str) -> Result<(), String> {
+	let url = Url::parse(arg).map_err(|e| e.to_string())?;
+
+	if url.scheme() == "ws" {
+		Ok(())
+	} else {
+		Err("Only websocket RPC currently supported".to_string())
```
koute, line 128: Print out the unexpected `url.scheme()` in the error message?

client/cli/src/lib.rs
```diff
@@ -131,6 +141,23 @@ pub struct RunCmd {
 	/// Note that this is the same as running with `--validator`.
 	#[clap(long, conflicts_with = "validator")]
 	pub collator: bool,
+
+	/// Specify relay chain full node to communicate
```
koute, line 145:
```suggestion
	/// Specify an URL to a relay chain full node to communicate with.
```

client/network/tests/sync_blocks_from_tip_without_collator.rs
```diff
@@ -27,12 +28,24 @@ async fn sync_blocks_from_tip_without_being_connected_to_a_collator() {
 	let para_id = ParaId::from(100);
 	let tokio_handle = tokio::runtime::Handle::current();
 
+	let ws_port = portpicker::pick_unused_port().expect("No free ports");
```
koute, line 31: Hmm.... while the chance is probably low, this could still return a port which isn't really free.

If we'd try to listen on port `0` the OS will atomically pick a free port by itself, so could we perhaps set the `ws_port` to `0` when calling `run_relay_chain_validator_node` and then just read which port was picked?

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+		.await
+	}
+}
+
+/// RelayChainNetwork is used to interact with a full node that is running locally
+/// in the same process.
+#[derive(Clone)]
+pub struct RelayChainNetwork {
+	rpc_client: RelayChainRPCClient,
+}
```
koute, line 261: Is there a point in having this split into two types (`RelayChainRPCClient` and `RelayChainNetwork`) considering `RelayChainRPCClient` is only used here, and `RelayChainNetwork` is for the most part only a thin wrapper over it?

client/service/src/lib.rs
```diff
@@ -193,6 +196,12 @@ where
 		.spawn_essential_handle()
 		.spawn("cumulus-consensus", None, consensus);
 
+	// PoV Recovery is currently not supported when we connect to the
+	// relay chain via RPC, so we return early.
+	if collator_options.relay_chain_rpc_url.is_some() {
+		return Ok(())
+	}
+
```
koute, line 204: I don't know what the consequences of this are, but is it fine to ignore it like this? (And if so I'd be nice to write in the comment why this is okay.)

client/relay-chain-local/src/lib.rs
```diff
@@ -345,7 +343,7 @@ fn build_polkadot_full_node(
 }
 
 /// Builds a relay chain interface by constructing a full relay chain node
-pub fn build_relay_chain_interface(
+pub fn build_relay_chain_local(
 	polkadot_config: Configuration,
```
koute, line 347: Please correct me if I'm wrong, but I assume we want to eventually remove this so that the only supported way will be to link the two nodes by RPC, right? If so would it make sense to print out a `warn!` that this is deprecated and that people should link the nodes by RPC? (Or is it too early for that yet?)

koute:
Do we have any performance comparisons regarding the total CPU usage running both nodes within the same process vs linking them through RPC?

Would be nice to run this on Rococo and see what the performance hit is. (If any.)

client/relay-chain-network/src/lib.rs
skunert: Yeah its only used here, in terms of abstraction levels it made sense to me.
- The RPCClient which is really just a "stupid" client that maps the full node RPC api 1:1, handles correct parameter handling and so on.
- The `RelayChainNetwork` which implements some logic like `wait_for_block` and also maps the RPC output to be in line with what `RelayChainLocal` already has. 

I could have done it in this in on struct, just seemed nicer to me to have this separated.

client/service/src/lib.rs
skunert: PoV recovery was recently introduced for full-nodes: #927 .

The node should work without it. As I continue to iterate on this issue, we should be able to add it back.

client/relay-chain-local/src/lib.rs
skunert: Yep, too early for now. There will be some follow-up PRs before I would actively advis users to switch to RPC. Might be worth adding a "experimental" flag or something to the command line 🤔

client/relay-chain-network/src/lib.rs
koute: If you want to separate them as an abstraction then it'd probably be a good idea to fully separate them since e.g. right now `RelayChainRPCClient` doesn't have its own constructor and `RelayChainNetwork::new` initializes it field-by-field, so it's not really abstracted away. (:

client/network/tests/sync_blocks_from_tip_without_collator.rs
skunert: As I see it the portpicker has logic to let the OS pick the port. Or do you mean the port could become occupied after picking the port here but before the full-node is started?

koute: It can let the OS pick the port, but it immediately releases it, so it's inherently racy and something which runs in parallel (another thread, another program on the same machine, etc.) could use that port before we manage to start the full node.

client/service/src/lib.rs
koute: Sounds good to me! (In this case I'd suggest to simply just add a short comment or something that this is not necessary and that the node will still work without it, so that there's no confusion.)

client/relay-chain-local/src/lib.rs
koute: Yeah, an experimental flag sounds good to me!

skunert: Added `EXPERIMENTAL` to the cli flag

client/network/tests/sync_blocks_from_tip_without_collator.rs
skunert: As discussed in the meeting, I do not see an easy way to get the port here right now. I will observe if we see failures due to this, but should be good for now.

parachain-template/node/src/service.rs
```diff
@@ -328,7 +356,7 @@ where
 			spawner,
 			parachain_consensus,
 			import_queue,
-			collator_key,
+			collator_key: collator_key.expect("Unable to start collator without collator key."),
```
bkchr, line 359: Same

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+	retry_strategy: ExponentialBackoff,
+}
+
+impl RelayChainRPCClient {
+	pub async fn new(url: Url) -> Self {
+		tracing::info!(target: LOG_TARGET, "Initializing RPC Client. Relay Chain Url: {}", url);
+		let ws_client = WsClientBuilder::default()
+			.build(url.as_str())
+			.await
+			.expect("Should be able to initialize websocket client.");
```
bkchr, line 68: Please return an error.

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+		retry_notify(
+			self.retry_strategy.clone(),
+			|| async {
+				self.ws_client
+					.request(method, params.clone())
+					.await
+					.map_err(|err| backoff::Error::Transient { err, retry_after: None })
+			},
+			|e, _| {
+				tracing::warn!("Unable to complete RPC request '{}', retrying. Error {}", method, e)
```
bkchr, line 126:
```suggestion
				tracing::warn!(target: LOG_TARGET, error = %e, %method, "Unable to complete RPC request, retrying")
```

`tracing` provides these nice ways of "attaching" data to the log output.

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+		payload: Option<Vec<u8>>,
+	) -> RelayChainResult<sp_core::Bytes> {
+		let payload_bytes = payload.map_or(sp_core::Bytes(Vec::new()), sp_core::Bytes);
+		let params = rpc_params! {
+			method_name,
+			payload_bytes,
+			hash
+		};
+		self.request("state_call", params).await
+	}
```
bkchr, line 90:
```suggestion
	async fn call_remote_runtime_function<R: Decode>(
		&self,
		method_name: &str,
		hash: &PHash,
		payload: Option<Vec<u8>>,
	) -> RelayChainResult<R> {
		let payload_bytes = payload.map_or(sp_core::Bytes(Vec::new()), sp_core::Bytes);
		let params = rpc_params! {
			method_name,
			payload_bytes,
			hash
		};
		let res = self.request("state_call", params).await?;
		
		Decode::decode(&mut res.0[..]).map_err(Into::into)
	}
```

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+	) -> Result<Option<CommittedCandidateReceipt>, RelayChainError> {
+		let response_bytes = self
+			.call_remote_runtime_function(
+				"ParachainHost_candidate_pending_availability",
+				at,
+				Some(para_id.encode()),
+			)
+			.await?;
+
+		Ok(Option::<CommittedCandidateReceipt<PHash>>::decode(&mut &*response_bytes.0)?)
```
bkchr, line 180:
```suggestion
		self
			.call_remote_runtime_function(
				"ParachainHost_candidate_pending_availability",
				at,
				Some(para_id.encode()),
			)
			.await
```

If you use my code from above.

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+	) -> Result<Vec<ValidatorId>, RelayChainError> {
+		let response_bytes =
+			self.call_remote_runtime_function("ParachainHost_validators", at, None).await?;
+
+		Ok(Vec::<ValidatorId>::decode(&mut &*response_bytes.0)?)
+	}
+
+	async fn parachain_host_persisted_validation_data(
+		&self,
+		block_id: &PHash,
```
bkchr, line 206: hash not a block_id :P

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+			None,
+		)
+		.await
+	}
+}
+
+/// RelayChainNetwork is used to interact with a full node that is running locally
+/// in the same process.
+#[derive(Clone)]
+pub struct RelayChainNetwork {
```
bkchr, line 272: The relay chain network is a bad name, because you normally call all relay chain networks the "relay chain network"

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+					tracing::error!(
+						target: LOG_TARGET,
+						"Encountered error in import notification stream: {}",
+						err
+					)
+				})
+				.ok()
+			});
+
+		Ok(Box::pin(imported_headers_stream))
```
bkchr, line 346:
```suggestion
		Ok(imported_headers_stream.boxed())
```

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+					tracing::error!(
+						target: LOG_TARGET,
+						"Encountered error in finality notification stream: {}",
+						err
+					)
+				})
+				.ok()
+			});
+
+		Ok(Box::pin(imported_headers_stream))
```
bkchr, line 367: Same

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+	///
+	/// The hash of the block to wait for is passed. We wait for the block to arrive or return after a timeout.
+	///
+	/// Implementation:
+	/// 1. Register a listener to all new blocks.
+	/// 2. Check if the block is already in chain. If yes, succeed early.
+	/// 3. Wait for the block to be imported via subscription.
+	/// 4. If timeout is reached, we return an error.
+	async fn wait_for_block(&self, wait_for_hash: PHash) -> RelayChainResult<()> {
+		let mut head_stream = self.rpc_client.subscribe_all_heads().await?;
```
bkchr, line 421: Do you know when this future returns here? Does it return when the other side has signaled us that the subscription is done?

client/relay-chain-network/src/lib.rs
```diff
@@ -0,0 +1,460 @@
...
+
+		let block_header = self.rpc_client.chain_get_header(Some(wait_for_hash)).await?;
+		if block_header.is_some() {
```
bkchr, line 424:
```suggestion
		if self.rpc_client.chain_get_header(Some(wait_for_hash)).await?.is_some() {
```

client/relay-chain-network/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+authors = ["Parity Technologies <admin@parity.io>"]
+name = "cumulus-relay-chain-network"
```
bkchr, line 3: Crate name should also be renamed.

polkadot-parachains/src/service.rs
```diff
@@ -607,7 +640,7 @@ where
 			spawner,
 			parachain_consensus,
 			import_queue,
-			collator_key,
+			collator_key: collator_key.expect("Can not start collator without collator key"),
```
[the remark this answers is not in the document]
skunert: Users can not encounter this message, if this is ever thrown we have a bug. They can only choose to supply a relay chain url (so full node only), then we do go this code path. Or they want to start a collator, then we start an inprocess full-node and have the collator key.

test/service/src/lib.rs
```diff
@@ -219,13 +246,13 @@ where
...
 
-	let relay_chain_interface = Arc::new(RelayChainLocal::new(
-		relay_chain_full_node.client.clone(),
-		relay_chain_full_node.backend.clone(),
-		Arc::new(Mutex::new(Box::new(relay_chain_full_node.network.clone()))),
-		relay_chain_full_node.overseer_handle.clone(),
-	));
-	task_manager.add_child(relay_chain_full_node.task_manager);
+	let relay_chain_interface = build_relay_chain_interface(
+		relay_chain_full_node,
```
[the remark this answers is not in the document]
skunert: We do not, the `collator_options` that go inside here decide whether we instantiate the full node or rpc version. We only build it when required.

client/relay-chain-interface/src/lib.rs
```diff
@@ -19,54 +19,63 @@ use std::{collections::BTreeMap, pin::Pin, sync::Arc};
...
 
 use async_trait::async_trait;
+use jsonrpsee_core::Error as JsonRPSeeError;
+use parity_scale_codec::Error as CodecError;
 use sp_api::ApiError;
 use sp_state_machine::StorageValue;
 
 pub type RelayChainResult<T> = Result<T, RelayChainError>;
 
 #[derive(thiserror::Error, Debug)]
```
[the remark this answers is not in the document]
koute: I don't think this can be done, since `std::error::Error` requires `Debug` unfortunately.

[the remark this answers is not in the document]
skunert: Yes I agree, I dont this change is possible

parachain-template/node/Cargo.toml
```diff
@@ -92,7 +92,8 @@ cumulus-client-service = { path = "../../client/service" }
 cumulus-primitives-core = { path = "../../primitives/core" }
 cumulus-primitives-parachain-inherent = { path = "../../primitives/parachain-inherent" }
 cumulus-relay-chain-interface = { path = "../../client/relay-chain-interface" }
-cumulus-relay-chain-local = { path = "../../client/relay-chain-local" }
+cumulus-relay-chain-inprocess-interface = { path = "../../client/relay-chain-inprocess-interface" }
+cumulus-relay-chain-rpc-interface = { path = "../../client/relay-chain-rpc-interface" }
```
bkchr, line 96: As this is not stable yet, we should not add this to the template for now.

polkadot-parachains/src/service.rs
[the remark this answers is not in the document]
bkchr: Hmm okay. Then please convert this into a proof "Bla bla bla; qed". Where "bla bla bla" is the reason this can not happen.

polkadot-parachains/src/service.rs
```diff
@@ -1433,7 +1481,7 @@ where
...
```
bkchr, line 1484: Same

polkadot-parachains/src/service.rs
```diff
@@ -1346,12 +1389,17 @@ where
...
+	let (relay_chain_interface, collator_key) = build_relay_chain_interface(
+		polkadot_config,
+		telemetry_worker_handle,
+		&mut task_manager,
+		collator_options.clone(),
+	)
+	.await
+	.map_err(|e| match e {
+		RelayChainError::ServiceError(polkadot_service::Error::Sub(x)) => x,
+		s => format!("{}", s).into(),
```
bkchr, line 1401:
```suggestion
		s => s.into_string().into(),
```

test/service/src/lib.rs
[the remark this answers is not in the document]
bkchr: I mean inside the `build_relay_chain_interface` function. When we should use a RPC, we just throw away the relay_chain_full_node we just have created

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+use sc_rpc_api::{state::ReadProof, system::Health};
+use sp_core::sp_std::collections::btree_map::BTreeMap;
+use sp_runtime::DeserializeOwned;
+use sp_state_machine::StorageValue;
+use sp_storage::StorageKey;
+use std::{pin::Pin, sync::Arc};
+
+pub use url::Url;
+
+const LOG_TARGET: &str = "relay-chain-external-interface";
```
bkchr, line 50:
```suggestion
const LOG_TARGET: &str = "relay-chain-rpc-interface";
```

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+	/// Perform RPC request
+	async fn request_tracing<'a, R, OR>(
+		&self,
+		method: &'a str,
+		params: Option<ParamsSer<'a>>,
+		trace_error: OR,
+	) -> Result<R, RelayChainError>
+	where
+		R: DeserializeOwned + std::fmt::Debug,
+		OR: Fn(&jsonrpsee::core::Error) -> (),
```
bkchr, line 143:
```suggestion
		OR: Fn(&jsonrpsee::core::Error),
```

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+		.await
+	}
+
+	async fn parachain_host_session_index_for_child(
+		&self,
+		at: PHash,
+	) -> Result<SessionIndex, RelayChainError> {
+		self.call_remote_runtime_function("ParachainHost_session_index_for_child", at, None)
+			.await
+		// Ok(SessionIndex::decode(&mut &*response_bytes.0)?)
```
bkchr, line 215:
```suggestion
```

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+			retry_strategy: ExponentialBackoff::default(),
+		})
+	}
+
+	/// Call a call to `state_call` rpc method.
+	async fn call_remote_runtime_function<R: Decode>(
+		&self,
+		method_name: &str,
+		hash: PHash,
+		payload: Option<Vec<u8>>,
```
bkchr, line 79:
```suggestion
		payload: Option<impl Encode>,
```

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+	async fn parachain_host_persisted_validation_data(
+		&self,
+		at: PHash,
+		para_id: ParaId,
+		occupied_core_assumption: OccupiedCoreAssumption,
+	) -> Result<Option<PersistedValidationData>, RelayChainError> {
+		self.call_remote_runtime_function(
+			"ParachainHost_persisted_validation_data",
+			at,
+			Some((para_id, occupied_core_assumption).encode()),
```
bkchr, line 234:
```suggestion
			Some((para_id, occupied_core_assumption)),
```

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+
+	async fn parachain_host_inbound_hrmp_channels_contents(
+		&self,
+		para_id: ParaId,
+		at: PHash,
+	) -> Result<BTreeMap<ParaId, Vec<InboundHrmpMessage>>, RelayChainError> {
+		self.call_remote_runtime_function(
+			"ParachainHost_inbound_hrmp_channels_contents",
+			at,
+			Some(para_id.encode()),
```
bkchr, line 247:
```suggestion
			Some(para_id),
```

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+		)
+		.await
+	}
+
+	async fn parachain_host_dmq_contents(
+		&self,
+		para_id: ParaId,
+		at: PHash,
+	) -> Result<Vec<InboundDownwardMessage>, RelayChainError> {
+		self.call_remote_runtime_function("ParachainHost_dmq_contents", at, Some(para_id.encode()))
```
bkchr, line 257:
```suggestion
		self.call_remote_runtime_function("ParachainHost_dmq_contents", at, Some(para_id))
```

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+	}
+
+	/// Call a call to `state_call` rpc method.
+	async fn call_remote_runtime_function<R: Decode>(
+		&self,
+		method_name: &str,
+		hash: PHash,
+		payload: Option<Vec<u8>>,
+	) -> RelayChainResult<R> {
+		let payload_bytes = payload.map_or(sp_core::Bytes(Vec::new()), sp_core::Bytes);
```
bkchr, line 81:
```suggestion
		let payload_bytes = payload.map_or(sp_core::Bytes(Vec::new()), |v| sp_core::Bytes(v.encode()));
```

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+
+	async fn parachain_host_candidate_pending_availability(
+		&self,
+		at: PHash,
+		para_id: ParaId,
+	) -> Result<Option<CommittedCandidateReceipt>, RelayChainError> {
+		self.call_remote_runtime_function(
+			"ParachainHost_candidate_pending_availability",
+			at,
+			Some(para_id.encode()),
```
bkchr, line 204:
```suggestion
			Some(para_id),
```

bkchr (approved):
Looks good, just some last nitpicks.

client/relay-chain-interface/src/lib.rs
```diff
@@ -98,7 +107,7 @@ pub trait RelayChainInterface: Send + Sync {
 	/// and the para already occupies a core.
 	async fn persisted_validation_data(
 		&self,
-		block_id: &BlockId,
+		block_id: &PHash,
```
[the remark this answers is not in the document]
bkchr: @skunert this wasn't done?

client/network/tests/sync_blocks_from_tip_without_collator.rs
```diff
@@ -16,6 +16,7 @@
 
 use cumulus_primitives_core::ParaId;
 use cumulus_test_service::{initial_head_data, run_relay_chain_validator_node, Keyring::*};
+use futures::join;
```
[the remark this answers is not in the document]
bkchr: Same

client/relay-chain-network/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+sp-core = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-api = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-runtime = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-state-machine = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sc-client-api = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sp-storage = { git = "https://github.com/paritytech/substrate", branch = "master" }
+sc-rpc-api = { git = "https://github.com/paritytech/substrate", branch = "master" }
+
+futures = { version = "0.3.1", features = ["compat"] }
```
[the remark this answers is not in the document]
bkchr: This was also not done.

parachain-template/node/Cargo.toml
skunert: I can remove it here, but are you sure its necessary? It is clearly marked experimental here https://github.com/paritytech/cumulus/pull/963/files#diff-782fb6433e0fb2f09182fc6daeb68222bb0b6c9bff76cacc6ffa23847e26cf70R148 . On removal, we would need to provide a second set of cli arguments (without url argument). And people can only run a full-node with this anyway.

client/relay-chain-interface/src/lib.rs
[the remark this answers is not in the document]
skunert: @bkchr Were you maybe looking at an outdated version? This was fixed after your last review

parachain-template/node/Cargo.toml
skunert: FYI: We had a live discussion about this, decided to leave it in.

Issue #969, opened 2022-02-09
Merge back to master from 7.0 branch


gilescope:
We should merge the following changes back to master:

- [x] https://github.com/paritytech/cumulus/commit/baaa58319ae7007f439d05f0f5fc0d23f84a6da6
- [x] https://github.com/paritytech/cumulus/commit/7881fe6a16498a949bdc6e892a7aee2365a51c06
- [x] https://github.com/paritytech/cumulus/commit/dfd7a9a8f558296a93c6b51f3be531ccbba4f71f - already in master. 
- [x] https://github.com/paritytech/cumulus/commit/be8b8e6916eeaec19302128b99700cf4e3d7d781
- [x] https://github.com/paritytech/cumulus/commit/692a6d10be1a8d7d7402cdad1730ea318350daf1 - already in master.
- [x] https://github.com/paritytech/cumulus/commit/16d3530fc9acfb0d5c4257f66dadcd5169db26f3 - already in master.
- [x] https://github.com/paritytech/cumulus/commit/f01c737e89096047a362134815cdf4cce8758be9
- [x] https://github.com/paritytech/cumulus/commit/664c12f8243de7f1fca2ceeea190f0f740deb2fb
- [x] and last but not least, we should on master remove the unique v1 migration.

gilescope:
ok just need to drop the uniques migration and then we are clean for next release.

gilescope:
all done.

PR #971, opened 2022-02-09
Set statemin runtime versions to 700

gilescope:
Note I have left rococo and seedling versions alone from the original commit ( https://github.com/paritytech/cumulus/commit/7881fe6a16498a949bdc6e892a7aee2365a51c06 ) because I don't think we should be bumping them.

(part of https://github.com/paritytech/cumulus/issues/969 )


hbulgarini (approved):

gilescope:
bot merge

PR #980, opened 2022-02-10
Add `canvas-kusama` runtime

athei:
The canvas runtime is currently hosted at https://github.com/paritytech/canvas. We have no capacities to keep up with substrate, cumulus and polkadot. Therefore we want to merge the runtime here so that it stays in sync and get some love by the common good team.

This is mostly equivalent to the statemine runtime but with `pallet-contracts` added. It is slightly different from what is currently hosted at the `canvas` repository in that it adds the `multisig` and `utilities` pallet.

The plan is to use the same runtime on Kusama and Rococo. This is why the runtime is called `canvas-kusama` but the chainspec is called `canvas-rococo`. There is no chain spec for kusama, yet.


polkadot-parachains/src/chain_spec.rs
```diff
@@ -834,3 +834,252 @@ fn westmint_genesis(
...
+const KANVAS_ED: kanvas_runtime::Balance = kanvas_runtime::constants::currency::EXISTENTIAL_DEPOSIT;
+
+pub fn rocanvas_development_config() -> KanvasChainSpec {
+	let mut properties = sc_chain_spec::Properties::new();
+	properties.insert("tokenSymbol".into(), "ROC".into());
+	properties.insert("tokenDecimals".into(), 12.into());
+
+	KanvasChainSpec::from_genesis(
+		// Name
+		"Rokanvas Development",
```
HCastano, line 854: This is gonna sound radical, but I think it would be much less confusing if this were simply `Canvas Development`.

Same with the `Canvas Local` somewhere down below.

We already use the Canvas name for the existing Rococo deployment anyways.

polkadot-parachains/src/chain_spec.rs
```diff
@@ -834,3 +834,252 @@ fn westmint_genesis(
...
+		Some(properties),
+		// Extensions
+		Extensions {
+			relay_chain: "rococo-local".into(), // You MUST set this to the correct network!
+			para_id: CANVAS_PARACHAIN_ID,
+		},
+	)
+}
+
+pub fn rocanvas_config() -> KanvasChainSpec {
```
HCastano, line 959:
```suggestion
pub fn canvas_rococo_config() -> KanvasChainSpec {
```

HCastano (commented):
Just here for some bikeshedding, will take a real look tomorrow

polkadot-parachains/src/chain_spec.rs
athei: I disagree. `Canvas` runtime doesn't even exist. Runtime is `Kanvas` (runtime for kusama) which we also use for rococo (statemine uses westmint for that). How would you name the function for kusama then? `canvas_kusama_config`?

athei:
All names but `Canvas` are up for debate. So please bring all your ideas forward for `Kanvas` and `Rokanvas`.

gilescope:
Kangas is Finnish for canvas and has a certain appeal?

HCastano:
> Just an observation that canvas and kanvas sound the same when pronounced out loud. This will lead to confusion alas. I love the spelling of kanvas. Maybe we just need an even cooler name when it launches on polkadot?

When coming up with the name having both sound the same was done on purpose. Personally, I still have a hard time keeping track of which chain Statemine and Statemint correspond to. Really the most important thing is the context in which you're talking about the chain, e.g are you talking about Polkadot or Kusama. Then it doesn't matter if they sound they same (or if they're written the same way)

polkadot-parachains/src/chain_spec.rs
athei: Either we give every parachain (i.e chain spec) a different name or we don't. But having `Kanvas` as a name for the Kusama parachain but naming the Rococo parachain `Canvas on Rococo` is inconsistent.

athei:
But when you tell someone in speaking "I deploy that on Kanvas" they need to ask "the polkadot or kusama one?" no matter how much knowledge they have. I think this will be confusing. I consider statemine and statemint actually a very good example of naming. You know they belong together but they are sill distinguishable.

athei:
So after some intense bike shedding the contracts team came to the agreement that we don't want to continue the tradition of having a unique name for every parachain. Instead, we want to focus on the `Canvas` name. This is why I pushed a commit that removes all names except `Canvas` and instead suffixes identifiers it with the relay chain it is deployed to.

cmichi:
Is the integration into the `cumulus` CI missing or is this something that already happens automatically now?

Also you might want to copy the `polkadot-launch` config over as well.

polkadot-parachains/src/chain_spec.rs
```diff
@@ -834,3 +834,254 @@ fn westmint_genesis(
...
+
+pub fn canvas_rococo_config() -> CanvasKusamaChainSpec {
+	// Give your base currency a unit name and decimal places
+	let mut properties = sc_chain_spec::Properties::new();
+	properties.insert("tokenSymbol".into(), "ROC".into());
+	properties.insert("tokenDecimals".into(), 12.into());
+
+	CanvasKusamaChainSpec::from_genesis(
+		// Name
+		"Canvas on Rococo",
```
HCastano, line 969: I wonder, is this a breaking change? E.g, would we need to restart the network after having changed the name?

This is a dumb Q for @riusricardo

HCastano (commented):
Looks good, but I just wanna confirm that dumb Q I have before ✅

README.md
```diff
@@ -54,14 +59,104 @@ CHAIN=westmint # or statemine
 
 Refer to the [setup instructions below](#local-setup) to run a local network for development.
 
-# Rococo :crown:
+## Canvas 🧑‍🎨
```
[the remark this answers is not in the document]
HCastano: We talked about this, and we can do this in a follow-up

HCastano (approved):

polkadot-parachains/src/chain_spec.rs
bkchr: No you don't need to. This is a value in the chain spec.

Issue #961, opened 2022-02-05
Create propoer benchmarking for XCMP QueueConfigData


bkchr:
The weights are currently just "guessed", but we should write some benchmarks to create them properly. As it is just about setting data, the benchmark should be relative easy.

See the following discussion:

https://github.com/paritytech/cumulus/pull/931#discussion_r796644140_

Doordashcon:
I understand that benchmarking your extrinsic means to account for the worst case scenario. For parameters with `u32` type, i imagine the worst case would be `u32::MAX` and `Weight::MAX` for parameter with `Weight` then a refund will be issued to the caller if needed but `Weight::MAX` seems too large.

Instead the weight definition could be `#[pallet::weight(T::WeightInfo::set_u32(*new))`?

KiChjang:
The worst case complexity in assigning a `u32` doesn't really require you to use `u32::MAX`, I doubt that would have any effect on the runtime complexity at all, so your question is on the wrong track.

Refer to the benchmarks on the paras configuration pallet on polkadot https://github.com/paritytech/polkadot/blob/master/runtime/parachains/src/configuration/benchmarking.rs for prior art.

Doordashcon:
I am unfamiliar with the benchmarking command/workflow used on Cumulus, for Substrate it's [https://docs.substrate.io/v3/runtime/benchmarking/](https://docs.substrate.io/v3/runtime/benchmarking/). 
Could you please direct me to a resource i'm trying to generate the weights.

gilescope:
It has to be done on a specific bench-marking machine so that the benchmarks are comparable. If there's a PR then I think we can kick off a benchmark run for it.

Doordashcon:
believe this has been addressed :)

PR #982, opened 2022-02-11
Add benchmarks for Xcmp QueueConfigData setters

Doordashcon:
This PR attempts to add benchmarks for the Xcmp Queue Pallet #961


pallets/xcmp-queue/Cargo.toml
```diff
@@ -45,3 +48,9 @@ std = [
 	"xcm/std",
 	"xcm-executor/std",
 ]
+
+runtime-benchmarks = [
+	"frame-benchmarking/runtime-benchmarks",
+	"frame-support/runtime-benchmarks",	
+	"frame-system/runtime-benchmarks",
+]
```
KiChjang, line 56: Missing newline.

gilescope, line 56:
```suggestion
]

```

gilescope:
I'm running a benchmark run at the moment for this. Will update the branch with results.

gilescope:
Why are the benchmarks defined in the cumulus xcmp queue pallet used on westmint to benchmark the cumulus dmp queue pallet? both pallets are used in all west/statemine/t parachains...

Is there any reason why we should not define benchmarks on both queue pallets or would doing so somehow double count the weight?

gilescope:
(I need to rerun benchmarks for statemine/t but want to understand if we should be benching one or two pallets first).

gilescope:
The `[cumulus_pallet_dmp_queue, XcmpQueue]` didn't make much sense to me in westmint lib.rs so changed it to `[cumulus_pallet_xcmp_queue, XcmpQueue]` - let me know if there was a reason behind that or if it was just a typo.

Doordashcon:
> The `[cumulus_pallet_dmp_queue, XcmpQueue]` didn't make much sense to me in westmint lib.rs so changed it to `[cumulus_pallet_xcmp_queue, XcmpQueue]` - let me know if there was a reason behind that or if it was just a typo.

LGTM

gilescope:
Will rerun the weights for the pallet again now that I have expanded the benchmark.

pallets/xcmp-queue/src/benchmarking.rs
```diff
@@ -21,24 +21,17 @@ use frame_benchmarking::{benchmarks, impl_benchmark_test_suite};
...
-	//   `test benchmarking::bench_accumulate_dummy ... ok`
-	//   `test benchmarking::bench_set_dummy_benchmark ... ok` in the result.
-	//
-	// The line generates three steps per benchmark, with repeat=1 and the three steps are
-	//   [low, mid, high] of the range.
+	set_config_with_threshold_weight {}: update_threshold_weight(RawOrigin::Root, 3_000_000)
+
+	set_config_with_drop_threshold {}: update_drop_threshold(RawOrigin::Root, 3_000_000)
+
+	set_config_with_suspend_threshold {}: update_suspend_threshold(RawOrigin::Root, 3_000_000)
```
KiChjang, line 34: We can actually combine all of these benchmarks in a single one, since there isn't any behavioral or weight difference between setting all of these `u32` parameters.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -198,7 +205,7 @@ pub mod pallet {
 		/// message sending may recommence after it has been suspended.
 		///
 		/// - `origin`: Must pass `Root`.
-		/// - `new`: Desired value for `QueueConfigData.resume_threshold`                                
+		/// - `new`: Desired value for `QueueConfigData.resume_threshold`
 		#[pallet::weight(10_000_000 as Weight + T::DbWeight::get().reads_writes(1, 1))]
```
KiChjang, line 209: These weights need to be updated once you have the benchmarked weight files.

gilescope: Ah yes - I see now that's how the WeighInfos get called.

pallets/xcmp-queue/src/benchmarking.rs
gilescope: Agreed - that's much nicer. I see now how that can be achieved.

gilescope:
Am hoping I have done everything right now?

gilescope (approved):
LGTM but then I would say that...

NachoPal (changes requested):
A few things to change:
- `update_threshold_weight` and `update_weight_restrict_decay` have `Weight`(u64) as argument, not `u32`. Use instead `set_config_with_weight()`. Example -> https://github.com/paritytech/polkadot/blob/master/runtime/parachains/src/configuration/benchmarking.rs#L29
- Add `DispatchClass::Operational`. Example -> https://github.com/paritytech/polkadot/blob/master/runtime/parachains/src/configuration.rs#L868-L870
- `WeightInfo` should be imported from the benchmark results under each runtime `./src//weights` folder, not from `SubstrateWeight`
    ```
    type WeightInfo = weights::cumulus_pallet_xcmp_queue::WeightInfo<Runtime>;
    ```

pallets/xcmp-queue/src/lib.rs
```diff
@@ -93,6 +98,8 @@ pub mod pallet {
 		/// The conversion function used to attempt to convert an XCM `MultiLocation` origin to a
 		/// superuser origin.
 		type ControllerOriginConverter: ConvertOrigin<Self::Origin>;
+
```
NachoPal, line 101:
```suggestion
/// The weight information of this pallet.
```

Doordashcon:
If we won't need to rerun Benchmarks i'll handle the requested changes @gilescope

gilescope:
I believe the first two points are addressed.
For the third point:
"WeightInfo should be imported from the benchmark results under each runtime ./src/weights folder, not from SubstrateWeight". Looking again at this, I think this is what the PR is doing for westmint, statmine, and statemint? I am using substrate weights for rococo, canvas and the parachain-template as these don't have benchmarking associated with them (though I am sure this will be coming for canvas).

gilescope:
bot merge

pallets/xcmp-queue/src/lib.rs
```diff
@@ -122,7 +130,7 @@ pub mod pallet {
 		///
 		/// Events:
 		/// - `OverweightServiced`: On success.
-		#[pallet::weight(weight_limit.saturating_add(1_000_000))]
+		#[pallet::weight((weight_limit.saturating_add(1_000_000), DispatchClass::Operational,))]
 		pub fn service_overweight(
```
KiChjang, line 134: I find it weird that this function is left unbenchmarked.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -145,7 +153,7 @@ pub mod pallet {
 		/// Suspends all XCM executions for the XCMP queue, regardless of the sender's origin.
 		///
 		/// - `origin`: Must pass `ControllerOrigin`.
-		#[pallet::weight(T::DbWeight::get().writes(1))]
+		#[pallet::weight((T::DbWeight::get().writes(1), DispatchClass::Operational,))]
```
bkchr, line 156: How is that ensured that this is only one write? `ensure_origin` could do any kind of operation. Same applies for the others.

pallets/xcmp-queue/src/lib.rs
bkchr: Yes. What is the reason for this?

gilescope:
Agreed - service_overweight is more of a normal user call. I will set this one back to normal when I do the service_overweight bench pr.

pallets/xcmp-queue/src/lib.rs
gilescope: That seemed a different call to all the rest. Will add it to the benchmarks in another PR.

pallets/xcmp-queue/src/lib.rs
gilescope: The right answer is to benchmark these methods too? Ok I can do that - that will mean they're all bench-marked which is great for consistency.

pallets/xcmp-queue/src/lib.rs
gilescope: Should I be doing the benchmark for service_overweight in dmp-queue at the same time?

pallets/xcmp-queue/src/lib.rs
bkchr: Yes this should also be benchmarked.

PR #995, opened 2022-02-19
Fix missing relay chain telemetry node name (issue #556)

### LINKED ISSUE #556: --name flag is not working for the embedded relay node
Parachain nodes are supposed to pass any cli flags that are supplied after a -- to the embedded relay chain node. However, it seems that the --name parameter does not affect the embedded relay chain node. --name works fine for the parachain node itself, and many flags work fine for the embedded relay, but the combination does not work together.

Example command to reproduce this issue: `moonbeam --chain moonriver --name test-node-parachain -- --name test-node-embedded-relay`

To see this, we can go to any node that this node is connected to and look at its relay peers. If we do that we will see that the name sent over the peer is something randomly generated and not what we sent in as a CLI flag, for example this one: `agreeable-step-8709`.

JelliedOwl:
I believe this fixes #556, but I'm sure there must be a cleaner option...  My first rust / substrate PR - feedback and suggestions would be most welcome.

Essentially, the first (original) call to SubstrateCli::create_configuration seems to be appropriately setting most of the relay chain node parameters, including using default that don't clash with the parachain side, but it sets the node name to an auto-generated value.  The second (added) call to SubstrateCli::create_configuration sets the node name according to the --name flag (or an another auto-generated value), but most of the other parameters are not appropriate for the relay chain connection.

I couldn't work out a way to move the "name" value in polkadot_cli to a place where the first call would find it, and all the other fixes I could think of either required changes to other crates or reproducing code from those crates in this function.  I wanted to avoid doing either of those.  Adding just the section of code needed would be a bit more computationally efficient but, since this code is only called once at node startup, I thought that probably wasn't a concern - reusing known-good functionality seemed better.

I think I found the only two places in the repo that needed updating.

### DIFF
--- a/parachain-template/node/src/command.rs
+++ b/parachain-template/node/src/command.rs
@@ -283,9 +283,16 @@ pub fn run() -> Result<()> {
 				let genesis_state = format!("0x{:?}", HexDisplay::from(&block.header().encode()));
 
 				let tokio_handle = config.tokio_handle.clone();
-				let polkadot_config =
+				let mut polkadot_config =
 					SubstrateCli::create_configuration(&polkadot_cli, &polkadot_cli, tokio_handle)
 						.map_err(|err| format!("Relay chain argument error: {}", err))?;
+
+				// Fix missing node name
+				let tokio_handle = config.tokio_handle.clone();
+				let polkadot_config_node_name =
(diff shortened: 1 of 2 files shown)


bkchr:
Hey, ty for the pr, but that isn't the correct fix. You can see the correct fix here: https://github.com/paritytech/cumulus/pull/1008

PR #1009, opened 2022-02-21
Canvas: Bump `spec_version` and `state_version`

athei:
We can use the new `state_version` as we intent to reset the rococo para chain with the next upgrade anyways. `spec_version` is bumped because it wasn't since the last runtime upgrade.


cmichi (approved):

cheme (approved):
:+1: would be the first chain (test chain) to run the new version.

athei:
bot merge

PR #1012, opened 2022-02-21
updating benchmarking script

gilescope:
we should be using production profile for benchmarks.
`--raw` does not exist: https://github.com/paritytech/substrate/pull/10771


NachoPal (approved):

ggwpez:
Do you actually use the stdout `json`? If you want it into a file you can use `--json-file=path` instead.  
@gilescope

PR #1013, opened 2022-02-21
Companion to "Updating scale to v3"

wigy-opensource-developer:
Companion to paritytech/substrate#10825


KiChjang (commented):
This is the repo where I expected changes to usages of `decode`, `decode_all`, `decode_with_depth_limit` and `decode_and_advance_with_depth_limit` because I remember coding them before, but I don't see any of those changes in this PR, is that intended?

wigy-opensource-developer:
I am so new to the cumulus codebase and using diener for 3 PRs on my machine that I wanted to rely on CI to tell me if the build or tests are broken. But the CI is now blocked on the rocksdb upgrade at the moment, so I did not get here yet. Also, CI does not run on draft PRs, so I could not use that to signal this is not ready yet.

PR #1031, opened 2022-02-22
Prepare `Canvas` for redeployment on rococo

athei:
I built the current runtime with the `production` profile and generated the chainspec from it. Also change the code to use this chainspec when `--chain canvas-rococo` is used.


cmichi:
Why do the binary files need to be committed to this repo?

athei:
They don't. The only file we need to commit is `canvas-rococo.json`. The rest can be generated from it. I could upload it somewhere else to not bloat the repo further. I figured I just do what the other chains did and just commit it.

athei:
Removed those files and forced pushed. Then they won't clutter the history.

cmichi (approved):

ascjones (approved):

athei:
bot merge

Issue #984, opened 2022-02-11
Runtime Only Releases for Parachains


joepetrowski:
We should not need to release a node binary (right now, `polkadot-collator`) with parachain runtimes. Validators execute the parachain blocks in Wasm anyway, so it should be fine to decouple the node release from the runtime release as a native runtime isn't needed.

Ideally, release processes would involve:

1. Release of `polkadot-vX.Y.Z`
2. Release of Cumulus `vX.Y.Z`
3. Release of parachain runtimes that will work on `polkadot-collator` versions `> vX.Y.Z`

With https://github.com/paritytech/cumulus/issues/948, this will be further simplified by removing step 2.

To accomplish this, I'd like to check the following:

- [x] Check Wasm allocator to ensure we don't run into problems (I don't know entirely what I'm talking about here, @bkchr flagged it to me)
- [x] Verify that nodes from the Cumulus `vX.Y.Z` release work properly (https://github.com/paritytech/devops/issues/1397)
- [x] Ensure that we can start local testnets from genesis for testing purposes using a node without the native runtime (point 3 of https://github.com/paritytech/substrate/issues/10579)

After these are OK, let's change release CI to not include a binary with runtime releases. We will also need to make sure that Cumulus `vX.Y.Z` does get a release and binary.

joepetrowski:
We are still beholden to a few things, like having the chain specs stored in Cumulus, that prevent complete extraction of the runtimes from the node, but I think we are quite good on this for the time being.

PR #1054, opened 2022-02-28
polkadot-collator: Switch to wasm only

bkchr:
This switches the polkadot-collator to run everything in wasm only mode. While we should not that
yet with the relay chain, because it can happen that we run out of memory (very unlikely). On the
relay chain that would be bad, because we only have at max 2 sessions to bring everything back, for
Parachains that isn't such a problem as they would only stall and we could roll out a release that
fixes it. Besides that, Parachain validation on the relay chain happens in Wasm already all the time
and there is the memory usage even higher then on block import.

Progress on: https://github.com/paritytech/cumulus/issues/984


skunert:
After this change, will the `--execution` flag not be useless? Can we remove it?

polkadot-parachains/src/service.rs
```diff
@@ -1510,7 +1475,7 @@ pub fn canvas_kusama_build_import_queue(
 			Ok((timestamp, slot))
 		},
 		registry: config.prometheus_registry(),
-		can_author_with: sp_consensus::CanAuthorWithNativeVersion::new(client.executor().clone()),
+		can_author_with: sp_consensus::AlwaysCanAuthor,
```
wigy-opensource-developer, line 1478: This was a bit weird for me, but @gilescope explained to me that the on-chain runtime is always up-to-date (hopefully), at least there is no reason why a node would assume it is out-of-date, therefore a new block can always be authored with it.

polkadot-parachains/src/service.rs
```diff
@@ -226,10 +230,11 @@ where
...
 		.transpose()?;
 
-	let executor = sc_executor::NativeElseWasmExecutor::<Executor>::new(
+	let executor = sc_executor::WasmExecutor::<HostFunctions>::new(
 		config.wasm_method,
 		config.default_heap_pages,
 		config.max_runtime_instances,
+		None,
 		config.runtime_cache_size,
 	);
```
wigy-opensource-developer, line 239: This seems to be the most important change, the rest is just a consequence.

wigy-opensource-developer (approved):

polkadot-parachains/src/service.rs
bkchr: The on-chain runtime is the canonical runtime, you don't need to hope that it is up-to-date. There is no way that this isn't up-to-date.

This `can_author_with` feature was added back in the days that when you break the authoring interface, that you have a nicer way of telling the node that it is outdated. However, we never really needed it and when the native runtime is removed it will be gone anyway. We may need to find another way or just ignore that and let the block building fail.

polkadot-parachains/src/service.rs
gilescope: Just taking out a level of indirection by always taking the same branch that was there before.

skunert (approved):

polkadot-parachains/src/service.rs
bkchr: What?

gilescope:
bot merge

Issue #920, opened 2022-01-19
Teleporting to Statemine is Oddly Expensive


joepetrowski:
Reported in https://github.com/polkadot-js/apps/issues/6742#issuecomment-1001986546 as a UI issue.

Fees on the sending side make sense - Kusama is 10x more expensive than Statemine. Fees on the receiving side, however, do not (0.004 on Statemine vs. 0.000106 on Kusama, a factor of almost 40x).

Either there is an error in the XCM weights/benchmarking on the Statemine side, or there is a logical explanation for it. But we need to figure out why there is a 40x difference and either fix it (if broken) or document it (if it's supposed to be that way).

Image pasted for convenience:

![image](https://user-images.githubusercontent.com/25483142/150172477-f4f5e24a-ae67-457e-bb81-7fa9c696330b.png)

![image](https://user-images.githubusercontent.com/25483142/150172505-cef7cac6-bf12-40e5-b8d0-0f5f4c243372.png)

hbulgarini:
To check if this gh issue about benchmarking is related: https://github.com/paritytech/ci_cd/issues/260

gilescope:
The cost is the weight for the downward message:
dmpQueue.ExecutedDownward
Downward message executed with the given outcome. XcmV2TraitsOutcome
{
  Complete: 4,000,000,000
}

KiChjang:
One thing that we would want to definitely check is the exact version of the code (both for Kusama and Statemine) that it was running. Since this was a while back, the weight calculation logic may have been different from master.

gilescope:
"since the destination is what determines the weight (and thereby the fees as well). " - The destination does not determine the fees applied at the source chain (kusama) for sending the message and that is where the high charge is. It is the enqueue of the message that incurs the high cost. As such its a benchmark on the kusama relay chain that will be needed to make the weights less conservative for this particular message type.

joepetrowski:
> "since the destination is what determines the weight (and thereby the fees as well). " - The destination does not determine the fees applied at the source chain (kusama) for sending the message and that is where the high charge is. It is the enqueue of the message that incurs the high cost. As such its a benchmark on the kusama relay chain that will be needed to make the weights less conservative for this particular message type.

No, it's the other way around.

NachoPal:
[...]

PR #1055, opened 2022-03-01
Use WeightToFee when calculating cost of each weight unit in XCM trader

KiChjang:
This PR aims to solve #920.

We change the `Trader`'s `WeightToFee` type parameter to each chain's `WeightToFee` struct that properly calculates and converts weight units into fees. Previously, it would do a 1-to-1 conversion between weight units and plancks.

### DIFF
--- a/polkadot-parachains/statemine/src/xcm_config.rs
+++ b/polkadot-parachains/statemine/src/xcm_config.rs
@@ -23,6 +23,7 @@ use frame_support::{
 	weights::Weight,
 };
 use pallet_xcm::XcmPassthrough;
+use parachains_common::impls::ToStakingPot;
 use polkadot_parachain::primitives::Sibling;
 use xcm::latest::prelude::*;
 use xcm_builder::{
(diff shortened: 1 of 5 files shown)


xlc (commented):
Some tests to ensure the numbers are sane?

polkadot-parachains/statemine/src/xcm_config.rs
```diff
@@ -162,7 +162,7 @@ impl xcm_executor::Config for XcmConfig {
 	type LocationInverter = LocationInverter<Ancestry>;
 	type Barrier = Barrier;
 	type Weigher = FixedWeightBounds<UnitWeightCost, Call, MaxInstructions>;
-	type Trader = UsingComponents<IdentityFee<Balance>, KsmLocation, AccountId, Balances, ()>;
+	type Trader = UsingComponents<WeightToFee, KsmLocation, AccountId, Balances, ()>;
```
KiChjang, line 165: One of the things that's still unresolved to me is where the extra weight should go when even the `refund_weight` method fails to refund all the surplus weight. Currently as we can see here, it is set to `()`, which means it all just gets dropped and forgotten, but in Polkadot, it gets refunded to the block author via the `ToAuthor` struct.

I see that in Cumulus, we have a `ToStakingPot` struct that sends nonzero balances to the staking pot, and perhaps this is what we should use?

gilescope (approved):
Looks good. We will deploy it and test it out on the test nets.

Issue #631, opened 2021-09-29
[CI/CD] Add benchmarking CI job


apopiak:
It would be great to have a manual job to execute runtime benchmarks we can start from gitlab that spits out a git patch of the adjusted weights we can apply afterwards.
Analogous to the benchmarking CI in the Polkadot repo.

NachoPal:
Is it work in progress? I need to update the Benchmark section in #661 and I am not sure if this is going to be done soon or we keep the manual process for now.

sea212:
Maybe https://github.com/zeitgeistpm/zeitgeist/blob/main/.github/workflows/benchmark.yml can become handy. 
Once "benchmark-required" label is set on a PR, the CI is triggered and pulls the branch. It benchmarks currently every pallet and the pallets are hard-coded, but this will be optimized in the future. A report is created and posted in the PR and the changes are committed, if no other changes have been added meanwhile. The CI indicates with a label that it's done and whether the benchmark was successful.
The following link references setting the label and the resulting actions from the benchmark (expand following comments): https://github.com/zeitgeistpm/zeitgeist/pull/207#event-4958808872

PR #1074, opened 2022-03-09
Add benchmarks to pipeline

alvicsam:
PR adds manual benchmarks job. It's divided to 2 jobs: `benchmarks-build` which builds binary on a faster ci runners and passes to `benchmarks` job which execute benchmarks. 
The `benchmarks` job runs script `benchmarks-ci.sh` for statemine, statemint and westmint and creates PR on github with updated weights (example: https://github.com/paritytech/cumulus/pull/1138)

Part of https://github.com/paritytech/ci_cd/issues/260
Closes https://github.com/paritytech/cumulus/issues/631


.gitlab-ci.yml
```diff
@@ -8,8 +8,19 @@
 stages:
   - test
   - publish
+  - benchmarks-build
+  - benchmarks
```
TriplEight, line 12:
```suggestion
  - benchmarks-run
```

.gitlab-ci.yml
```diff
@@ -172,3 +193,47 @@ publish-s3:
...
+  <<:                              *docker-env
+  <<:                              *collect-artifacts
+  <<:                              *benchmarks-refs
+  script:
+    - time cargo build --profile production --locked --features runtime-benchmarks
+    - mkdir artifacts
+    - cp $CARGO_TARGET_DIR/production/polkadot-collator ./artifacts/
+
+benchmarks-statemine:
+  stage:                           benchmarks
```
TriplEight, line 210:
```suggestion
  stage:                           benchmarks-run
```

TriplEight (changes requested):

TriplEight (approved):

alvicsam:
There is an additional condition that should be implemented (after discussing it with @NachoPal in element):

Weights should be pushed to the same branch where the binary for benchmarks was compiled.

Example:

1. Select branch release-parachains-vX.Y.Z -> Click Run benchmarks button
2. The binary is built, benchmarks are run.
3. When the new weights are ready, a PR is open against release-parachains-vX.Y.Z
4. We review the PR and merge it to release-parachains-vX.Y.Z

NachoPal (changes requested):
Thanks @alvicsam . I realised that we'll need an input to select the branch we want to use for building the binary, running the benchmarks and opening a PR with the new weights.

The flow would be as follows:
1. Select branch release-parachains-vX.Y.Z -> Click Run benchmarks button
2. The binary is built, benchmarks are run.
3. When the new weights are ready, a PR is open against release-parachains-vX.Y.Z
4. We review the PR and merge it to release-parachains-vX.Y.Z

NachoPal:
What Benchmarking machine is used to run the benchmarks?

alvicsam:
`bm4`, `bm5`, `bm6`

PR #1076, opened 2022-03-09
Co #5064: Use runtime dependant weights

ggwpez:
Companion for https://github.com/paritytech/polkadot/pull/5064

- Use the weights from `polkadot-runtime` instead of `polkadot-runtime-common` since the weights are now not *common* anymore but depend on the runtime.


ggwpez:
bot rebase

kianenigma (approved):

polkadot-parachains/canvas-kusama/Cargo.toml
```diff
@@ -58,7 +58,7 @@ kusama-runtime-constants = { git = "https://github.com/paritytech/polkadot", def
 pallet-xcm = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "master" }
 polkadot-core-primitives = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "master" }
 polkadot-parachain = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "master" }
-polkadot-runtime-common = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "master" }
+polkadot-runtime = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "master" }
```
bkchr, line 61: Wait wait. Sorry that I'm late to the party. But, this is `canvas-kusama`. Why did you even removed the `runtime-common`? 

Because the `RocksDbWeight` is now different per runtime?

It is really bad that include here the entire runtime into another runtime. Especially as `RocksDbWeight` should be defined here in each runtime and should not be taken from Polkadot.

ggwpez: The `polkadot-runtime-common` does not contain the `RocksDbWeight` now anymore, I did not think that it is a bad idea to include one runtime into another.  

So I should rather also copy the weights into all the parachain configs as well? Makes much more sense when I write it out like this :facepalm:

bkchr: Good :D

bkchr: I mean you now automatically generate this weight? So, this weight could also be generated for the parachains?

ggwpez: Theoretically it should work. I just never tried to run the collators myself.  
Also not sure if their CLI uses the command yet.  

So I fixed the parachain weights in https://github.com/paritytech/polkadot/pull/5091 and https://github.com/paritytech/cumulus/pull/1081, but I'm a bit tired now, so hope this all makes sense.

bkchr: Yeah that looks like a good start. However, every substrate based chain should generate these weights on their own. The same applies to all the other parachains. So, you should make sure that they also can generate these weights. It makes no sense to only create them for polkadot.

ggwpez: Yes sure that was the long term plan. For now this just fixes the dependency on the Polkadot runtime.  
In the future the Storage and Block/Extrinsic weights should all be generated by the new bench bot.  
Can we merge the mentioned MRs or do you need something else for that? I was under the impression that this should be fixes ASAP.

bkchr: Okay, yeah I will review them now :)

PR #1078, opened 2022-03-11
service_overweight not operational dispatch

gilescope:
This puts the dispatch level back to what it was for service_overweight.
I mistakenly added it in https://github.com/paritytech/cumulus/pull/982
(The other comments mentioned in the PR will be fixed in a separate PR as those relate to improvements/bugfixes of existing code)

### DIFF
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -130,7 +130,7 @@ pub mod pallet {
 		///
 		/// Events:
 		/// - `OverweightServiced`: On success.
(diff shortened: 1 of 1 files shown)


bkchr:
But isn't that an operational transaction? You need to have some special origin to call this and it is about executing overweights messages. Sounds to me like some operational transaction.

gilescope:
Maybe I misunderstood your comment in https://github.com/paritytech/cumulus/pull/982 about service_overweight? If we're happy that it is an operational dispatch then master is fine and I can close this.

bkchr:
My comment as the comment of @KiChjang as about the missing benchmark.

PR #1084, opened 2022-03-14
--json-file param is nicer.

gilescope:
As suggested by comment [here](https://github.com/paritytech/cumulus/pull/1012): to use `--json-file`

### DIFF
--- a/scripts/benchmarks.sh
+++ b/scripts/benchmarks.sh
@@ -35,7 +35,7 @@ do
 		--extrinsic='*' \
 		--steps=$steps  \
 		--repeat=$repeat \
-		--json \
+		--json-file=./bench-statemine.json \
         --header=./file_header.txt \
 		--output=$statemineOutput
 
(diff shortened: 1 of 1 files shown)


scripts/benchmarks.sh
```diff
@@ -35,7 +35,7 @@ do
 		--extrinsic='*' \
 		--steps=$steps  \
 		--repeat=$repeat \
-		--json \
+		--json-file=./bench-statemine.json \
         --header=./file_header.txt \
```
ggwpez, line 39: Something with the tabs/spaces indentation here is wrong.

scripts/benchmarks.sh
```diff
@@ -35,7 +35,7 @@ do
 		--extrinsic='*' \
 		--steps=$steps  \
 		--repeat=$repeat \
-		--json \
+		--json-file=./bench-statemine.json \
```
ggwpez, line 38: If you do not plan on analyzing the resulting json files, you can also kick it out completely.

ggwpez (approved):

scripts/benchmarks.sh
gilescope: good spot. fixed.

scripts/benchmarks.sh
gilescope: They won't take any time to write out and if you did want to have a look at them afterwards you don't really want to re-run the benchmarks as this could take a long time.

PR #1092, opened 2022-03-15
companion #5131 (libp2p-0.43.0)

kpp:
https://github.com/paritytech/polkadot/pull/5131


kpp:
https://github.com/paritytech/polkadot/pull/5131 works without this companion PR.

PR #1099, opened 2022-03-21
xcmp swap - express intent better

gilescope:
`swap` is defined on slice so we can use it here. (`mem::swap` would complain about borrowing the same thing)


KiChjang (approved):

wigy-opensource-developer (approved):

gilescope:
bot merge

PR #1100, opened 2022-03-21
xcmp-queue / less indexing

gilescope:
No need to use indexes in a couple of places.
(Cleared up a couple of clippies while in the area.)

### DIFF
--- a/pallets/xcmp-queue/src/lib.rs
+++ b/pallets/xcmp-queue/src/lib.rs
@@ -499,7 +499,7 @@ impl<T: Config> Pallet<T> {
 			details
 		} else {
 			s.push(OutboundChannelDetails::new(recipient));
-			s.last_mut().expect("can't be empty as just pushed")
+			s.last_mut().expect("can't be empty; a new element was just pushed; qed")
 		};
 		let have_active = details.last_index > details.first_index;
 		let appended = have_active &&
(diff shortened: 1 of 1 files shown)


wigy-opensource-developer (approved):

pallets/xcmp-queue/src/lib.rs
```diff
@@ -495,13 +495,15 @@ impl<T: Config> Pallet<T> {
...
 		let mut s = <OutboundXcmpStatus<T>>::get();
-		let index = s.iter().position(|item| item.recipient == recipient).unwrap_or_else(|| {
+		let details = if let Some(details) = s.iter_mut().find(|item| item.recipient == recipient) {
+			details
+		} else {
 			s.push(OutboundChannelDetails::new(recipient));
-			s.len() - 1
-		});
-		let have_active = s[index].last_index > s[index].first_index;
+			s.last_mut().expect("can't be empty as just pushed")
```
KiChjang, line 502:
```suggestion
			s.last_mut().expect("can't be empty; a new element was just pushed; qed")
```

The QED is important there, and I'm only half-joking.

wigy-opensource-developer: Okay, I am then half-laughing. Is it a cultural thing at Parity, or do we have a more formal description on what abstraction level to write `expect("...; qed")` proofs? :wink:

gilescope: Idk but we live in a world where there's not enough "qed"s about so I'm going with it.

PR #1114, opened 2022-03-27
Regenesis rococo

brenzi:
On rococo we had an issue producing blocks because genesis demanded specVersion 0, but we actually wanted to launch with specVersion 1 as the runtime api has changed since 0

with this, we re-genesis rococo spec


brenzi:
wrong repo

PR #1120, opened 2022-03-29
Update weights from bm2 for v800

gilescope:
Couple of things that pop out to me:

- remark now has some weight associated with it (again). I think either is fine as most of the remark weight is done via the tx length.
- set_invulnerables seems to have an extra read weight.


gilescope:
The set_invulnerables weight change is down to this change: https://github.com/paritytech/cumulus/commit/b2f663475fcd18f79e4a112105e8999559e452f0

NachoPal (commented):
Numbers are lower in general, but that's ok. Maybe more suspicious the case of `kill_storage` and `frame_system`

NachoPal:
Seems that it's been `0` before and right now is `0` in Polkadot weights also. It should be fine.

LGTM

NachoPal (approved):

hbulgarini (approved):
LGTM

gilescope:
bot merge

PR #1126, opened 2022-03-31
merge updated weights from v800 back to master (#1120)

gilescope:
For completeness merging the weights back from v800 to master.
There's no new migrations we need to remove. There's nothing else on the v800 branch that we need to backport to master.


shawntabrizi:
bot merge

PR #1083, opened 2022-03-14
Add babe randomness to relay state proof

4meta5:
cc @bkchr This is steps 2-4 for https://github.com/paritytech/cumulus/issues/463#issuecomment-1054559103

PR #1135, opened 2022-04-04
Make functionality to read relay state proof entries public

4meta5:
Follow up for #1083 which makes the functionality to read arbitrary entries in the relay state proof public. We need this to read the BABE randomness values from the proof.

Let me know if there is a better way to do this @bkchr


pallets/parachain-system/src/relay_state_snapshot.rs
```diff
@@ -273,4 +277,18 @@ impl RelayChainStateProof {
 		)
 		.map_err(Error::UpgradeRestriction)
 	}
+
+	pub fn read_entry<T>(&self, key: &[u8], fallback: Option<T>) -> Result<T, Error>
```
bkchr, line 281: Why not directly expose the functions to read the babe randomness? As we do for all the other values?

4meta5: Then we will need functions for every value we would like to get. This will be at least:
```
read_current_block_randomness
read_one_epoch_ago_randomness
read_two_epochs_ago_randomness
read_epoch_index
```

It is less code to write one `read_entry<T>` which is generic over the successful result value returned.

bkchr: I get what you say, but currently it also requires that you add these values before on the client side :P 

We can add this function, I don't really care, but then please add some docs :)

bkchr:
@4meta5 can you please merge master

pallets/parachain-system/src/relay_state_snapshot.rs
```diff
@@ -273,4 +277,29 @@ impl RelayChainStateProof {
...
+	/// chain state and no fallback was provided.
+	pub fn read_entry<T>(&self, key: &[u8], fallback: Option<T>) -> Result<T, Error>
+	where
+		T: Decode,
+	{
+		read_entry(&self.trie_backend, key, fallback).map_err(Error::ReadEntry)
+	}
+
+	/// Read an optional entry given by the key and try to decode it.
+	/// Returns `None` if the value specified by the key according to the proof is empty.
```
bkchr, line 295:
```suggestion
```

bkchr (approved):

PR #1144, opened 2022-04-05
Don't allow reserve transfers for the relay chain (as this can be trusted teleported to)

gilescope:
Potential fix for https://github.com/paritytech/polkadot/issues/5233

Maybe the definition should be in parachains-common...

### DIFF
--- a/polkadot-parachains/canvas-kusama/src/xcm_config.rs
+++ b/polkadot-parachains/canvas-kusama/src/xcm_config.rs
@@ -19,7 +19,7 @@ use super::{
 };
 use frame_support::{
 	match_types, parameter_types,
-	traits::{EnsureOneOf, Everything, Nothing},
+	traits::{EnsureOneOf, Everything, EverythingBut, Nothing},
 	weights::Weight,
 };
 use frame_system::EnsureRoot;
@@ -117,6 +117,12 @@ match_types! {
 	};
 }
 
+match_types! {
+	pub type IsParent: impl Contains<(MultiLocation, crate::Vec<MultiAsset>)> = {
+		(MultiLocation { parents: 1, interior: Here }, _)
+	};
+}
+
 pub type Barrier = (
 	TakeWeightCredit,
 	AllowTopLevelPaidExecutionFrom<Everything>,
@@ -170,7 +176,7 @@ impl pallet_xcm::Config for Runtime {
 	type XcmExecuteFilter = Nothing;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
 	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Everything;
+	type XcmReserveTransferFilter = EverythingBut<IsParent>;
 	type Weigher = FixedWeightBounds<UnitWeightCost, Call, MaxInstructions>;
 	type LocationInverter = LocationInverter<Ancestry>;
 	type Origin = Origin;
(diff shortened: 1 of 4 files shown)


polkadot-parachains/canvas-kusama/src/xcm_config.rs
```diff
@@ -170,7 +176,7 @@ impl pallet_xcm::Config for Runtime {
 	type XcmExecuteFilter = Nothing;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
 	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Everything;
+	type XcmReserveTransferFilter = EverythingBut<IsParent>;
```
joepetrowski, line 179: I would add some docs here explaining why, like:
```suggestion
	/// Statemine acts as a `NativeAsset` reserve to all chains except the Relay Chain, which
	/// recognizes no reserves.
	type XcmReserveTransferFilter = EverythingBut<IsParent>;
```

polkadot-parachains/canvas-kusama/src/xcm_config.rs
```diff
@@ -170,7 +176,7 @@ impl pallet_xcm::Config for Runtime {
 	type XcmExecuteFilter = Nothing;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
 	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Everything;
+	type XcmReserveTransferFilter = Nothing;
```
HCastano, line 179: Okay so confirm the intention of this change, this will mean that anybody will still be able to `Teleport` any asset in and out of Canvas, but will **not** be able to perform `RevervedTransfer`s, correct?

HCastano, line 179: Okay so confirm the intention of this change, this will mean that anybody will still be able to `Teleport` any asset in and out of Canvas, but will **not** be able to perform `RevervedTransfer`s, correct?

joepetrowski: They should probably both be set to `Nothing` at genesis.

HCastano: If we set both to `Nothing` then users won't be able to get `ROC` tokens on Canvas to pay for fees. So for `Teleport` maybe it should be `EverythingBut<IsParent>` instead?

joepetrowski: Oh I didn't know this was what is used on Rococo. Then yes.

KiChjang (approved):

KiChjang:
Do remember that this method doesn't actually prevent users from creating a `DepositReserveAsset` or `TransferReserveAsset` XCM instruction and executing it via the `execute` dispatchable on the XCM pallet -- all this PR does is limit the usage of the `reserve_asset_transfer` dispatchable on the XCM pallet.

NachoPal:
> Do remember that this method doesn't actually prevent users from creating a `DepositReserveAsset` or `TransferReserveAsset` XCM instruction and executing it via the `execute` dispatchable on the XCM pallet -- all this PR does is limit the usage of the `reserve_asset_transfer` dispatchable on the XCM pallet.

I should be fine also. `XcmExecuteFilter` is set to `Nothing`

gilescope:
Problem: Tested it and it the other way around. It's the source rather than the dest that gets passed into the filter and you can't tell from the source that the dest is the relay chain. I think dest and beneficiary ought to be added to the filter arguments (as part of XCM3?).

gilescope:
closing this in favour of Barrier

Issue #1128, opened 2022-03-31
Provide binary for test service 


bkchr:
The test service is currently only used by integration tests. There is a builder pattern for building a node with the wanted features. Most of these integration tests should be rewritten with Zombienet. This will require a binary that can be started. This binary should basically expose the builder pattern via cli.

The builder currently also uses in memory addresses for libp2p. This will not work when the nodes are their own processes. So, this needs to be changed as well to also support normal libp2p multi addresses.

Cc @pepoviola

skunert:
@bkchr  That means we also need the polkadot test service as new binary right? To start the relay chain nodes for these integration tests.

bkchr:
Probably yes, but maybe we can also just use rococo as runtime. Needs to be checked.

skunert:
@pepoviola I have draft open for the testing binary #1158.

Mostly the node remains the same from what we have in our current integration tests. From what I understand you will migrate the first test for us right?

Mostly I exposed networking configuration like node-key, so we can do address calculation in zombienet.
```
--parachain-id
--relay-chain-bootnodes
--alice, bob, ...
--reserved-nodes
--reserved-only
--node-key
--bootnodes

// following arguments are used to build specific scenarios for the tests
--use-null-consensus
--disabled-block-announcements
```
Will that be enough for zombienet? Let met know here if you need more options on it.

As for the relay chain nodes, will look into using rococo.

PR #1158, opened 2022-04-08
Create binary for cumulus testing node #1128

skunert:
In this PR I create a simple binary for a cumulus testing node. This is a draft until we have checked that this is a good fit for testing with zombienet.

Supported cli args:
```
--parachain-id
--relay-chain-bootnodes
--alice, bob, ...
--reserved-nodes
--reserved-only
--node-key
--bootnodes
--relay-chain-bootnodes
--collator
--relay-chain-spec
--port

export-genesis-wasm
export-genesis-state

// following arguments are used to build specific scenarios for the tests
--use-null-consensus
--disabled-block-announcements
```


skunert:
I changed the PR to not use the builder that is mentioned in the issue. 
Reasoning:
- We now have the same CLI as the collator
- Configuration is handled in a way that is close to production

The CLI is copy pasted from the original collator in `polkadot-parachains`.

test/service/src/lib.rs
```diff
@@ -398,7 +398,8 @@ pub struct TestNode {
 	pub transaction_pool: TransactionPool,
 }
 
-enum Consensus {
+#[allow(missing_docs)]
+pub enum Consensus {
```
wigy-opensource-developer, line 402: `pub(crate)` might be what you are looking for here? Then you do not need to document it either.

skunert: Hmm yes but I am using this in a binary which can afaik only access pub items. I would need to create some other `main()` in a  and call it from the binary. Is that the usual way to do it?

wigy-opensource-developer: Right, sorry, I have overlooked something because you seemed to be even in the same module based on the imports. Never had 2 crates in the same folder before and thought for some reason they were the same crate. But according to the [Rust documentation](https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html):

> If a package contains src/main.rs and src/lib.rs, it has two crates: a binary and a library, both with the same name as the package.

So the `lib.rs` is imported from a different crate with the same name.

koute:
The cargo fmt job is failing. (:

So I have a few questions:
- Do we already have any tests (perhaps on a separate branch somewhere) which will use this new binary?
- Are you looking to merge this in before the test are written, or after?
- Since the CLI closely matches the production binary anyway can you explain why this new binary is necessary? Can't we run the zombienet tests with normal cumulus binary? I see that you've added e.g. `--use-null-consensus` so I suppose those tests that require that functionality can't be run with the normal binary, but what about the rest? Is the intent to use this binary for all of zombienet tests, or for only a few of them?

skunert:
I should have probably given a bit more context. 

We are planning to migrate the cumulus integration tests to zombienet. Zombienet requires the binary. @pepoviola is working on zombienet and agreed to help us migrate the integration tests once I implemented the binary for him. My original plan was to merge this and have a separate PR for the tests themselves. For a short period, the binary would be unused. We could also leave this PR open and have the test migration here, but this comes at the cost of maintaining this PR.

The extra binary is to have a clear separation between the testing infrastructure and the prod binary. They are similar, but the testing binary offers access to the test runtime and the extra arguments you mentioned (we are free to add more specific scenarios easily to the test-collator). We intend to use this test binary for all zombienet tests in cumulus. At some point, we could probably even remove the builder pattern for the old integration tests and have this test-collator only.

pepoviola:
Hi, I already have a couple of integration test ready to add in this `pr` if you want. We will need also to build an image with the `test-collator` to use in gitlab ci. I can also add that step to the current setup.

Thanks!

test/service/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+		.remove(sp_core::storage::well_known_keys::CODE)
+		.ok_or_else(|| "Could not find wasm file in genesis state!".into())
+}
+
+pub fn wrap_announce_block() -> Box<dyn FnOnce(AnnounceBlockFn) -> AnnounceBlockFn> {
+	Box::new(|_| {
+		// Never announce any block
+		Arc::new(|_, _| {})
+	})
+}
```
koute, line 43: This is not used anywhere; I'm assuming you'll add a command line argument for this, right? (For the `pov_recovery` test.)

koute:
> I should have probably given a bit more context.
> 
> We are planning to migrate the cumulus integration tests to zombienet. Zombienet requires the binary. @pepoviola is working on zombienet and agreed to help us migrate the integration tests once I implemented the binary for him. My original plan was to merge this and have a separate PR for the tests themselves. For a short period, the binary would be unused. We could also leave this PR open and have the test migration here, but this comes at the cost of maintaining this PR.
> 
> The extra binary is to have a clear separation between the testing infrastructure and the prod binary. They are similar, but the testing binary offers access to the test runtime and the extra arguments you mentioned (we are free to add more specific scenarios easily to the test-collator). We intend to use this test binary for all zombienet tests in cumulus. At some point, we could probably even remove the builder pattern for the old integration tests and have this test-collator only.

I see!

I'd personally would probably prefer if we'd use the production binary for those tests that actually can just use the production binary since, you know, over time the behavior between the production binary and the test binary might go out of sync and then we might not be testing what is actually relevant. We could also consider (assuming the changes are unintrusive enough, and I don't know if they'd be) adding an extra cargo opt-in feature for those special test-only args that the tests need and just use the production binary for all of them? But I know practical logistics of the whole thing might make this infeasible and/or a bad idea (and I'm not the one implementing this) so I'll leave this decision to you. (:

koute (approved):
It mostly looks fine to me. Obviously it's still incomplete, but since this is purely for tests I think it's fine to just merge it and iterate on it instead of just waiting until everything's really and do it in one huge PR.

test/service/src/cli.rs
```diff
@@ -0,0 +1,343 @@
...
+	#[clap(name = "export-genesis-state")]
+	ExportGenesisState(ExportGenesisStateCommand),
+
+	/// Build a chain specification.
+	BuildSpec(sc_cli::BuildSpecCmd),
+}
+
+#[derive(Debug, Parser)]
+#[clap(
+	propagate_version = true,
```
davxy, line 76:
```suggestion
	version,
	propagate_version = true,
```

Without this the application doesn't have the `version` option, thus can't use the `propagate_version`. 
Indeed it panics it the binary is built without `--release` flag

davxy:
I've not enough experience with zombienet testing to give a strong opinion WRT the final application of this PR... but once the  missing "version" thing is fixed, from a formal pow it LGTM

bkchr:
I'm fine to merge this first and then have the test enabled later

test/service/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+			let tokio_handle = tokio_runtime.handle();
+			let mut config = cli
+				.run
+				.normalize()
+				.create_configuration(&cli, tokio_handle.clone())
+				.expect("Should be able to generate config");
+
+			let parachain_id = ParaId::from(cli.parachain_id);
+			config.chain_spec =
+				Box::new(cumulus_test_service::get_chain_spec(parachain_id)) as Box<_>;
```
bkchr, line 116: Why do we need this? The `chain_spec` should already be the correct one through the `load_spec` above?

test/service/src/main.rs
```diff
@@ -0,0 +1,192 @@
...
+				.unwrap_or(cumulus_test_service::Consensus::RelayChain);
+
+			let block_announce_wrapper: Option<
+				Box<dyn FnOnce(AnnounceBlockFn) -> AnnounceBlockFn>,
+			> = if cli.disable_block_announcements {
+				tracing::info!("Block announcements disabled.");
+				Some(Box::new(|_| {
+					// Never announce any block
+					Arc::new(|_, _| {})
+				}))
```
bkchr, line 168: Why don't you use your `wrap_block_announce` function above?

test/service/src/main.rs
```diff
@@ -0,0 +1,192 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
bkchr, line 1:
```suggestion
// Copyright 2022 Parity Technologies (UK) Ltd.
```

test/service/src/cli.rs
davxy: From what I experienced, other binaries (`polkadot-collator`) doesn't have the same problem.

I think because it is using a custom build script to manually define the version (that is augmented with the commit hash:

https://github.com/paritytech/cumulus/blob/a5bfd11b5f16abde954bf204ad2de70424de0f59/polkadot-parachains/build.rs#L20

And 

https://github.com/paritytech/substrate/blob/d45e3837bc89338c6eae1ff5dc250d01e1cae0e6/utils/build-script-utils/src/version.rs#L21

skunert: Ah I see, was not aware of that. I left it only for the test binary, should be fine now

PR #1169, opened 2022-04-12
Deny using relay chain for reserve transfers

gilescope:
Fixes [#5233](https://github.com/paritytech/polkadot/issues/5233)

I will move DenyThenTry to polkadot xcm and then reference it from there in a later release, but we'd like to ship this for v900.


polkadot-parachains/parachains-common/src/lib.rs
```diff
@@ -109,3 +116,56 @@ pub mod opaque {
...
+		_origin: &MultiLocation,
+		message: &mut Xcm<Call>,
+		_max_weight: Weight,
+		_weight_credit: &mut Weight,
+	) -> Result<(), ()> {
+		if message.0.iter().any(|inst| {
+			matches!(
+				inst,
+				DepositReserveAsset { dest: MultiLocation { parents: 1, .. }, .. } |
+					TransferReserveAsset { dest: MultiLocation { parents: 1, .. }, .. }
```
KiChjang, line 163: You need to also specify the junction as well, otherwise you'd block other parachain destinations as well.

KiChjang: Also, this seems to be denying XCMs that is trying to _send_ these two instructions to the relay chain, but it doesn't seem to restrict _executing_ these two kinds of instruction locally and then sending a `ReserveAssetDeposited` instruction onwards to the relay. I think the match condition here would thus need to be expanded to include that case.

polkadot-parachains/parachains-common/src/lib.rs
```diff
@@ -109,3 +116,56 @@ pub mod opaque {
...
+//TODO: move DenyThenTry to polkadot's xcm module.
+/// Deny executing the xcm message if it matches any of the Deny filter regardless of anything else.
+/// If it passes the Deny, and matches one of the Allow cases then it is let through.
+pub struct DenyThenTry<Deny, Allow>
+where
+	Deny: ShouldExecute,
+	Allow: ShouldExecute,
+{
+	_deny: PhantomData<Deny>,
+	_allow: PhantomData<Allow>,
```
KiChjang, line 129: Why do these need to be named fields?

polkadot-parachains/parachains-common/src/lib.rs
```diff
@@ -109,3 +116,56 @@ pub mod opaque {
...
+	Deny: ShouldExecute,
+	Allow: ShouldExecute,
+{
+	fn should_execute<Call>(
+		origin: &MultiLocation,
+		message: &mut Xcm<Call>,
+		max_weight: Weight,
+		weight_credit: &mut Weight,
+	) -> Result<(), ()> {
+		if Deny::should_execute(origin, message, max_weight, weight_credit).is_ok() {
```
KiChjang, line 143: The name of this function is `should_execute`, and so when it returns an Ok, I expect it to be able to execute, not be denied or blocked.

polkadot-parachains/parachains-common/src/lib.rs
```diff
@@ -17,8 +17,15 @@
 
 pub mod impls;
 pub use constants::*;
+use core::marker::PhantomData;
+use frame_support::weights::Weight;
 pub use opaque::*;
 pub use types::*;
+use xcm::{
+	opaque::v2::prelude::{DepositReserveAsset, TransferReserveAsset},
```
KiChjang, line 25: Just import `xcm::latest::prelude::*` instead of importing types one by one.

polkadot-parachains/parachains-common/src/lib.rs
```diff
@@ -109,3 +116,56 @@ pub mod opaque {
...
+	) -> Result<(), ()> {
+		if Deny::should_execute(origin, message, max_weight, weight_credit).is_ok() {
+			return Err(())
+		}
+		Allow::should_execute(origin, message, max_weight, weight_credit)
+	}
+}
+
+// See issue #5233
+pub struct IsReserveTransferToRelayChain;
```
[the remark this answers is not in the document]
gilescope: I guess no one would use that filter on the Allow side probably (though it would work on that side too).

polkadot-parachains/parachains-common/src/lib.rs
gilescope: all hell breaks loose if I do that. I think it upsets the other imports in the file. Maybe I should break it out into its own file...

polkadot-parachains/parachains-common/src/lib.rs
gilescope: You want me to create a ShouldNotExecute trait?

polkadot-parachains/parachains-common/src/lib.rs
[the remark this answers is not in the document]
gilescope: renamed

polkadot-parachains/parachains-common/src/lib.rs
gilescope: good point. now a tuple.

polkadot-parachains/parachains-common/src/lib.rs
KiChjang: No, I think `should_execute` should return an `Err` when it judges that the XCM should be denied from execution.

polkadot-parachains/parachains-common/src/xcm_config.rs
```diff
@@ -0,0 +1,58 @@
...
+		if message.0.iter().any(|inst| {
+			matches!(
+				inst,
+				DepositReserveAsset {
+					dest: MultiLocation { parents: 1, interior: Junctions::Here },
+					..
+				} | TransferReserveAsset {
+					dest: MultiLocation { parents: 1, interior: Junctions::Here },
+					..
+				}
```
KiChjang, line 50: > 

Also, this seems to be denying XCMs that is trying to send these two instructions to the relay chain, but it doesn't seem to restrict executing these two kinds of instruction locally and then sending a ReserveAssetDeposited instruction onwards to the relay. I think the match condition here would thus need to be expanded to include that case.

Still needs to be addressed.

polkadot-parachains/parachains-common/src/xcm_config.rs
```diff
@@ -0,0 +1,58 @@
...
+		_weight_credit: &mut Weight,
+	) -> Result<(), ()> {
+		if message.0.iter().any(|inst| {
+			matches!(
+				inst,
+				DepositReserveAsset {
+					dest: MultiLocation { parents: 1, interior: Junctions::Here },
+					..
+				} | TransferReserveAsset {
+					dest: MultiLocation { parents: 1, interior: Junctions::Here },
```
KiChjang, line 48:
```suggestion
					dest: MultiLocation { parents: 1, interior: Here },
```

polkadot-parachains/parachains-common/src/xcm_config.rs
```diff
@@ -0,0 +1,58 @@
...
+		_origin: &MultiLocation,
+		message: &mut Xcm<Call>,
+		_max_weight: Weight,
+		_weight_credit: &mut Weight,
+	) -> Result<(), ()> {
+		if message.0.iter().any(|inst| {
+			matches!(
+				inst,
+				DepositReserveAsset {
+					dest: MultiLocation { parents: 1, interior: Junctions::Here },
```
KiChjang, line 45:
```suggestion
					dest: MultiLocation { parents: 1, interior: Here },
```

NukeManDan:
DumbQ: Any action needed for the template in this PR?

NachoPal:
> DumbQ: Any action needed for the template in this PR?

Considering that the template will be used mostly by future Parachains, I would say yes, it is a good idea to reject by default `DepositReserveAsset` and `TransferReserveAsset` if the destination is the `MultiLocation { parents: 1, interior: Here }`

The proper way to reserve transfer funds back to the reserve is by making use of `InitiateReserveWithdraw`

Unless somebody disagrees, can you @gilescope add also the change to the `parachain-template` runtime?

polkadot-parachains/parachains-common/src/lib.rs
gilescope: Should I be looking for a Transact with a ReserveAssetDeposited somewhere inside the encoded call? or do you mean a InitiateReserveWithdraw with a ReserveAssetDeposited inside it's xcm field?

gilescope: Have updated, does that look like what you were expecting?

polkadot-parachains/parachains-common/src/xcm_config.rs
```diff
@@ -0,0 +1,60 @@
...
+		_max_weight: Weight,
+		_weight_credit: &mut Weight,
+	) -> Result<(), ()> {
+		if message.0.iter().any(|inst| {
+			matches!(
+				inst,
+				InitiateReserveWithdraw {
+					reserve: MultiLocation { parents: 1, interior: Here },
+					..
+				} | TransferReserveAsset { dest: MultiLocation { parents: 1, interior: Here }, .. }
```
KiChjang, line 45: This is now missing `DepositReserveAsset`.

polkadot-parachains/parachains-common/src/xcm_config.rs
```diff
@@ -0,0 +1,60 @@
...
+			)
+		}) {
+			return Err(()) // Deny
+		}
+
+		// Don't allow reserve transfers to arrive from relay chain
```
joepetrowski, line 51: Why do we want to block these? I would think that in this case we at least want a message back to the Relay Chain to return the assets to their origin.

polkadot-parachains/parachains-common/src/xcm_config.rs
```diff
@@ -0,0 +1,60 @@
+use core::marker::PhantomData;
+use frame_support::weights::Weight;
+use xcm::latest::prelude::*;
+use xcm_executor::traits::ShouldExecute;
+
+//TODO: move DenyThenTry to polkadot's xcm module.
+/// Deny executing the xcm message if it matches any of the Deny filter regardless of anything else.
```
joepetrowski, line 7:
```suggestion
/// Deny executing the XCM if it matches any of the Deny filter regardless of anything else.
```

joepetrowski (commented):
From my understanding of `Barrier`s, they are more about preventing execution of _incoming_ messages, while `Filter`s are more about preventing the sending of _outgoing_ messages. However, there seems to be a limitation on filters in that they only filter for the sender's `Origin`, and here we actually want to prevent outgoing messages with certain contents (namely wherein a parachain acts as a reserve location to the Relay Chain).

Based on that understanding, this PR seems to change the semantic meaning of `Barrier` and use it to _filter_ outgoing messages. If that's what we want, wouldn't it be better to improve filters to support more expressive limitations? Then it would be common syntax that all chains could use to filter outgoing messages.

NachoPal:
> From my understanding of `Barrier`s, they are more about preventing execution of _incoming_ messages, while `Filter`s are more about preventing the sending of _outgoing_ messages. However, there seems to be a limitation on filters in that they only filter for the sender's `Origin`, and here we actually want to prevent outgoing messages with certain contents (namely wherein a parachain acts as a reserve location to the Relay Chain).
> 
> Based on that understanding, this PR seems to change the semantic meaning of `Barrier` and use it to _filter_ outgoing messages. If that's what we want, wouldn't it be better to improve filters to support more expressive limitations? Then it would be common syntax that all chains could use to filter outgoing messages.

`Barrier` is checked every time a XCM message has to be executed, regardless it is _incoming_ or _outgoing_. Therefore it can also act as a "filter" for those actions that imply some XCM execution by the sender (for example for `reserve_transfer_asset` with a `TransferReserveAsset`)

It is also important to notice that while `Filters` are `pallet_xcm` specific types, `Barrier` is part of the `XcmExecutor`. Subsequently, we can see `Barrier` as a "cross-pallet-filter" for all those pallets that make use of the same `XcmExecutor` (not only for `pallet_xcm`).

I agree that currently `Filters` for the `xcm_pallet` are a bit limited. It would be nice to extend them to be able to check also the destination (from  `(origin, message)` to `(origin, destination, message)`. Maybe that is not a good idea because of some design reasons I can not see and the proper way must be only making use of `Barrier`.

polkadot-parachains/parachains-common/src/xcm_config.rs
gilescope: Fixed - Sorry re-read your chat and I understand where I went awry.

gilescope:
Good point that Barrier is used on ingress and egress. Agreed we should push for more context on filters and have logged that here: https://github.com/paritytech/polkadot/issues/5304

parachain-template/runtime/src/xcm_config.rs
```diff
@@ -87,12 +87,19 @@ match_types! {
 	};
 }
 
-pub type Barrier = (
-	TakeWeightCredit,
-	AllowTopLevelPaidExecutionFrom<Everything>,
-	AllowUnpaidExecutionFrom<ParentOrParentsExecutivePlurality>,
-	// ^^^ Parent and its exec plurality get free execution
-);
+#[path = "../../../polkadot-parachains/parachains-common/src/xcm_config.rs"]
```
KiChjang, line 90: What is this syntax here?

bkchr: Why is the template modified at all?

NachoPal: > Why is the template modified at all?

Because of this: https://github.com/paritytech/cumulus/pull/1169#issuecomment-1097908080

NachoPal:
Has it been tested?

parachain-template/runtime/src/xcm_config.rs
bkchr: Why can a parachain not act as a reserve for another parachain?

Imo this makes only sense when you assume that reserves are done on the relay chain.

KiChjang: Hmm, so the problem here is that the relay chain recognizes absolutely no chain to be a reserve, and so any parachain _should not_ be sending `ReserveAssetDeposited` _towards the relay chain_, because that's what we are specifically banning here.

I'm still weirded out by the syntax used here however. Why can't we just import the Rust module properly via Cargo.toml?

bkchr: > Hmm, so the problem here is that the relay chain recognizes absolutely no chain to be a reserve, and so any parachain _should not_ be sending `ReserveAssetDeposited` _towards the relay chain_ as a result of executing `DepositReserveAsset` or `TransferReserveAsset` locally, because that's what we are specifically banning here.

Ahh yeah sorry. It is matched for the sending stuff upwards.

> I'm still weirded out by the syntax used here however. Why can't we just import the Rust module properly via Cargo.toml?

Yeah, this will not be accepted.

gilescope: Dan wasn't so keen on the idea of the template depending on the parachain common code. If you don't like the above, I can brake it out into a separate very-common crate or I can duplicate the code or I could just not include it in the template. Happy to do whatever we can agree on.

gilescope:
RE testing: I've tested that attempting to do a limited reserve asset transfer on statemine/statemint leads to a failure of Barrier. I've not run up a local westmint which does not filter Execute and tried executing individual instructions but that is something we could do.

parachain-template/runtime/src/xcm_config.rs
gilescope: For now I have duplicated the code. Some of that code will be removed once `DenyThenTry` gets pushed down into polkodot's xcm module.

NachoPal (approved):

gilescope:
So just to confirm, I've checked that we can still do teleports back and forth between the relay and statemine, and that we can do a reserve asset transfer on statemine to the community parachain wtihout it getting blocked by the barrier.

parachain-template/runtime/src/xcm_config.rs
```diff
@@ -87,12 +88,78 @@ match_types! {
...
+			return Err(()) // Deny
+		}
+
+		// allow reserve transfers to arrive from relay chain
+		if matches!(origin, MultiLocation { parents: 1, interior: Here }) &&
+			message.0.iter().any(|inst| matches!(inst, ReserveAssetDeposited { .. }))
+		{
+			log::info!(
+				target: "runtime::xcm-barier",
+				"Unexpected Reserve Assets Deposited on the relay chain",
```
KiChjang, line 146:
```suggestion
				target: "xcm::barriers",
				"Unexpected ReserveAssetDeposited from the relay chain",
```

hbulgarini (approved):
Approved.

polkadot-parachains/parachains-common/src/xcm_config.rs
```diff
@@ -0,0 +1,67 @@
...
+		}) {
+			return Err(()) // Deny
+		}
+
+		// allow reserve transfers to arrive from relay chain
+		if matches!(origin, MultiLocation { parents: 1, interior: Here }) &&
+			message.0.iter().any(|inst| matches!(inst, ReserveAssetDeposited { .. }))
+		{
+			log::info!(
+				target: "runtime::xcm-barier",
```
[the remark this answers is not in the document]
gilescope: Good spot. Sorry I was searching on "xcm::barriers".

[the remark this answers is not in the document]
gilescope: Sorry: https://github.com/paritytech/cumulus/pull/1190

parachain-template/runtime/src/xcm_config.rs
```diff
@@ -87,12 +88,78 @@ match_types! {
...
+		message: &mut Xcm<Call>,
+		max_weight: Weight,
+		weight_credit: &mut Weight,
+	) -> Result<(), ()> {
+		Deny::should_execute(origin, message, max_weight, weight_credit)?;
+		Allow::should_execute(origin, message, max_weight, weight_credit)
+	}
+}
+
+// See issue #5233
```
AurevoirXavier, line 115: Should be paritytech/polkadot#5233

joepetrowski: Thanks, meant to make a follow up PR to docs in parachain-template. Will fix this too.

PR #1170, opened 2022-04-13
Unify Cargo.toml format

koushiro:
- Add `taplo` config for unifying `Cargo.toml` format
- Add CI job for checking `Cargo.toml` format

### DIFF
--- a/.github/workflows/quick-check.yml
+++ b/.github/workflows/quick-check.yml
@@ -38,3 +38,9 @@ jobs:
         with:
           command: fmt
           args: --all -- --check
+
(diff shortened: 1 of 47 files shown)


gilescope:
FYI I think we are seeing some movement on cargo fmt doing Cargo.tomls finally: https://github.com/rust-lang/rustfmt/pull/5240

bkchr:
Ty @koushiro, but for now we will not use any external tool to format the toml files. Maybe when it is in rustfmt and stable, we can again speak about this.

PR #1190, opened 2022-04-20
warn not info

gilescope:
missed one of the locations when doing https://github.com/paritytech/cumulus/pull/1169

### DIFF
--- a/polkadot-parachains/parachains-common/src/xcm_config.rs
+++ b/polkadot-parachains/parachains-common/src/xcm_config.rs
@@ -57,8 +57,8 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
 			message.0.iter().any(|inst| matches!(inst, ReserveAssetDeposited { .. }))
 		{
 			log::warn!(
-				target: "runtime::xcm-barier",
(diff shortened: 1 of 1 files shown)


polkadot-parachains/parachains-common/src/xcm_config.rs
```diff
@@ -56,7 +56,7 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
 		if matches!(origin, MultiLocation { parents: 1, interior: Here }) &&
 			message.0.iter().any(|inst| matches!(inst, ReserveAssetDeposited { .. }))
 		{
-			log::info!(
+			log::warn!(
 				target: "runtime::xcm-barier",
 				"Unexpected Reserve Assets Deposited on the relay chain",
```
KiChjang, line 61:
```suggestion
				target: "xcm::barrier",
				"Unexpected ReserveAssetDeposited from the relay chain",
```

gilescope: That makes more sense that they should be the same.

KiChjang (approved):

PR #1197, opened 2022-04-21
Make sure when storage is empty we set the block number

girazoki:
I made a PR yesterday that only updated the relay number if the ValidationData was non-empty, but we run into cases where we need also to update it if its empty. This PR fixes this.

The previous PR https://github.com/paritytech/cumulus/pull/1187


bkchr (approved):
I had thought about this as well, but then had discarded the thought :P

PR #1207, opened 2022-04-22
Default not implemented for non-std builds for PersistedValidationData

girazoki:
cc @bkchr sorry to bother you again about the same thing. When going to benchmark moonbase I realized that PersistedValidationData does not implement default for non-std builds, so we cannot really use unwrap_or_default. Instead I need to initialize manually each of the fields :(.

Here is the fix. I tested the benchmark and it works with this fix. Again, sorry for not realizing about this before!

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -1049,7 +1049,7 @@ impl<T: Config> BlockNumberProvider for RelaychainBlockNumberProvider<T> {
 	}
 	#[cfg(feature = "runtime-benchmarks")]
 	fn set_block_number(block: Self::BlockNumber) {
-		let mut validation_data = Pallet::<T>::validation_data().unwrap_or(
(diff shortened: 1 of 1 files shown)


pallets/parachain-system/src/lib.rs
```diff
@@ -1049,7 +1049,15 @@ impl<T: Config> BlockNumberProvider for RelaychainBlockNumberProvider<T> {
 	}
 	#[cfg(feature = "runtime-benchmarks")]
 	fn set_block_number(block: Self::BlockNumber) {
-		let mut validation_data = Pallet::<T>::validation_data().unwrap_or_default();
+		let mut validation_data = Pallet::<T>::validation_data().unwrap_or(
```
bkchr, line 1052:
```suggestion
		let mut validation_data = Pallet::<T>::validation_data().unwrap_or_else(||
```

bkchr (approved):
Haha :D

bkchr:
bot merge

PR #1209, opened 2022-04-24
Rpc cors for app and explorer

brenzi:
our collator blocked the mobile app connections with :
```
2022-04-24 11:45:03.245  WARN ThreadId(74) jsonrpc_ws_server::session: Blocked connection to WebSockets server from untrusted origin: Some("null") 
```

this PR solves that


brenzi:
wrong repo..again

Issue #1222, opened 2022-04-28
Is `rococo-parachain` necessary?


NachoPal:
Coming from: https://github.com/paritytech/cumulus/pull/1223

I would like to understand what is the role of `rococo-parachain`. I think `parachain-template` should be sufficient and the way to go if you want to tests/tinker/build a parachain.

It seems to me that having both runtimes coexisting is redundant. `rococo-parachain` configuration is closer to what we (the Common Goods parachains team) want to achieve (a parachain template ready to interact with a Relay chain and its Common Good parachains), however I think that logic should go to the template. Furthermore, right now it is unclear to me what configuration `parachain-template` is following.

I would like to propose removing `rococo-parachain` unless there is a reason for keeping both runtimes. It will be also one runtime less to maintain.

NukeManDan:
https://github.com/paritytech/cumulus/pull/1223#issuecomment-1112495530
> This template is used to start most parachain projects from, and thus is intended to be minimal AKAIU, thus adding more functionality here, especially divergent from the [substrate template](https://github.com/paritytech/substrate/tree/master/bin/node-template), isn't ideal.
> 
> If common good chains need a more purpose built template, I would think it should live independent from the soverign template (we have now) if significantly different. Especially in the XCM configuration, because this template should never be set such that a soverign chain basing off the settings here unchanged could allow for the same trust with the relay and other siblings. Issues happened in this light, addressed in https://github.com/paritytech/cumulus/pull/696 (at least this was attempted, not merged now I see grimacing )

bkchr:
> a parachain template ready to interact with a Relay chain and its Common Good parachains

I don't really get why we need a template for everything. These should be tutorials and not just random templates. In the end there would be thousands of templates, while they actually just have 2-3 xcm things configured differently.

hbulgarini:
@PierreBesson is `rococo-parachain` used in the testnets?

bkchr:
Tick/Trick/Track are currently based on `rococo-parachain`. 

>  of the CLI with something like --chain staging-parachain to build "dumb parachains" chainspecs and replace the tick, trick and track raw chainspecs.

Yeah that could be done.

hbulgarini:
> Tick/Trick/Track are currently based on rococo-parachain.
@bkchr are we still using Tick/Trick/Track? If we do, then we will keep parachain runtime. If not, i would say that we might remove it.

PierreBesson:
[...]

PR #1223, opened 2022-04-28
Update `parachain-template`

NachoPal:
This PR configures `parachain-template` to be ready to properly interact with the Common Goods parachains (only Statemint for now but more are coming)

It is unclear to me what configuration `parachain-template` was following.

This makes me think about the role of `rococo-parachain` runtime, whose configuration seems much closer to what we want to achieve. To discuss about it I created this issue with my view: https://github.com/paritytech/cumulus/issues/1222

I'll switch the PR to Draft while @gilescope finalise the XCM config


NukeManDan:
This template is used to start most parachain projects from, and thus is intended to be _minimal_ AKAIU, thus adding more functionality here, especially divergent from the [substrate template](https://github.com/paritytech/substrate/tree/master/bin/node-template), isn't ideal.

If common good chains need a more purpose built template, I would think it should live independent from the soverign template (we have now) if significantly different. _Especially_ in the XCM configuration, because this template should never be set such that a soverign chain basing off the settings here unchanged could allow for the same trust with the relay and other siblings. Issues happened in  this light, addressed in https://github.com/paritytech/cumulus/pull/696 (at least this was attempted, not merged now I see :grimacing: )

bkchr:
I'm here with @NukeManDan. These templates are about the bare minimum you need to run a parachain and that clearly doesn't involve communicating with any common good parachain. Which very probably also doesn't exist on local networks when you begin to code.

The template should not be misused to act as some kind of tutorial. Here you actually want to have some proper tutorial that could live on substrate docs. This would also be much better for explaining on what is what doing.

NachoPal:
Ok, I understand, it can make sense to have a bare minimum parachain template. `parachain-template` will stay as it is.

Then I'll propose to make these changes to `rococo-parachain`. I see that actually it has already some XCM config to support Statemine.

Is it ok?

bkchr:
> Ok, I understand, it can make sense to have a bare minimum parachain template. `parachain-template` will not be modified.
> 
> Then I'll propose to make these changes in `rococo-parachain`. I see that actually it has already some XCM config to support Statemine.
> 
> Is it ok?

If it is about testing. Makes sense.

parachain-template/node/src/command.rs
```diff
@@ -23,14 +23,44 @@ use sp_runtime::traits::Block as BlockT;
 use std::{io::Write, net::SocketAddr};
 
 fn load_spec(id: &str) -> std::result::Result<Box<dyn sc_service::ChainSpec>, String> {
+	let (norm_id, para_id) = extract_parachain_id(id);
+
 	Ok(match id {
```
gilescope, line 28: `match id`  should be `match norm_id`

PR #1225, opened 2022-04-29
WIP: rococo renamed to pal. no longer publishing.

gilescope:
This is the rename of the rococo runtime on the grounds that no one else is using it.

We will then reconfigure it to be able to do better reserve transfer testing.

Rather than having our own runtime API `pal_parachain_runtime::RuntimeApi`, could we reuse a common one like shell or seedling's RuntimeApi ?


polkadot-parachains/pal-parachain/Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [package]
-name = "rococo-parachain-runtime"
+name = "pal-parachain-runtime"
```
bkchr, line 2: What does pal stands for? 

I don't get the renaming?

polkadot-parachains/Cargo.toml
```diff
@@ -107,4 +107,4 @@ try-runtime = [
 	"statemine-runtime/try-runtime",
 	"westmint-runtime/try-runtime",
 	"shell-runtime/try-runtime",
-]
+]
```
chevdor, line 110: eof :)

chevdor (approved):
Thank you for including the whole release notes fixes, LGTM

polkadot-parachains/pal-parachain/Cargo.toml
gilescope: (sorry I thought I replied to this via email but must not have sent)
Pal is another word for friend. In order to test parachains xcm functionality you need other parachains to talk to. In particular we'd like this chain to mint and burn derivative-tokens to represent the reserve asset being transferred as this would simulate how we expect others to interact with statemine/t and also be an example to others of how such can be supported.

(we could keep rococo-parachain and have Pal as a separate runtime if people still have need of it.)

bkchr: I mean we can change it, but can we not come up with a different name? AFAIK it is also still being used for Versi testing. 

> example to others of how such can be supported.

You are speaking again about tutorials. We should create a tutorial for this and not really have some code lying around that no one understands.

gilescope:
I'm going to close this as this should be replaced with two separate PRs. one to remove the rococo runtime and another to add penpal runtime.

gilescope:
Pal means friend. The goal of the chain is to be a good friend to other
chains and help testing things like reserve transfers on statemine/t that
we can’t quite fully test without a friendly parachain to talk to. You can
see in @nachopal’s cumulus PR the contents that we intend to transform the
chain to (though in that pr it tried to take over parachain-template) . Our
team doesn’t see any value in the rococo runtime as it is. If there is some
reason why it should stay then we can add the pal chain as well as the
rococo runtime. At the moment we have a whole fork of cumulus where the pal
chain is being developed ( nachopal/community-parachain ). It didn’t seem
appropriate for it to be called community and so the team settled on pal as
a name. This is about consolidating what we already have rather than
creating anything new.

PR #1228, opened 2022-05-02
Add event for showing the hash of an UMP sent message

girazoki:
Adds Event showing Hash of message sent to parent through UMP

The goal is to match this ID with the one that the UMP queue shows in the event: https://github.com/paritytech/polkadot/blob/0fca26fb0a5bdf00bde70a9bcc31805dad61b2a9/runtime/parachains/src/ump.rs#L139


pallets/parachain-system/src/lib.rs
```diff
@@ -422,6 +426,8 @@ pub mod pallet {
 		/// Downward messages were processed using the given weight.
 		/// \[ weight_used, result_mqc_head \]
 		DownwardMessagesProcessed(Weight, relay_chain::Hash),
+		/// An Upward message was sent to the parent.
+		UpwardMessageSent(Option<MessageId>),
```
bkchr, line 430: Why don't you use `T::Hash` here as well?

girazoki: I just tried to mimic as much as possible what ump shows so that both ids match. But I am happy to change it if I found no issues doing so

KiChjang: This really should be done as part of XCMv3, since there is a concept of XCM hashes there, and we can use it here as well.

girazoki: I can PR the xvm V3 branch if prefered

gilescope: (Keith pointed me out this SE answer when I was sniffing around for something similar: https://substrate.stackexchange.com/questions/1893/is-there-an-id-for-xcm-message/1906#1906 )

girazoki: Yes I agree with what Gav said. And although the message hash being different is something each chain can work on indepentently, I think to help explorers identify these hashes the ump queue should emit an event with this hash. I will re-work this PR to go into xcm V3, as it seems we already have the notion of "hash of the message" there

girazoki:
Back to draft because I need to test it with cargo override

gilescope:
any news?

girazoki:
I think I can make it reafdy for merge, as I tested it against the xcmv3 branch back on its day

girazoki:
For some reason, I obtained different results using` message.using_encoded(sp_io::hashing::blake2_256)` vs using `sp_io::hashing::blake2_256(message)`. I can investigate further, but I chose using the later as it is what the ump queue outputs too https://github.com/paritytech/polkadot/blob/ae53f58f9975a03c6d6b282e51b9d46f5a92f991/runtime/parachains/src/ump.rs#L92

girazoki:
Alright using_encoded yields a different slice (because it prepends a first byte indicating the lenght). Since we want to mimic the ump event I think we should stay with `sp_io::hashing::blake2_256(message)`

albertov19:
This will not get merged into XCM the current XCM implementation? Having a hash to match a XCM message sent from a parachain to a relay chain (or the other way around) is key to improve the UX of XCM in general. 

We have been offering a XCM explorer bounty for two hackathons in a row, but without this feature, it is not possible to have a straightforward way to link XCM messages between chains. Note that from parachain to parachain this is not true, and it is possible to link XCM message sent on one chain, to XCM message executed in the other chain.

girazoki:
Hi! Just wanted to know whether there is any plan of incorporating this or it conflicts with https://github.com/paritytech/polkadot/issues/6121

CC: @KiChjang @joepetrowski

joepetrowski:
I don't see how it conflicts, unless 6161 implies the removal of `send_upward_message` from Parachain System? Nonetheless, the event will still be useful for indexing services.

KiChjang:
I don't think this conflicts, but will need to double-check whether XCM v3 already includes an XCM hash in the event. My guess is no, in which case this PR does not conflict with anything at all.

gilescope:
Personally I would love to see this land.

KiChjang (approved):

pallets/parachain-system/src/lib.rs
```diff
@@ -1059,13 +1062,20 @@ impl<T: Config> Pallet<T> {
 				// Thus fall through here.
 			},
 		};
-		<PendingUpwardMessages<T>>::append(message);
-		Ok(0)
+		<PendingUpwardMessages<T>>::append(message.clone());
+
+		// The relay ump does not use using_encoded
+		// We apply the same this to use the same hash
+		// TODO: shall we change both ump and this to
```
bkontur, line 1069: what about this TODO?

girazoki: It seems https://github.com/paritytech/polkadot/pull/6271 will still use `sp_io::hashing::blake2_256(message)`. I guess we can leave it as is (and remove the todo), as it is the only way hashes will match

bkchr:
@KiChjang do we want this merged?

KiChjang:
Yes, but not at this current state as it's failing CI.

bkchr:
@girazoki can you make the CI working? :)

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2215208

girazoki:
I fixed the fmt checks, but the build jobs wont pass because this is going to the `xcmv3` branch and the CI does not pass there. If you prefer, I can make it go to master, but then the `xcmv3` branch will need to be updated

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2215233

bkchr:
```
error[E0412]: cannot find type `XcmHash` in this scope
   --> primitives/core/src/lib.rs:100:61
    |
96  | pub trait UpwardMessageSender {
    |                              - help: you might be missing a type parameter: `<XcmHash>`
...
100 |     fn send_upward_message(msg: UpwardMessage) -> Result<(u32, XcmHash), MessageSendError>;
    |                                                                ^^^^^^^ not found in this scope
error[E0412]: cannot find type `XcmHash` in this scope
   --> primitives/core/src/lib.rs:103:62
    |
102 | impl UpwardMessageSender for () {
    |     - help: you might be missing a type parameter: `<XcmHash>`
103 |     fn send_upward_message(_msg: UpwardMessage) -> Result<(u32, XcmHash), MessageSendError> {
    |                                                                 ^^^^^^^ not found in this scope
    ```

But these messages look valid?

**Comment by girazoki** [comment]
I think this is because the branch still is pointing at the master branch of `polkadot`, which does not contain the `xcmHash` type. I see similar errors on the `xcmv3` branch:

```
error[E0412]: cannot find type `XcmHash` in this scope
   --> pallets/dmp-queue/src/lib.rs:479:11
    |
479 |             _hash: XcmHash,
    |                    ^^^^^^^ not found in this scope
```

bkchr:
Ahh okay!

@KiChjang please verify that this branch compiles for you and then let's merge it.

KiChjang:
Ah okay, let me just merge this as-is.

Issue #979, opened 2022-02-10
Rename `polkadot-collator` bin


joepetrowski:
Related to https://github.com/paritytech/cumulus/issues/928

A lot of people I've talked to recently are confused about what a collator is and often think they need to have a collator in order to run a node. Of course, it is the other way around. We should rename the binary such that being a `collator` is an option. I think `parachain-node` would be OK but happy with anything more general than "collator".

NukeManDan:
I agree :point_up: I would suggest also moving the template to `parachain-node-template` so that cumulus is roughly homomorphic to Substrate with a `parachain-` prefix on the main binaries.

joepetrowski:
I don't have an opinion on that but it's beyond the scope of this issue (naming of "Assets" in the releases vs. code structuring). I'll close this once CI starts calling the binary `parachain-node`, but feel free to log another issue about code/development templates.

bkchr:
I would go for `polkadot-parachains`. What does `parachain-node` convey to you? Which parachain? All? 

In the far future I would actually like to see some kind of "Ueber" node that supports all parachains, at least for syncing them :)

joepetrowski:
> What does parachain-node convey to you?

That it's a node for a parachain :)

> Which parachain? All?

I thought that with Wasm execution, yes.

bkchr:
But even with Wasm execution, the node is not prepared to run every parachain. Statemine/Statemint for example need to go through this change from shell to the actual runtime and that requires some special code on the node side. Statemint is then using different crypto etc. There is a ton of stuff why this node will not be used for *all* parachains that easily.

xlc:
I never know it suppose to work with other parachains? Or does it?

bkchr:
No. It was never written in any way to work with other parachains. As I said, we could probably make that possible in the future, but that will not happen in the near future.

joepetrowski:
> I would go for polkadot-parachains

OK, I'm good with this. Would drop the plural and just call it `polkadot-parachain` but won't put up too much of a fight :)

bkchr:
I also don't care about dropping the plural. I'm also no native speaker, so I'm the last one that will say something about correct naming :D

PR #1241, opened 2022-05-06
rename polkadot-collator to polkadot-parachain

gilescope:
Not every node has to be a collator.

Fixes https://github.com/paritytech/cumulus/issues/979


alvicsam (approved):

joepetrowski (approved):
Approving the name change, but someone more familiar with CI/versioning should review that everything works as expected.

bkchr (approved):

sergejparity (approved):

docker/polkadot-collator_builder.Containerfile
```diff
@@ -1,36 +1,36 @@
 # This file is sourced from https://github.com/paritytech/polkadot/blob/master/scripts/ci/dockerfiles/polkadot/polkadot_builder.Dockerfile
```
skunert, line 1: I think this file needs to be renamed too. It still says `polkadot-collator...` but is referenced in multiple places as `polakdot-parachain...`.

chevdor: you are right, good catch

.github/workflows/extrinsic-ordering-check-from-bin.yml
```diff
@@ -10,7 +10,7 @@ on:
         required: true
       binary_url:
         description: A url to a Linux binary for the node containing the runtime to test
-        default: https://github.com/paritytech/cumulus/releases/download/statemine_v3/polkadot-collator
+        default: https://github.com/paritytech/cumulus/releases/download/statemine_v3/polkadot-parachain
```
chevdor, line 13: https://releases.parity.io/cumulus/polkadot-v0.9.21/polkadot-parachain
```suggestion
        default: https://releases.parity.io/cumulus/polkadot-v0.9.21/polkadot-parachain
```

README.md
```diff
@@ -33,7 +33,7 @@ and treat as best.
 ### Collator
 
 A Polkadot [collator](https://wiki.polkadot.network/docs/en/learn-collator) for the parachain is
-implemented by the `polkadot-collator` binary.
+implemented by the `polkadot-parachain` binary.
```
chevdor, line 36:
```suggestion
implemented by the `polkadot-parachain` binary (previously called `polkadot-collator`).
```

polkadot-parachains/Cargo.toml
```diff
@@ -1,12 +1,13 @@
 [package]
-name = "polkadot-collator"
-version = "5.4.0"
+name = "polkadot-parachain"
+version = "0.9.19"
```
chevdor, line 3: No, why would that be?
In the future, will we have a new version of the `polkadot-parachain` for each new version of polkadot ? and never in between ?
The answer maybe "kinda yes today" but that is not the plan and this is why the collator version is decoupled from polkadot.

```suggestion
version = "5.5.0"
```

chevdor (changes requested):
A few points to fix IMO but thanks for the big renaming!

.github/workflows/extrinsic-ordering-check-from-bin.yml
gilescope: Can we pull the version from a github variable somehow? Worst case we could 'extract' it from the Cargo.lock file... don't really want to change it every time.

polkadot-parachains/Cargo.toml
gilescope: Less of an issue once polkadot reaches `1.0`. For now we can do `0.9.19-patch1` if we really need to. I get the 'in theory we might need to do a patch', but we have a lot of users and this makes things simpler if we in general we keep the version numbers in check. If my polkadot parachain collator version is matching the relay chain version I get a warm fuzzy feeling inside that I just don't get with `v5.5`.

joepetrowski: Because (a) the Cumulus node contains a Polkadot node, so having a major Cumulus release doesn't make sense, and (b) it was a bad change [in the first place](https://github.com/paritytech/cumulus/pull/649/files#diff-645962a0e1b239c8d20e814dcdb19ea39e7d9ab3adb51d59b4abfd7571f1be65R3), bumping from 0.1.0 to 4.0.0.

We shouldn't have a 1.0.0 client release at all.

docker/polkadot-collator_builder.Containerfile
gilescope: fixed

.github/workflows/extrinsic-ordering-check-from-bin.yml
chevdor: There is no need, this is just the default to help the user. It is nice to keep it close to the "good" value but no need to make that part over complicated, and definitely no need to bump that with every version. In the end, this is just a URL to the binary. I just took the chance to edit this here as the github url **cannot** work and since we rename the binary, having a pattern close to the reality for the S3 url is helpful when we call this GHA.

polkadot-parachains/Cargo.toml
joepetrowski: > We cannot go back in terms of release number.

Why not? Very few people are running nodes now. If there's a time to do it, it's now.

gilescope: It's got a different binary name so it's allowed a different version number. Windows is now on 11. I remember when it was 95...

gilescope: Remember it's only statemine/t that use this directly. I've checked a lot of other parachain repos and they have no mention of `polkadot-collator`.

joepetrowski: Yeah but others will (like Alliance, governance, etc)

skunert (approved):

chevdor (approved):
I am not convinced about this `0.9.190` scheme but I prefer it to the `-1` options,  everyone seems to agree about the the need to follow the prolkadot version and I don't want to slow this down.

hbulgarini (changes requested):
When the node is started without `--collator` option the node still prints the `Polkadot Collator` log:

```
./polkadot-collator --tmp --port 40337 --ws-port 9948 -- --execution wasm --chain polkadot-local-raw.json  --port 30337
2022-05-12 13:45:26 Polkadot collator
2022-05-12 13:45:26 ✌️  version 5.4.0-4bbedb30b
...
2022-05-12 13:45:26 Is collating: no
```

hbulgarini (approved):
LGTM

PR #1255, opened 2022-05-11
Companion for paritytech/polkadot#5388

slumber:
paritytech/polkadot#5388

### DIFF
--- a/client/consensus/common/src/lib.rs
+++ b/client/consensus/common/src/lib.rs
@@ -22,6 +22,7 @@ mod parachain_consensus;
 #[cfg(test)]
 mod tests;
 pub use parachain_consensus::run_parachain_consensus;
+pub use polkadot_primitives::v2::CollationForecast;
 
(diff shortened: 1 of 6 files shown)


client/consensus/common/src/lib.rs
```diff
@@ -67,6 +75,15 @@ impl<B: BlockT> ParachainConsensus<B> for Box<dyn ParachainConsensus<B> + Send +
...
 	}
+
+	async fn is_collating(
+		&mut self,
+		parent: &B::Header,
+		relay_parent: PHash,
+		validation_data: &PersistedValidationData,
+	) -> Option<CollationForecast> {
+		(*self).is_collating(parent, relay_parent, validation_data).await
+	}
```
slumber, line 86: @bkchr should we preconnect for relay-chain consensus? Seems like we always want to do this if the core is free, right?

bkchr: You added the comment to the wrong position, but yeah for the "relay-chain consensus" you always want to connect.

PR #1257, opened 2022-05-13
Rename of polkadot-collator to polkadot-parachain

gilescope:
Rename collator bin and make this same code available under the new docker container name.
( backport of https://github.com/paritytech/cumulus/pull/1241 )


bkchr:
Why is this backported? There will be a new release next week.

chevdor:
Indeed, there is no reason to modify 0.9.19, those releases are out already.

gilescope:
We have nothing currently in docker at `polkadot-parachain:latest` - this could cause trouble for zombinet next week if they start using the new bin name.

gilescope:
@joepetrowski suggested this as a simple solution and I tend to agree with him.

gilescope:
ok let's close this as next version is almost upon us.

PR #1260, opened 2022-05-14
Use named events in cumulus pallets

gilescope:
This migrates cumulus pallets to use named events.

Partially fixes: https://github.com/paritytech/polkadot/issues/5422

Reviewers please pay special attention to naming consistency with other naming event PRs:

https://github.com/paritytech/polkadot/pull/5423
https://github.com/paritytech/polkadot/pull/5461

### DIFF
--- a/pallets/dmp-queue/src/lib.rs
+++ b/pallets/dmp-queue/src/lib.rs
@@ -152,7 +152,7 @@ pub mod pallet {
 			let used = Self::try_service_message(weight_limit, sent_at, &data[..])
 				.map_err(|_| Error::<T>::OverLimit)?;
 			Overweight::<T>::remove(index);
-			Self::deposit_event(Event::OverweightServiced(index, used));
+			Self::deposit_event(Event::OverweightServiced { overweight_index: index, used });
 			Ok(Some(used.saturating_add(1_000_000)).into())
 		}
 	}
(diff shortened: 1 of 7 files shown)


gilescope:
(I pulled in the latest polkadot master as well, possibly that was not required)

gilescope:
Should we be calling things message_id or message_hash? Personally I think as it's not guaranteed to be unique we should be calling it what it is: a message hash. But there's an argument that it could change to something more unique in the future and then maybe message id is justified?

pallets/parachain-system/src/lib.rs
```diff
@@ -763,7 +765,10 @@ impl<T: Config> Pallet<T> {
 			weight_used += T::DmpMessageHandler::handle_dmp_messages(message_iter, max_weight);
 			<LastDmqMqcHead<T>>::put(&dmq_head);
 
-			Self::deposit_event(Event::DownwardMessagesProcessed(weight_used, dmq_head.head()));
+			Self::deposit_event(Event::DownwardMessagesProcessed {
+				used: weight_used,
+				dmq_head: dmq_head.head(),
```
gilescope, line 770: Is there a better event name than dmq_head? last_chain_head_processed? last_blockhash_processed?

pallets/dmp-queue/src/lib.rs
```diff
@@ -162,22 +162,26 @@ pub mod pallet {
 	pub enum Event<T: Config> {
 		/// Downward message is invalid XCM.
 		/// \[ id \]
```
bkchr, line 164: Please remove these comments. They are useless now.
```suggestion
```

pallets/parachain-system/src/lib.rs
bkchr: No, that is the correct name.

pallets/xcm/src/lib.rs
```diff
@@ -67,13 +67,13 @@ pub mod pallet {
 	pub enum Event<T: Config> {
 		/// Downward message is invalid XCM.
 		/// \[ id \]
-		InvalidFormat([u8; 8]),
+		InvalidFormat { message_id: [u8; 8] },
```
bkchr, line 70: No xcm changes outside the xcm-v3 branch.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -263,21 +263,26 @@ pub mod pallet {
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// Some XCM was executed ok.
-		Success(Option<T::Hash>),
+		Success { message_id: Option<T::Hash> },
```
bkchr, line 266: Xcm related as well.

bkchr (approved):

gilescope:
I'm inclided to rename `used` and `required` to `used_weight` and `required_weight`. Otherwise I read `used` and have to wonder used what until I find the type.

joepetrowski (approved):

Issue #1247, opened 2022-05-09
Statemine/t: Support self sufficient asset for XCM fee payment


xlc:
Currently only KSM is accepted for XCM fee payment.
We should make it accept any self sufficient asset for XCM fee payment as well.

https://github.com/paritytech/cumulus/blob/0c779d926beeb6e1f3f2aa777d9bcec792c0bdac/polkadot-parachains/statemine/src/xcm_config.rs#L171

Relevant tests: https://github.com/AcalaNetwork/Acala/pull/2097

girazoki:
@notlesh and I have worked on a potential implementation of WeightTrader to support this. Its a very early draft to check how this could be worked. Here is the main idea of it, we can keep working on it if the main idea sounds good:

```
pub struct TakeFirstAssetTrader<
	WeightToFee: WeightToFeePolynomial<Balance = Currency::Balance>,
	AssetId: Get<MultiLocation>,
	AccountId,
	Currency: CurrencyT<AccountId>,
	CON: BalanceConversion<Currency::Balance, Assets::AssetId, Assets::Balance>,
	Matcher: MatchesFungibles<Assets::AssetId, Assets::Balance>,
	Assets: fungibles::Mutate<AccountId> + fungibles::Transfer<AccountId>,
	OnUnbalanced: OnUnbalancedT<Currency::NegativeImbalance>,
>(
	Weight,
	Assets::Balance,
	Option<(MultiLocation, Assets::AssetId)>,
	PhantomData<(WeightToFee, AssetId, AccountId, Currency, CON, Matcher, Assets, OnUnbalanced)>,
);
impl<
		WeightToFee: WeightToFeePolynomial<Balance = Currency::Balance>,
		AssetId: Get<MultiLocation>,
		AccountId,
		Currency: CurrencyT<AccountId>,
		CON: BalanceConversion<Currency::Balance, Assets::AssetId, Assets::Balance>,
		Matcher: MatchesFungibles<Assets::AssetId, Assets::Balance>,
		Assets: fungibles::Mutate<AccountId> + fungibles::Transfer<AccountId>,
		OnUnbalanced: OnUnbalancedT<Currency::NegativeImbalance>,
	> WeightTrader for UsingComponentsWithFirstAsset<WeightToFee, AssetId, AccountId, Currency, CON, Matcher, Assets, OnUnbalanced>
{

	fn new() -> Self {
		Self(0, Zero::zero(), None, PhantomData)
	}

	// We take first asset
[...]

PR #1266, opened 2022-05-17
dir restructure to support more CGP

gilescope:
Readying the ground for more CGP. Fixes #1240

PR #1278, opened 2022-05-20
Statemint runtimes to accept sufficient assets as xcm fee payment

girazoki:
Solves #1247 

It adds a new trader `TakeFirstAssetTrader`, which takes the first asset and tries to charge xcm execution on such asset. If such asset in a sufficient asset in `pallet-assets`, then the `fee` (corresponding to how much it would cost in the native token) is converted to an `asset_fee`(corresponding to how much it would cost in the specified asset) based on the difference on the existential deposits.

Charged fee is for now given to the block author. I am open to suggestions on this

Missing tests, I am unsure where to add these tests. Open to suggestions on this too


joepetrowski:
> Missing tests, I am unsure where to add these tests. Open to suggestions on this too.

We are working on a [dir restructure](https://github.com/paritytech/cumulus/pull/1266) that will include a folder for all asset parachain runtimes. I think we could probably add a `tests.rs` file to that to test behavior that should be common for all Statemint variants.

hbulgarini:
Thanks for the PR @girazoki ! could you please fix the conflicting files? as Joe mentioned the [dir restructure](https://github.com/paritytech/cumulus/pull/1266) might originates the conflicts described.

girazoki:
sure!

girazoki:
I will now proceed to add the appropriate` tests.rs` file testing this. Thanks!

girazoki:
Alright I think this is ready for review. In summary:

I added a tests.rs file per statemine runtime. The reason for this is that the configuration of the new trader depends heavily (e.g., the existential deposit). Let me know if this is sufficient. I also made the trait a bit more generic, so that other teams can use the TakeFirstAssetTrader trait too.w
Importantly we had to require the `TakeRevenue` trait for `TakeFirstAssetTrader`. The reason for this is that we need to call `mint_into` from pallet-assets. In XCM, withdraw_assets from `FungiblesAdapater` burns the tokens. Therefore, we cannot use the `AssetsToBlockAuthor` because it works with CreditOf<AccountIdOf<R>, which does not mint the previously burned tokens to the author. Instead it would only increase its balance, but we need to mint them

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,157 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+		AccountId: Clone,
+		ReceiverAccount: frame_support::traits::Get<Option<AccountId>>,
+	> TakeRevenue for XcmFeesToAccount<Assets, Matcher, AccountId, ReceiverAccount>
+{
+	fn take_revenue(revenue: MultiAsset) {
+		match Matcher::matches_fungibles(&revenue) {
+			Ok((asset_id, amount)) =>
+				if let Some(receiver) = ReceiverAccount::get() {
+					if !amount.is_zero() {
+						let ok = Assets::mint_into(asset_id, &receiver, amount).is_ok();
```
joepetrowski, line 197: I don't understand why this is minting. Besides the native asset, everything uses reserve backed transfers, so the asset must already be there in the sovereign account of the fee-paying `MultiLocation`. Shouldn't this debit the fee payer and credit the author?

girazoki: I used `HandleCredit` before, similar to what it used in `pallet_asset_tx_payment`. We would achieve the same behavior as we have now, but without an event showing that the author has been credited some tokens. I think the event is important to show how fees are divided between block author and destination address, but I am happy to revert back these changes

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,157 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+		payment: xcm_executor::Assets,
+	) -> Result<xcm_executor::Assets, XcmError> {
+		log::trace!(target: "xcm::weight", "TakeFirstAssetTrader::buy_weight weight: {:?}, payment: {:?}", weight, payment);
+		// We take the very first multiasset from payment
+		// TODO: revisit this clone
+		let multiassets: MultiAssets = payment.clone().into();
+		let first = multiassets.get(0).ok_or(XcmError::TooExpensive)?;
+		// We get the local asset id in which we can pay for fees
+		let (local_asset_id, _) =
+			Matcher::matches_fungibles(&first).map_err(|_| XcmError::TooExpensive)?;
```
NachoPal, line 106: I think it is more informative and accurate `AssetNotFound`
```suggestion
			Matcher::matches_fungibles(&first).map_err(|_| XcmError::AssetNotFound)?;
```

NachoPal (commented):
Is it possible to move `XcmAssetFeesHandler` to `/parachains/common/`?

parachains/common/src/xcm_config.rs
```diff
@@ -65,3 +70,32 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
...
+	>,
+	AccountIdOf<R>:
+		From<polkadot_primitives::v2::AccountId> + Into<polkadot_primitives::v2::AccountId>,
+{
+	fn charge_weight_in_fungibles(
+		asset_id: <pallet_assets::Pallet<R> as Inspect<AccountIdOf<R>>>::AssetId,
+		weight: Weight,
+	) -> Result<<pallet_assets::Pallet<R> as Inspect<AccountIdOf<R>>>::Balance, XcmError> {
+		let amount = WeightToFee::calc(&weight);
+		CON::to_asset_balance(amount, asset_id).map_err(|_| XcmError::TooExpensive)
```
NachoPal, line 99: If we are going to use `BalanceToAssetBalance::to_asset_balance` I think we would have to review the `minimum_balance` of the sufficient Assets we currently have in Statemine/t.

It can lead to very unbalanced fees if the `min_balance` ratios are not well adjusted to the assets price, encouraging people to use a specific asset over others to pay the fees.

**E.g**:
- KSM 
  - existential deposit = 3,333,333
  - decimals = 12  
- USDT
  - minimum balance = 1,000
  - decimals = 6
  
A fee of 1,000,000,000,000 (1 KSM) is equivalent to  1,000,000,000,000 * 1,000/3,333,333 = 300,000,000 USDT (300 USDT)

However, 1 KSM currently is 73$. It means that paying fees with KSM would be x5 times cheaper .

Please let know if the numbers are correct.

girazoki: This is already a problem with the existing pallet_asset_tx_payment 
 https://github.com/paritytech/cumulus/blob/8befd605b6a29210cb47c74ed484e90f90ebaaac/parachains/runtimes/assets/statemine/src/lib.rs#L501 no?

In that sense, this PR does not introduce any change. However, my personal opinion is that we should review definitely the minimum amount of sufficient assets. The numbers seem definitely correct to me.

girazoki:
I think I addressed the comments, could you take another pass? The only concern are the potential difference in fees  between the native token and an asset, but I think that should not be part of this PR as it is already an issue with `pallet_asset_tx_payment`.

joepetrowski:
> I think I addressed the comments, could you take another pass? The only concern are the potential difference in fees between the native token and an asset, but I think that should not be part of this PR as it is already an issue with `pallet_asset_tx_payment`.

Cool, we are all at a team retreat, and my first priority right now is fixing the failed Statemint HRMP channel opening, but will definitely get to this soon.

NachoPal:
@girazoki one of the tests is failing: `test_asset_xcm_trader_refund_not_possible_since_amount_less_than_ed`

```
test test_asset_xcm_trader_refund_not_possible_since_amount_less_than_ed ... FAILED
failures:
---- test_asset_xcm_trader_refund_not_possible_since_amount_less_than_ed stdout ----
thread 'test_asset_xcm_trader_refund_not_possible_since_amount_less_than_ed' panicked at '`mint_into` cannot generally fail; qed', /builds/parity/mirrors/cumulus/primitives/utility/src/lib.rs:198:25
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: <cumulus_primitives_utility::TakeFirstAssetTrader<AccountId,FeeCharger,Matcher,Assets,HandleRefund> as core::ops::drop::Drop>::drop
   3: std::thread::local::LocalKey<T>::with
   4: core::ops::function::FnOnce::call_once
   5: core::ops::function::FnOnce::call_once
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
failures:
    test_asset_xcm_trader_refund_not_possible_since_amount_less_than_ed
```

girazoki:
weird, they pass locally for me. I merged master back in, lets see if that fixes it

NachoPal:
> weird, they pass locally for me. I merged master back in, lets see if that fixes it

It is still failing. I reached the same results performing my own testing. I noticed that for Statemine the fees were being issued to the block author, however they weren't for Statemint.

The reason is that `mint_into` will fail if the `amount` is lower that `asset.min_balance` and the block author's `Account` does not exists (it does not hold at least the `min_balance`) https://github.com/paritytech/substrate/blob/master/frame/assets/src/functions.rs#L113-L140

The error is not handled here: https://github.com/girazoki/cumulus/blob/girazoki-statemint-runtimes-to-accept-sufficient-assets/primitives/utility/src/lib.rs#L197-L198

Currently, Statemine weights are hardcoded to `1_000_000_000` per XCM instructions, and the `WeightToFee` still returns values higher than `1_000` (`min_balance`) for USDT, but it will change as soon as we included XCM benchmarks where weights are way way lower. Subsequently, no fee issuance to block author will happen there either.

If we want it to work we would have to warn all block authors to have `min_balance` in their accounts to be able to receive the fee. If this does not happen, the total issuance of the asset would decreased.

girazoki:
Ah, I see, the reason why I wasnt detecting it is because I was running the test in release mode. It makes sense that in debug mode we reach that line.

However the test is precisely testing this case, so I am not sure what to do in this one. Should I just remove it?

Concerning block authors, they can always use the `touch` functionality in pallet-assets to make sure they are able to receive the payment. Otherwise we should warn them as you said that they need to have the minimum balance of the asset

NachoPal:
@girazoki For assets like USDT that are fully backed by collateral, we can't have it happen. We can not depend on trusting author blocks to have some `min_balance`.

We propose to increase fee to `asset.min_balance` in case the fee `amount` is lower than it. Notice that the check have to be done also for burning the assets.

Could you make those changes?

girazoki:
I think this can be done, yes. Will do that ASAP

girazoki:
Unrelated to this PR: Take into account that if someone transfers an amount below the ED of the particular asset this amount is burnt since it cannot be deposited anywhere. Is this an issue? Because I believe this is the case even without this PR

girazoki:
> No, it handles it: When the [amount sent is too low](https://github.com/paritytech/substrate/blob/0ba251c/frame/assets/src/functions.rs#L132) and when there is [dust left in the sender's account](https://github.com/paritytech/substrate/blob/0ba251c/frame/assets/src/functions.rs#L258).

But this refers to the source account I believe, and the source account does hold enough tokens to be burnt. Let's imagine the following situation: Sovereign account of parachain A has 1000 units of an asset. A user from parachain A wants to withdraw from such sovereign 5 units, and deposit them into a new account with 0 units. The existential deposit is 10 units. I think the following happens:

- 5 units are burnt from the sovereign (total supply at this point is 995).
- Assume BuyExecution is paid by KSM.
- 5 units are tried to be deposited into a new account and the deposit fails. Total supply now stays at 995

joepetrowski:
Oh, I see what you are saying. I thought you meant a problem with the Assets pallet when users transfer an asset.

This should try to cover that case and maintain total issuance. Maybe it can place those units into an asset trap that the parachain can reclaim into its sovereign account?

NachoPal:
> > No, it handles it: When the [amount sent is too low](https://github.com/paritytech/substrate/blob/0ba251c/frame/assets/src/functions.rs#L132) and when there is [dust left in the sender's account](https://github.com/paritytech/substrate/blob/0ba251c/frame/assets/src/functions.rs#L258).
> 
> But this refers to the source account I believe, and the source account does hold enough tokens to be burnt. Let's imagine the following situation: Sovereign account of parachain A has 1000 units of an asset. A user from parachain A wants to withdraw from such sovereign 5 units, and deposit them into a new account with 0 units. The existential deposit is 10 units. I think the following happens:
> 
> * 5 units are burnt from the sovereign (total supply at this point is 995).
> * Assume BuyExecution is paid by KSM.
> * 5 units are tried to be deposited into a new account and the deposit fails. Total supply now stays at 995

How would you do that? If for example you do a `WitdrawAsset` + `DepositAsset` it will fail for `DepositAsset`, and the assets in the holding registry from the `WithdrawAsset` will be trapped.

If you try to do an asset transfer instead, before burning anything it will check also if the beneficiary `can_increase()` or not.

NachoPal:
The problem here is introduced by the new trader `TakeFirstAssetTrader`.

Whereas for `UsingComponents` the fees are sent to the `StakingPot` without an issue, `TakeFirstAssetTrader` sometimes performs a failing `mint_into()`. The total issuance of the asset would be drained by all those failing fee minting.

NachoPal (approved):
LGTM

parachains/runtimes/assets/statemint/tests/tests.rs
```diff
@@ -0,0 +1,317 @@
...
+			let asset_multilocation = MultiLocation::new(
+				0,
+				X2(
+					PalletInstance(
+						<Runtime as frame_system::Config>::PalletInfo::index::<Assets>().unwrap()
+							as u8,
+					),
+					GeneralIndex(1),
+				),
+			);
```
gilescope, line 297: I feel like there's a macro to be had here maybe? 
Something like:
`mulitlocation_pallet!(Assets, instance)` (with the instance arg being optional and defaulting to 1)

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,157 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+	fn refund_weight(&mut self, weight: Weight) -> Option<MultiAsset> {
+		log::trace!(target: "xcm::weight", "TakeFirstAssetTrader::refund_weight weight: {:?}", weight);
+		let weight = weight.min(self.0);
+		if let Some((asset_location, local_asset_id)) = self.2.clone() {
+			// Calculate asset_balance
+			// This read should have already be cached in buy_weight
+			let asset_balance =
+				FeeCharger::charge_weight_in_fungibles(local_asset_id, weight).ok()?;
+
+			self.0 -= weight;
```
gilescope, line 143: Why shouldn't this be a saturating sub?

gilescope (approved):

parachains/runtimes/assets/statemint/tests/tests.rs
joepetrowski: This sounds like a follow up issue?

girazoki:
Please do not merge yet, there is a couple of things I would like to test first

girazoki:
Ok I added some additional checks, and adapted tests. I think now I am happy with it

gilescope (approved):
new changes look good to me.

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,163 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
 		}
 	}
 }
+
+pub struct TakeFirstAssetTrader<
```
KiChjang, line 65: Would really help if there are doc comments above this struct.

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,191 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
 		}
 	}
 }
+
+/// Contains information to handle refund/payment for xcm-execution
+#[derive(Clone, Eq, PartialEq, Debug)]
+struct AssetTraderRefunder<Balance, AssetId> {
+	// The amount of weight bought minus the weigh already refunded
+	weight: Weight,
```
KiChjang, line 69: Is this `weight_outstanding` then?

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,191 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+	// The balance paid for the weight bought minus the balance already refunded
+	balance: Balance,
```
KiChjang, line 71: `outstanding_balance`?

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,191 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+	// The multilocation representing the fungible asset in which fees were paid
+	location: MultiLocation,
```
KiChjang, line 73: This is assuming that the asset is `Concrete`, which might not be the case. I suppose this should really be an `AssetId` instead.

KiChjang, line 73: Is this representing the `MultiLocation` of the asset, or the location of the purchaser? If it's the former, that is already covered by the `AssetId` down below. If it's the latter, then this field should be renamed as `purchaser`.

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,191 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+/// Then charges for this assetId as described by FeeCharger
+/// Weight, paid balance, local asset Id and the multilocation is stored for
+/// later refund purposes
+/// Important: Errors if the Trader is being called twice by 2 BuyExecution instructions
+/// Alternatively we could just return payment in the aforementioned case
+pub struct TakeFirstAssetTrader<
+	AccountId,
+	FeeCharger: ChargeWeightInFungibles<AccountId, Assets>,
+	Matcher: MatchesFungibles<Assets::AssetId, Assets::Balance>,
+	Assets: fungibles::Mutate<AccountId> + fungibles::Transfer<AccountId> + fungibles::Balanced<AccountId>,
```
KiChjang, line 90:
```suggestion
	ConcreteAssets: fungibles::Mutate<AccountId> + fungibles::Transfer<AccountId> + fungibles::Balanced<AccountId>,
```

to better communicate intent.

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,191 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+		// Make sure we dont enter twice
+		if self.0.is_some() {
+			return Err(XcmError::NotWithdrawable)
+		}
+
+		// We take the very first multiasset from payment
+		let multiassets: MultiAssets = payment.clone().into();
+
+		// Take the first multiasset from the selected MultiAssets
+		let first = multiassets.get(0).ok_or(XcmError::TooExpensive)?;
```
KiChjang, line 128: While `TooExpensive` makes some sense, is there an XCM error that expresses something like "asset not found"? That seems to be a better fit here.

primitives/utility/src/lib.rs
girazoki: Alright what I did instead is store everything as a MultiAsset, now the struct contains only two fields: weight and outstanding_concrete_asset.

primitives/utility/src/lib.rs
```diff
@@ -51,3 +61,191 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+	// The balance paid for the weight bought minus the balance already refunded
+	outstanding_balance: Balance,
+	// The multilocation representing the fungible asset in which fees were paid
+	location: MultiLocation,
+	// The local assetId the aforementioned location corresponds to
+	asset: AssetId,
+}
+
+/// Charges for exercution in the first multiasset of those selected for fee payment
+/// Only succeeds for Concrete Fungible Assets
```
KiChjang, line 72: So the limitation here is purely because you store an `Option` field in your struct, correct? If so, perhaps we can just store a `Vec` instead? Because I can see that it's totally valid to use multiple `BuyExecution`s when one asset isn't enough to pay for fees, and another asset is provided to fill in the remaining deficit.

primitives/utility/src/lib.rs
```diff
@@ -51,3 +58,199 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+		AccountId: Clone,
+		ReceiverAccount: frame_support::traits::Get<Option<AccountId>>,
+	> TakeRevenue for XcmFeesToAccount<ConcreteAssets, Matcher, AccountId, ReceiverAccount>
+{
+	fn take_revenue(revenue: MultiAsset) {
+		match Matcher::matches_fungibles(&revenue) {
+			Ok((asset_id, amount)) =>
+				if let Some(receiver) = ReceiverAccount::get() {
+					if !amount.is_zero() {
+						let ok = ConcreteAssets::mint_into(asset_id, &receiver, amount).is_ok();
```
KiChjang, line 236: I'm not 100% sure about this mint, why can't we instead transfer the balance to the recipient account?

KiChjang:
I think in general if we want to productionize something, we can do so on Kusama (i.e. Statemine), but it's quite dangerous to do so immediately on Polkadot/Statemint until the Kusama network had some experience interacting with it. Thus my suggestion is to revert the changes on Statemint but keep the ones on Statemine. We'll revisit adding this functionality to Statemint when we are confident that it is running smoothly on Statemine.

primitives/utility/src/lib.rs
```diff
@@ -51,3 +58,199 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+					payment.checked_sub(required.clone()).map_err(|_| XcmError::TooExpensive)?;
+
+				// record weight, balance, multilocation and assetId
+				self.0 = Some(AssetTraderRefunder {
+					weight_outstanding: weight,
+					outstanding_concrete_asset: required,
+				});
+				Ok(unused)
+			},
+			_ => return Err(XcmError::TooExpensive),
```
KiChjang, line 152: I think we should also use `AssetNotFound` here as well.

Ideally, there should be some form of static assertion in this struct here that forbids anything other than a concrete asset type from being assigned as the 4th type parameter, there might be a way to do this but I'm currently blanking out right now...

parachains/common/src/xcm_config.rs
```diff
@@ -66,3 +72,58 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
...
+/// a given WeightToFee implementation an amount charged in
+/// a particular assetId from pallet-assets
+pub struct AssetFeeAsExistentialDepositMultiplier<R, WeightToFee, CON>(
+	PhantomData<(R, WeightToFee, CON)>,
+);
+impl<CurrencyBalance, R, WeightToFee, CON>
+	cumulus_primitives_utility::ChargeWeightInFungibles<AccountIdOf<R>, pallet_assets::Pallet<R>>
+	for AssetFeeAsExistentialDepositMultiplier<R, WeightToFee, CON>
+where
+	R: pallet_assets::Config,
```
KiChjang, line 86: This `R` here is the `Runtime` right? Let's use the full name here.

primitives/utility/src/lib.rs
girazoki: Because `WIthdrawAsset` burns the assets according to https://github.com/paritytech/polkadot/blob/f6831e91ab574d9e33bf75e9368e8f86a08b8dcb/xcm/xcm-builder/src/fungibles_adapter.rs#L252. Hence there is a "burnt" number of tokens we need to "mint" in the receiver account

primitives/utility/src/lib.rs
girazoki: Right now the only thing that I know of making sure you are buying enough weight is the `AllowTopLevelPaidExecutionFrom` barrier. Hence inserting a second  `BuyExecution` would not help for anything but giving away tokens (but I might be missing something)

primitives/utility/src/lib.rs
KiChjang: But isn't this a leaky abstraction? Minting something here assumes that we're using the fungibles adapter to handle assets, which might not always be the case.

If we need to make use of the entity that is used to withdraw assets from to deposit them back into an account, then I think the best idea here is to accept a `TransactAsset` type parameter to this struct, and make use of the `deposit_asset` method there.

It would also be a good idea to mention in the doc comments that the same `TransactAsset` type that is used to withdraw the fungible asset must be supplied here.

girazoki: Oh I see, yeah makes sense. Let me modify this

primitives/utility/src/lib.rs
girazoki: Although again this is making an assumption on a specific barrier being implemented. However, I find it hard to handle the `refund` case when you have multiple assets, in which one of them should we refund?

primitives/utility/src/lib.rs
girazoki: I finally reworked the logic so that this match isnt needed

primitives/utility/src/lib.rs
```diff
@@ -51,3 +58,206 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+					minimum_balance
+				} else {
+					amount
+				}
+			})?
+			.try_into()
+			.map_err(|_| XcmError::Overflow)?;
+
+		// Convert to the same kind of multiasset, with the required fungible balance
+		let required = first.clone().id.into_multiasset(asset_balance.into());
```
KiChjang, line 144: Seems like you can clone the `id` instead of the whole `first`?

primitives/utility/src/lib.rs
KiChjang: Hmm, so having this extra functionality does mean that `AllowTopLevelPaidExecutionsFrom` would have to change a bit in order to check whether other assets is enough to pay for the weights. I do not know if looking up other assets entail a database lookup, but if it does, then it's an additional problem because barrier conditions need to be lightweight to check, otherwise it's an attack vector.

In either case, this looks like a really tough problem to solve, so I don't mind if we simply restrict it to 1 `BuyExecution` for now and move on from here.

girazoki: perfect, I agree this is a tough problem with respect to DB reads

girazoki:
I reworked the refund process a little bit. We want to satisfy two main things for fully-collateralized assets:
- That the xcm fee payment is at least the ED. This makes sure that it will be payable to the receiver (or drop handler)
- That the refund process does not drop the xcm fee payment below ED. In this case what I did is, in case the refund would imply dropping the payment below the asset ED, just refund the difference. This way we make sure the payment always go through

primitives/utility/src/lib.rs
```diff
@@ -51,3 +58,212 @@ impl<T: UpwardMessageSender, W: WrapVersion> SendXcm for ParentAsUmp<T, W> {
...
+	fn new() -> Self {
+		Self(None, PhantomData)
+	}
+	// We take first multiasset
+	// Check whether we can convert fee to asset_fee (is_sufficient, min_deposit)
+	// If everything goes well, we charge.
+	fn buy_weight(
+		&mut self,
+		weight: Weight,
+		payment: xcm_executor::Assets,
```
KiChjang, line 109: Still trying to understand this, the `buy_weight` function takes ownership of `payment` and only takes the first `MultiAsset` out of it, and the rest is then dropped after this function ends? That doesn't sound right to me.

girazoki: The rest is not dropped, but rather returned (the unused part). Whatever returns from here is credited back to the holding register.

KiChjang: Where's the code for that? Is that from the `Drop` implementation of `xcm_executor::Assets`?

girazoki: Its part of the xcm-executor code for BuyExecution. https://github.com/paritytech/polkadot/blob/7e11707a55fcad806c71179c8cab6505cde65991/xcm/xcm-executor/src/lib.rs#L453.

Unless I am missing something, whatever we return from buy_weight should be credited to the holding

KiChjang (approved):
Eventually we should all attempt to move everything upstream to polkadot/xcm, but this'll do for now.

We'll also want to have a plan in mind when it comes to when we should release this on Polkadot.

joepetrowski:
> We'll also want to have a plan in mind when it comes to when we should release this on Polkadot.

We still don't have sufficient assets there, although it'd be nice to have this already in place when we do. But I don't think it needs any special logic - some time on Kusama plus audits before putting on Polkadot.

joepetrowski (approved):

parachains/runtimes/assets/statemint/tests/tests.rs
```diff
@@ -0,0 +1,166 @@
...
+	// collators to test block prod
+	collators: Vec<AccountId>,
+}
+
+impl Default for ExtBuilder {
+	fn default() -> ExtBuilder {
+		ExtBuilder { balances: vec![], collators: vec![] }
+	}
+}
+use frame_support::traits::GenesisBuild;
```
kianenigma, line 25: can't these builders be reused?

parachains/runtimes/assets/statemint/tests/tests.rs
```diff
@@ -0,0 +1,166 @@
...
+		ext.execute_with(|| {
+			System::set_block_number(1);
+		});
+
+		ext
+	}
+}
+
+use codec::Encode;
+use sp_runtime::{Digest, DigestItem};
```
kianenigma, line 80: nicer to move to top.

parachains/runtimes/assets/statemint/tests/tests.rs
```diff
@@ -0,0 +1,166 @@
+use frame_support::{
+	assert_noop, assert_ok, traits::PalletInfo, weights::WeightToFee as WeightToFeeT,
+};
```
joepetrowski, line 3:
```suggestion
use frame_support::{
	assert_noop, assert_ok,
	traits::{GenesisBuild, PalletInfo},
	weights::WeightToFee as WeightToFeeT,
};
use codec::Encode;
use sp_runtime::{Digest, DigestItem};
```

parachains/runtimes/assets/statemint/tests/tests.rs
joepetrowski, line 25:
```suggestion
```

parachains/runtimes/assets/statemint/tests/tests.rs
joepetrowski, line 80:
```suggestion
```

parachains/runtimes/assets/statemint/tests/tests.rs
girazoki: I can put the builder in common so that all runtimes can access them yes

girazoki: Moved the builders to asset-test-utils package, now they are shared across all statemine runtimes

kianenigma (approved):

joepetrowski:
bot merge

PR #1280, opened 2022-05-20
Ensure relay chain block number strictly increases

bkchr:
This is a safeguard for when async backing is enabled on the relay chain and multiple parachain
blocks per relay chain block are legal. We will need to later change this to support then multiple
parachain blocks per relay chain block.


xlc:
This means we cannot easily swap the relaychain of a testnet parachain, which is a very useful feature to maintain a long term testnet.

bkchr:
Stuff like the latest message head already prevent this (it may currently works for you because you don't have send any messages on your test chain).

However, I can make this configurable from the beginning.

xlc:
The latest message head can be fixed by setting storages on relaychain with sudo. We can't just bump block number on relaychain (or can we?)

rphmeier (approved):

bkchr:
> The latest message head can be fixed by setting storages on relaychain with sudo. We tried this with last rococo reset. We can't just bump block number on relaychain (or can we?)

No, but if I make it configurable, you can ignore this test on your test parachain.

bkchr:
> The latest message head can be fixed by setting storages on relaychain with sudo. We tried this with last rococo reset. We can't just bump block number on relaychain (or can we?)

No, but if I make it configurable, you can ignore this test on your test parachain.

gilescope (approved):

xlc:
> > The latest message head can be fixed by setting storages on relaychain with sudo. We tried this with last rococo reset. We can't just bump block number on relaychain (or can we?)
> 
> No, but if I make it configurable, you can ignore this test on your test parachain.

That works. Can we make it configurable? Should I make a new issue?

bkchr:
I will push the changes to this pr before merging it.

bkchr:
I have made it configurable. However, my naming is really shitty. @xlc or @rphmeier someone of you may have a better idea for the naming?

rphmeier (approved):

xlc:
Thanks. The name looks fine to me.

dvdplm (approved):

pallets/parachain-system/src/lib.rs
```diff
@@ -474,6 +529,11 @@ pub mod pallet {
 	#[pallet::storage]
 	pub(super) type DidSetValidationCode<T: Config> = StorageValue<_, bool, ValueQuery>;
 
+	/// The relay chain block number associated to the last parachain block.
```
andresilva, line 532:
```suggestion
	/// The relay chain block number associated with the last parachain block.
```

pallets/parachain-system/src/tests.rs
```diff
@@ -107,6 +107,7 @@ impl Config for Test {
 	type ReservedDmpWeight = ReservedDmpWeight;
 	type XcmpMessageHandler = SaveIntoThreadLocal;
 	type ReservedXcmpWeight = ReservedXcmpWeight;
+	type CheckAssociatedRelayNumber = AnyRelayNumber;
```
andresilva, line 110: Shouldn't this be `RelayNumberStrictlyIncreases` for the test below to work?

pallets/parachain-system/src/lib.rs
```diff
@@ -88,6 +88,51 @@ pub use relay_state_snapshot::{MessagingStateSnapshot, RelayChainStateProof};
 
 pub use pallet::*;
 
+/// Something that can check the associated relay block number.
+///
+/// Each Parachain block is build in the context of a relay chain block. This relay chain block
+/// has a block number associated this trait is about. With async backing it is legal to build
```
andresilva, line 94:
```suggestion
/// Each Parachain block is built in the context of a relay chain block, this trait allows us
/// to validate the given relay chain block number. With async backing it is legal to build
```

andresilva (commented):
lgtm just need to fix the test

pallets/parachain-system/src/lib.rs
andresilva: Although I think this part is a bit redundant.

PR #1283, opened 2022-05-23
Runtime version bump to 922

chevdor:
see also #1284


drahnr (approved):

joepetrowski:
This doesn't give us any room to bump a version if there's a patch fix.

Historically, v4 went to v400 to allow v401, v402, etc. and we've continued with 500, 600, ..., 900. We should either continue that with 1000 being the next release, or commit to following the Relay Chain runtime directly (probably the latter is better). 

Therefore, I'd make these either 1000 or 9220.

drahnr (approved):

PR #697, opened 2021-10-23
XCM v3 Companion

gavofyork:
## TODO

- [x] Ensure price for UMP message sending is configurable.
- [x] Ensure price for XCMP message sending is configurable.

PR #1286, opened 2022-05-23
Add event with XCM executor outcome, which includes weight fee

girazoki:
Throws an event with the outcome of the xcm executor, which in case of both Success or Error shows the weight used. The weight used information is useful for monitoring purposes, e.g., in case Fees are charged. I didnt want to disrupt the existing events, so I mainteined them.


girazoki:
I also prefer that. Let me change it to that

girazoki:
Modified

joepetrowski (commented):
Agree that this is useful, but does seem strange to partially change the `Event` variants to structs instead of tuples.  @KiChjang I think worked on updating events, should we combine that with this?

girazoki:
Is it this PR https://github.com/paritytech/cumulus/pull/1260/files? If so it does not look like it changed those events in xcmp-queue. I can switch back to tuples or change everything to named events, let me know which is preferred

joepetrowski:
Yeah exactly, looks like that PR missed this. I think going to all named events makes sense.

girazoki:
Updated!

joepetrowski (approved):
LGTM. I'd just like @KiChjang or @gavofyork to weigh in on the `message_id` naming w/r/t https://github.com/paritytech/polkadot/pull/4756 so that XCM v3 doesn't change the meaning of ID.

But based on the definition and usage of `id` in this [section in the UMP queue](https://github.com/paritytech/polkadot/blob/a2b16867f3cbed456e83929e4e08d9225db6f946/runtime/parachains/src/ump.rs#L93L148) I think it is OK (or we should use `sp_io::hashing::blake2_256` instead of `T::Hashing::hash`):

```rust
/// Returns a [`MessageId`] for the given upward message payload.
fn upward_message_id(data: &[u8]) -> MessageId {
	sp_io::hashing::blake2_256(data)
}

impl<XcmExecutor: xcm::latest::ExecuteXcm<C::Call>, C: Config> UmpSink for XcmSink<XcmExecutor, C> {
	fn process_upward_message(
		origin: ParaId,
		mut data: &[u8],
		max_weight: Weight,
	) -> Result<Weight, (MessageId, Weight)> {
		/* snip */

		let id = upward_message_id(&data[..]);

		/* snip */

		match maybe_msg_and_weight {
			Err(_) => {
				Pallet::<C>::deposit_event(Event::InvalidFormat(id));
				Ok(0)
			},
			Ok((Err(()), weight_used)) => {
				Pallet::<C>::deposit_event(Event::UnsupportedVersion(id));
				Ok(weight_used)
			},
			Ok((Ok(xcm_message), weight_used)) => {
				let xcm_junction = Junction::Parachain(origin.into());
				let outcome = XcmExecutor::execute_xcm(xcm_junction, xcm_message, id, max_weight);
				match outcome {
					Outcome::Error(XcmError::WeightLimitReached(required)) => Err((id, required)),
					outcome => {
						let outcome_weight = outcome.weight_used();
						Pallet::<C>::deposit_event(Event::ExecutedUpward(id, outcome));
						Ok(weight_used.saturating_add(outcome_weight))
					},
				}
			},
		}
	}
}
```

girazoki:
I think https://github.com/paritytech/cumulus/pull/697 changes it already to `sp_io::hashing::blake2_256`.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -263,21 +263,26 @@ pub mod pallet {
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// Some XCM was executed ok.
-		Success(Option<T::Hash>),
+		Success { message_id: Option<T::Hash>, weight: Weight },
```
KiChjang, line 266: I would still opt to call it the message hash instead of ID, because the hash isn't necessarily unique when two XCMs containing the exact same instructions will have the same hash, but not necessarily the same ID.

girazoki: sounds good, let me change that

pallets/xcmp-queue/src/lib.rs
```diff
@@ -15,7 +15,7 @@
 // along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
 
 //! A pallet which uses the XCMP transport layer to handle both incoming and outgoing XCM message
-//! sending and dispatch, queuing, signalling and backpressure. To do so, it implements:
+//! sending and dispatch, queuing, signalling and backpressure. To do so, it implements:https://www.tripadvisor.com/Restaurant_Review-g187497-d6216371-Reviews-Micu_Maku-Barcelona_Catalonia.html#photos;aggregationId=101&albumid=101&filter=7&ff=293721822
```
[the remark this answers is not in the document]
girazoki: thanks :)

KiChjang:
Can this merge into the XCMv3 branch instead?

girazoki:
@KiChjang can you merge master in https://github.com/paritytech/cumulus/pull/697=

KiChjang (approved):
Actually, it doesn't matter, XCMv3 can merge master right afterwards and it'll be fine.

PR #1289, opened 2022-05-24
sudo was removed from statemine/t but we were still pulling in the crate.

gilescope:
sudo was removed from statemine/t but we were still pulling in the crate.


chevdor (approved):
Thanks for taking care of it :)

joepetrowski (approved):

gilescope:
bot merge

PR #1296, opened 2022-05-25
deleted unnecessary repo checkout to fix fetch binary step

EgorPopelyaev:
Context: due to the renaming of the client from `polkadot-collator` to `polkadot-parachain`, fetch the binaries conflicted with the `polkadot-parachain` folder of the repo. But checking out the repo is not required for this workflow.

No repo => no conflict :)

PR #1299, opened 2022-05-25
bringing back repo checkout and adding a temp dir to avoid name duplication between binary and repo dir

EgorPopelyaev:
As it turned out we still do need to checkout a repo to be able to use scripts later in pipeline. That's why we need to introduce a temp dir where we download and run binary.

reverting changes from [1296](https://github.com/paritytech/cumulus/pull/1296)

### DIFF
--- a/.github/workflows/extrinsic-ordering-check-from-bin.yml
+++ b/.github/workflows/extrinsic-ordering-check-from-bin.yml
@@ -26,7 +26,7 @@ jobs:
       BIN_URL: ${{github.event.inputs.binary_url}}
       REF_URL: ${{github.event.inputs.reference_url}}
       BIN: node-bin
-      BIN_PATH: ./tmp/node-bin
+      BIN_PATH: ./tmp/$BIN
 
     steps:    
       - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
@@ -36,7 +36,7 @@ jobs:
           echo Creating a temp dir to download and run binary
           mkdir -p tmp
           echo Fetching $BIN_URL
-          wget $BIN_URL -O tmp/node-bin
+          wget $BIN_URL -O $BIN_PATH
           chmod a+x $BIN_PATH
           $BIN_PATH --version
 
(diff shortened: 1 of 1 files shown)


.github/workflows/extrinsic-ordering-check-from-bin.yml
```diff
@@ -26,9 +26,14 @@ jobs:
       BIN_URL: ${{github.event.inputs.binary_url}}
       REF_URL: ${{github.event.inputs.reference_url}}
 
-    steps:
+    steps:    
+      - uses: actions/checkout@v3
```
chevdor, line 30:
```suggestion
      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
```

chevdor (approved):
ok, once the Github action uses the hash for the version

sergejparity (approved):

coderobe (approved):

.github/workflows/extrinsic-ordering-check-from-bin.yml
```diff
@@ -25,23 +25,28 @@ jobs:
...
-    steps:
       - name: Fetch binary
         run: |
+          echo Creating a temp dir to download and run binary
+          mkdir -p tmp
           echo Fetching $BIN_URL
-          wget $BIN_URL
-          chmod a+x polkadot-parachain
-          ./polkadot-parachain --version
+          wget $BIN_URL -O tmp/node-bin
```
chevdor, line 38:
```suggestion
          wget $BIN_URL -O $BIN
```

.github/workflows/extrinsic-ordering-check-from-bin.yml
```diff
@@ -69,7 +74,7 @@ jobs:
 
       - name: Stop our local node
         run: |
-          pkill polkadot-parachain
+          pkill node-bin
```
chevdor, line 77: Ah it hurts having to use it here.
How about:
```
BIN=node-bin
BIN_PATH=./tmp/$BIN
```
that gives you all the VARS you need to NOT type `node-bin` anywhere but at the beginning when defining the VARS.

chevdor (changes requested):
If we use VARS, we should use them all the way

radupopa2010 (approved):

EgorPopelyaev:
bot merge

PR #1303, opened 2022-05-26
Remove redundant binary specification

athei:
The binary of our template node was named `parachain-collator` and the name of the actual node `polkadot-parachain`. This proofed to be confusing for people which mistakenly thought that `parachain-collator` would be "the real thing". Evidence of this is that the template was deployed for the contracts node making the contracts RPCs unavailable.

This PR renames the `parachain-collator` binary to `parachain-template-node` and removes the redundant explicit specification of the binary.


bkchr (approved):

gilescope:
Excellent <3

chevdor (approved):

parachain-template/polkadot-launch/config.json
```diff
@@ -17,7 +17,7 @@
 	},
 	"parachains": [
 		{
-			"bin": "../target/release/parachain-collator",
+			"bin": "../target/release/polkadot-parachain",
```
gilescope, line 20: as this is part of the template, should the bin being executed be `polkadot-template-node` ?

gilescope (approved):

parachain-template/polkadot-launch/config.json
gilescope, line 20:
```suggestion
			"bin": "../target/release/polkadot-template-node",
```

bernardoaraujor: `parachain-template-node`

athei: I think the rename as I did it is what we want. It was using the template but it was a mistake IMHO.

joepetrowski:
bot merge

PR #1307, opened 2022-05-30
Ensure we send the correct polkadot node version to telemetry

bkchr:
This is basically a hack, but with the current way we initialize the node this is the easiest
solution to ensure we send the correct version etc.

Closes: https://github.com/paritytech/substrate/issues/11541


koute (approved):
I've checked to make sure, and yes, after this PR it sends the polkadot version to the relaychain telemetry and the parachain version to the parachain telemetry.

Looks good to me!

PR #1310, opened 2022-05-30
bump txn versions due to reduced calls

EgorPopelyaev:
We need to bump transaction versions due to reduced calls. Based on Extrinsic ordering check: #1288

### DIFF
--- a/parachains/runtimes/assets/statemine/src/lib.rs
+++ b/parachains/runtimes/assets/statemine/src/lib.rs
@@ -87,7 +87,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 9220,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 6,
+	transaction_version: 7,
 	state_version: 0,
 };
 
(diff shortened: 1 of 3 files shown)


gilescope (approved):

parachains/runtimes/contracts/contracts-rococo/src/lib.rs
```diff
@@ -118,7 +118,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 9220,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 2,
```
joepetrowski, line 121: Is this necessary? @cmichi

hbulgarini (changes requested):
hold on, i think there is an issue with the dependencies on the [release-parachains-v9220](https://github.com/paritytech/cumulus/tree/release-parachains-v9220) which are pointing to master branch by mistake. 
This transaction version should happen on the next release.

EgorPopelyaev:
@hbulgarini Hector, should I close this PR then?

hbulgarini:
Ok! let's leave it open so we can reuse this change for the next release.

hbulgarini:
This change was already addressed in the https://github.com/paritytech/cumulus/pull/1340.

PR #1312, opened 2022-05-31
add benchmarks output dir creation

sergejparity:
Fix `benchmarks-ci.sh` failure `Error: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })`

Appears that `polkadot-parachain benchmark` failing when target output dir is not present on the filesystem.

### DIFF
--- a/scripts/ci/benchmarks-ci.sh
+++ b/scripts/ci/benchmarks-ci.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-steps=5
-repeat=2
+steps=50
+repeat=20
 chainName=$1
 
 benhcmarkOutput=./parachains/runtimes/$chainName/src/weights
--- a/scripts/ci/benchmarks-ci.sh
+++ b/scripts/ci/benchmarks-ci.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-steps=50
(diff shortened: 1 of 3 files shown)


scripts/ci/benchmarks-ci.sh
```diff
@@ -21,6 +21,8 @@ pallets=(
 	frame_system
 )
 
+mkdir -p $benhcmarkOutput
```
alvicsam, line 24:
```suggestion
mkdir -p $benhcmarkOutput || echo "Directory $benhcmarkOutput exists"
```

Otherwise it'll fail if directory exists

alvicsam (changes requested):

scripts/ci/benchmarks-ci.sh
sergejparity: No, it shouldn't because `-p` option takes care about it.

chevdor, line 24: I suggest we fix the typo as another PR fixes it and that will introduce conflicts.

```suggestion
mkdir -p $benchmarkOutput
```

chevdor: See #1300

PR #1313, opened 2022-05-31
fix benchmarks output dir creation

sergejparity:
Fix `benchmarks-ci.sh` failure `Error: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })`

Appears that `polkadot-parachain benchmark` failing when target output dir is not present on the filesystem.

Similar to https://github.com/paritytech/cumulus/pull/1312 but fixes `master`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -211,9 +211,9 @@ benchmarks:
   <<:                              *collect-artifacts
   <<:                              *benchmarks-refs
   script:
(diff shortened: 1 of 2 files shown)


chevdor (changes requested):
Same comment #1312

chevdor:
See also See #1300

sergejparity:
Closing as will be resolved by https://github.com/paritytech/cumulus/pull/1300

PR #1321, opened 2022-05-31
Zombienet add tests

pepoviola:
Hi @skunert, this add the missing integration test to `zombienet_test` and rename the `command` of the previous. Also, there is an small change in the test binary to allow to `load` the chain-spec from json because is needed to create a different genesis-state in the `migrate_solo_to_para` test.

*NOTE*: The `ci` configuration isn't included here, tracked in this issue: https://github.com/paritytech/ci_cd/issues/436

Thanks!


zombienet_tests/0005-migrate_solo_to_para.toml
```diff
@@ -0,0 +1,44 @@
...
+  command = "test-parachain"
+  args = ["-lparachain=debug"]
+
+[[parachains]]
+id = 2000
+cumulus_based = true
+add_to_genesis = false
+register_para = false
+[parachains.genesis.runtime.runtime_genesis_config.sudo]
+key = "5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty"
```
bkchr, line 35: Why do we need to override the sudo key?

zombienet_tests/0004-runtime_upgrade.toml
```diff
@@ -0,0 +1,35 @@
...
+
+chain = "rococo-local"
+
+  [[relaychain.nodes]]
+  name = "alice"
+  validator = true
+
+  [[relaychain.nodes]]
+  name = "bob"
+  image = "docker.io/paritypr/polkadot-debug:5236-0.9.18-c55660e9-be16bd72"
```
bkchr, line 14: Why do we need this image here?

pepoviola: Hi @bkchr, no this image should be set as `env var` by the ci job. The `ci` isn't ready yet, I will update this in all the files.

Thanks!

zombienet_tests/0005-migrate_solo_to_para.toml
pepoviola: The sudo key is updated just to update some value in the spec to produce a different `genesis-state`. The original test use this fn 

```rust
		// Set some random value in the genesis state to create a different genesis hash.
		.update_storage_parachain(|| {
			sp_io::storage::set(b"test", b"test");
		})
```
but for this overriding the sudo key works for the same porpouse.

Thanks!

bkchr: Ahh okay, yeah then we should at least keep the comment or add some second chain spec.

zombienet_tests/0004-runtime_upgrade.feature
```diff
@@ -0,0 +1,13 @@
...
+Network: ./0004-runtime_upgrade.toml
+Creds: config
+
+alice: is up
+bob: is up
+charlie: is up
+dave: is up
+alice: parachain 2000 is registered within 225 seconds
+charlie: reports block height is at least 5 within 250 seconds
+charlie: parachain 2000 perform upgrade with <wasm_binary_spec_version_incremented.rs.compact.compressed.wasm> within 200 seconds
```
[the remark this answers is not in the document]
skunert: How does this work in practice? We upload the to-be-upgraded runtime somewhere and enter a url here which would befetches?

skunert:
This looks good to me. @pepoviola title says WIP, what needs to be done here?
@bkchr after this is merged and CI is working, we want to completely remove the old integration tests, correct?

pepoviola:
> This looks good to me. @pepoviola title says WIP, what needs to be done here? @bkchr after this is merged and CI is working, we want to completely remove the old integration tests, correct?

Hi @skunert, yes. I will need to make a couple of changes and also add the test to `.gitlab-ci.yaml`. 
@bkchr is ok to use the image `polkadot-debug:master` for the relay chain nodes?

Thanks!

bkchr:
> @bkchr is ok to use the image `polkadot-debug:master` for the relay chain nodes?

IDK. What does this entails?

alvicsam (approved):
CI part looks good

pepoviola:
Hi @skunert, all tests are now migrated and working on gitlab.

- 0001-sync_blocks_from_tip_without_connected_collator
- 0002-pov_recovery
- 0003-full_node_catching_up
- 0004-runtime_upgrade
- 0005-migrate_solo_to_para

Let me know if you want to add more tests.

Thanks!

bkchr:
@pepoviola sounds really good! So, we can merge this now?

pepoviola:
> @pepoviola sounds really good! So, we can merge this now?

Hi @bkchr, thanks! Yes, we can merge now :)

sergejparity (approved):

skunert:
Nice work here 👍.
One thing though, I compared all the tests again and noticed that none of the zombienet tests test the RPC functionality.
For example, the original test https://github.com/paritytech/cumulus/blob/f7e5d70ef7e397868089d02bdd3069696c1c2ba4/test/service/tests/full_node_catching_up.rs#L83-L84 connects one node via RPC, this does not happen in the zombienet tests.

Missing for tests
- 001
- 002
- 003
 
If this is a technical limitation for now, we could merge nevertheless but can not deprecate the rust tests until we have this. As I understand both zombienet and the old ones would run on CI for now?

pepoviola:
> Nice work here 👍. One thing though, I compared all the tests again and noticed that none of the zombienet tests test the RPC functionality. For example, the original test
> 
> https://github.com/paritytech/cumulus/blob/f7e5d70ef7e397868089d02bdd3069696c1c2ba4/test/service/tests/full_node_catching_up.rs#L83-L84
> 
> connects one node via RPC, this does not happen in the zombienet tests.
> Missing for tests
> 
> * 001
> * 002
> * 003
> 
> If this is a technical limitation for now, we could merge nevertheless but can not deprecate the rust tests until we have this. As I understand both zombienet and the old ones would run on CI for now?

Hi @skunert, thanks for the feedback!. Those test (0001, 0002, 0003) check if the `parachain 2000 is registered` through the query interface using polakadotjs api.

```typescript
const parachains = (await api.query.paras.parachains<Vec<ParaId>>()) || [];
```
If you want we can add one extra rcp call like `api.rpc.system.chain()`

Thanks!

skunert:
Hey @pepoviola,
checking that the parachain is registered is not the problem here.

In test 0003, in the original rust test we start node _eve_ as full node with cli argument `--relay-chain-rpc-url <url_to_alice>:<alice_ws_port>` which tests the rpc functionality. So this is what we would need to do in zombienet too.

pepoviola:
> Hey @pepoviola, checking that the parachain is registered is not the problem here.
> 
> In test 0003, in the original rust test we start node _eve_ as full node with cli argument `--relay-chain-rpc-url <url_to_alice>:<alice_ws_port>` which tests the rpc functionality. So this is what we would need to do in zombienet too.

Ahh... my bad @skunert. Yes, we can modify the network config to test the `rpc` with that cli argument. I will change that and ping you again.

Thanks!!

pepoviola:
Hi @skunert, I add the logic to use `relay-chain-rpc-url` in zombienet and I updated the tests.

Thanks!

skunert (approved):
LGTM!

pepoviola:
Ping @skunert we need one more review to merge this one :)

bkchr (approved):

PR #1324, opened 2022-06-01
Companion to #11398

jsidorenko:
https://github.com/paritytech/substrate/pull/11398


parachains/runtimes/assets/statemint/src/lib.rs
```diff
@@ -381,6 +381,8 @@ impl InstanceFilter<Call> for ProxyType {
 					Call::Uniques(pallet_uniques::Call::clear_attribute { .. }) |
 					Call::Uniques(pallet_uniques::Call::clear_collection_metadata { .. }) |
 					Call::Uniques(pallet_uniques::Call::set_collection_max_supply { .. }) |
+					Call::Uniques(pallet_uniques::Call::set_price { .. }) |
+					Call::Uniques(pallet_uniques::Call::buy_item { .. }) |
```
joepetrowski, line 385: This proxy type `AssetOwner` is more for Collection owner/admin function calls. Since these calls can be used by anyone (who would probably not have this proxy type set up anyway) I think it doesn't make sense to include them. I think those users would just have the more general `Assets` proxy, under which this is already included since it includes all of Assets/Uniques.

shawntabrizi (approved):

joepetrowski (commented):
Just need to remove `set_price` and `buy_item` from the proxy in Westmint and Statemine, then it's good.

joepetrowski (approved):

PR #1332, opened 2022-06-06
Lower all Assets / NFT deposits by 10x

shawntabrizi:
This PR lowers all the deposits of Statemine, Statemint, Westmint, and Contracts-Rococo by 10x.

This was decided based on a high level analysis of the cost of NFT deposits compared to other platforms.

This can always be adjusted later back up if needed.

Also fixes an issue where `contracts-rococo` did not use Rococo deposits.

### DIFF
--- a/parachains/runtimes/assets/statemint/src/lib.rs
+++ b/parachains/runtimes/assets/statemint/src/lib.rs
@@ -240,7 +240,7 @@ impl pallet_transaction_payment::Config for Runtime {
 }
 
 parameter_types! {
-	pub const AssetDeposit: Balance = UNITS / 10; // 1 / 10 UNITS deposit to create asset
+	pub const AssetDeposit: Balance = 10 * UNITS; // 10 UNITS deposit to create fungible asset class
 	pub const AssetAccountDeposit: Balance = deposit(1, 16);
 	pub const ApprovalDeposit: Balance = EXISTENTIAL_DEPOSIT;
 	pub const AssetsStringLimit: u32 = 50;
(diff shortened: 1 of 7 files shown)


shawntabrizi:
TODO on this PR is to also reduce the base deposits which are not controlled by this formula

kianenigma:
Have you made sure that all deposits in all pallets are recorded, so changing this does not break any operation?

gilescope (approved):

shawntabrizi:
> Have you made sure that all deposits in all pallets are recorded, so changing this does not break any operation?

Is that my responsibility, or the responsibility of all Pallet developers who are writing pallets?

What would be the action item here?

shawntabrizi:
Base deposits are now also updated. This should be good to go.

shawntabrizi:
@kianenigma checked:

* assets
* multisig
* proxy
* uniques


And these seem to be all the affected pallets.

joepetrowski:
Would be good to add a `fn redeposit` to the Assets pallet similar to [Uniques](https://github.com/paritytech/substrate/blob/3823ec4a3153a8b53a0effea3ad5a9191b46a153/frame/uniques/src/lib.rs#L618).

joepetrowski (dismissed):

parachains/runtimes/assets/statemint/src/lib.rs
```diff
@@ -537,8 +537,8 @@ impl pallet_asset_tx_payment::Config for Runtime {
 }
 
 parameter_types! {
-	pub const CollectionDeposit: Balance = UNITS; // 1 UNIT deposit to create asset class
-	pub const ItemDeposit: Balance = UNITS / 100; // 1/100 UNIT deposit to create asset instance
+	pub const CollectionDeposit: Balance = UNITS / 10; // 1 / 10 UNIT deposit to create asset class
```
joepetrowski, line 540: This is already really low compared to the fungible asset class. Perhaps just leave this?

bkchr (approved):

shawntabrizi:
bot merge

PR #1333, opened 2022-06-06
Bounded ping pallet

KiChjang:


parachains/pallets/ping/src/lib.rs
```diff
@@ -55,7 +60,11 @@ pub mod pallet {
 
 	/// The target parachains to ping.
 	#[pallet::storage]
-	pub(super) type Targets<T: Config> = StorageValue<_, Vec<(ParaId, Vec<u8>)>, ValueQuery>;
+	pub(super) type Targets<T: Config> = StorageValue<
+		_,
+		BoundedVec<(ParaId, BoundedVec<u8, MaxPayloadSize>), MaxParachains>,
```
ggwpez, line 65: Why is this bound with `MaxParachains`? Can one parachain not initiate multiple pings?

ggwpez (approved):

parachains/pallets/ping/src/lib.rs
KiChjang: It can, but this isn't what the Vec represents, it simply stores the known list of target parachains that can receive a ping/return a pong, and so the maximum number of targets is naturally the "maximum" number of parachains that the relay chain can support, which is 100.

gilescope (approved):
LGTM

shawntabrizi (approved):

shawntabrizi:
bot merge

PR #1346, opened 2022-06-11
Add Collectives Parachain

joepetrowski:
TODOs done, ready for review.

- [x] Impl slash handler to teleport to the Treasury @muharem  (https://github.com/paritytech/cumulus/pull/1433)
- [x] Add chain specs for testing locally and on Westend @gilescope ( https://github.com/paritytech/cumulus/pull/1441 )
- [x] Rename/refactor `statemint_build_import_queue` to be generic for other paras @gilescope (https://github.com/paritytech/cumulus/pull/1389)
- [x] Ditto for `start_statemint_node` @gilescope (https://github.com/paritytech/cumulus/pull/1402)
- [x] Include [benchmarks](https://github.com/paritytech/cumulus/blob/3de8e69a4f429d671ecef9b2b4b72283d70edd0f/parachains/runtimes/collectives/collectives-polkadot/src/lib.rs#L629) (https://github.com/paritytech/cumulus/pull/1427)


parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -0,0 +1,833 @@
...
+					Call::Utility { .. } | Call::Multisig { .. }
+			),
+		}
+	}
+	fn is_superset(&self, o: &Self) -> bool {
+		match (self, o) {
+			(x, y) if x == y => true,
+			(ProxyType::Any, _) => true,
+			(_, ProxyType::Any) => false,
+			(ProxyType::NonTransfer, _) => true,
```
gilescope, line 311: Why this NonTrarsfer line? It's not in statemint... I'm thinking it should be in both or neither.

gilescope (changes requested):
Looks good. It's a thumb's up from me once the TODOs in the code are addressed.

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
joepetrowski: Because in Statemint, `NonTransfer` is [not a superset](https://github.com/paritytech/cumulus/blob/release-parachains-v9230/parachains/runtimes/assets/statemint/src/lib.rs#L351) to the other proxy types.

```rust
ProxyType::NonTransfer =>
	!matches!(c, Call::Balances { .. } | Call::Assets { .. } | Call::Uniques { .. }),
```

However, Statemint should indeed add a `(ProxyType::NonTransfer, ProxyType::CollatorSelection) => true,` entry.

gilescope: Ok, but should the line above be `(ProxyType::NonTransfer, o) => o != ProxyType::Any` for soundness of supertypes?
(I'm assuming we don't want NonTransfer and Any to both be supertypes of each other... that would feel to me like a footgun)

joepetrowski: But `match` goes in order and short circuits, right? So `(_, ProxyType::Any) => false,` would prevent any footguns w/r/t `Any`.

gilescope: It's complicated: https://www.reddit.com/r/rust/comments/pqnufj/rust_match_arm_priority/
As it turns out they do go in order because your using the guard clauses. Without the guard clauses it could optimise it in some fun ways.

bkchr: It must always use a fixed order when there is any `_` involved. There also doesn't seem to be any official documentation. There only is this reddit thread, where all are just guessing and don't have any information.

gilescope: I'm just going to come out with it: Basti's right. From the rust reference it's going to use top down ordering (or anything that optimises to an equivalent of that).

.gitlab-ci.yml
```diff
@@ -265,48 +265,71 @@ benchmarks-build:
...
-    - export BRANCHNAME="weights-${CI_COMMIT_BRANCH}"
-    # Set git config
+    - ./scripts/benchmarks-ci.sh assets statemine ./artifacts
+    - ./scripts/benchmarks-ci.sh assets statemint ./artifacts
+    - ./scripts/benchmarks-ci.sh assets westmint ./artifacts
+    - export BRANCHNAME="weights-statemint-${CI_COMMIT_BRANCH}"
+    - *git-commit-push
+    # create PR
+    - curl -u ${GITHUB_USER}:${GITHUB_TOKEN}
+      -d '{"title":"[benchmarks] Update weights for statemine/t","body":"This PR is generated automatically by CI.","head":"'$BRANCHNAME'","base":"'${CI_COMMIT_BRANCH}'"}'
```
rcny, line 299: @alvicsam What does `/t` mean here?

rcny (approved):

gilescope (approved):

parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -0,0 +1,189 @@
...
+	// Use this currency:
+	Balances,
+	// Use this currency when it is a fungible asset matching the given location or name:
+	IsConcrete<DotLocation>,
+	// Convert an XCM MultiLocation into a local account id:
+	LocationToAccountId,
+	// Our chain's account ID type (we can't get away without mentioning it explicitly):
+	AccountId,
+	// We don't track any teleports of `Balances`.
+	(),
```
KiChjang, line 74: Honestly, this `CheckingAccount` parameter still trips me up, and I think we really need to document it better to say that the entire purpose of it is to track _all_ outgoing and incoming teleported assets in an account, and the account is basically inaccessible anywhere else.

It is then possible to put `()` here to say that we don't need to track any, but I needed to go through a few hoops before it clicks for me.

KiChjang: What is even more bizarre to me then is to see that `FungiblesAdapter` doesn't have an `Option` for the `CheckedAccount`, and it's something that I still haven't figured out why yet.

parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -0,0 +1,189 @@
...
+		MultiLocation { parents: 1, interior: Here } |
+		MultiLocation { parents: 1, interior: X1(_) }
+	};
+}
+
+pub type Barrier = DenyThenTry<
+	DenyReserveTransferToRelayChain,
+	(
+		TakeWeightCredit,
+		AllowTopLevelPaidExecutionFrom<Everything>,
```
KiChjang, line 122: It feels to me that this should be placed at the very end of the tuple, so that it's checked at the last position. The logic here is that the ordering should go from the more restrictive rules to the lesser ones.

sergejparity (approved):

parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -0,0 +1,191 @@
...
+pub struct XcmConfig;
+impl xcm_executor::Config for XcmConfig {
+	type Call = Call;
+	type XcmSender = XcmRouter;
+	type AssetTransactor = CurrencyTransactor;
+	type OriginConverter = XcmOriginToTransactDispatchOrigin;
+	// Collectives does not recognize a reserve location for any asset. Users must teleport DOT
+	// where allowed (e.g. with the Relay Chain).
+	type IsReserve = ();
+	type IsTeleporter = NativeAsset; // <- should be enough to allow teleportation of DOT
```
KiChjang, line 143: This would also allow any other chain to teleport their native token to the collectives parachain. I think to really reduce the liability of the collectives parachain, we really should only be accepting the DOT/KSM token from the relay chain, and nothing else.

parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -0,0 +1,191 @@
...
+);
+
+impl pallet_xcm::Config for Runtime {
+	type Event = Event;
+	// We want to disallow users sending (arbitrary) XCMs from this chain.
+	type SendXcmOrigin = EnsureXcmOrigin<Origin, ()>;
+	type XcmRouter = XcmRouter;
+	// We support local origins dispatching XCM executions in principle...
+	type ExecuteXcmOrigin = EnsureXcmOrigin<Origin, LocalOriginToLocation>;
+	// ... but disallow generic XCM execution. As a result only teleports and reserve transfers are allowed.
```
KiChjang, line 175:
```suggestion
	// ... but disallow generic XCM execution. As a result only teleports are allowed.
```

parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -0,0 +1,191 @@
...
+	type XcmExecuteFilter = Nothing;
+	type XcmExecutor = XcmExecutor<XcmConfig>;
+	type XcmTeleportFilter = Everything;
+	type XcmReserveTransferFilter = Nothing; // This parachain is not meant as a reserve location.
```
KiChjang, line 179:
```suggestion
	type XcmReserveTransferFilter = Nothing; // This parachain is not meant to receive reserve asset transfers.
```

parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -0,0 +1,191 @@
...
+		MultiLocation { parents: 1, interior: X1(Plurality { id: BodyId::Executive, .. }) }
+	};
+	pub type ParentOrSiblings: impl Contains<MultiLocation> = {
+		MultiLocation { parents: 1, interior: Here } |
+		MultiLocation { parents: 1, interior: X1(_) }
+	};
+}
+
+pub type Barrier = DenyThenTry<
+	DenyReserveTransferToRelayChain,
```
KiChjang, line 119: This tripped me up for a moment, because I've been describing the `Barrier` as a firewall, but it in fact isn't -- it's a list of _preconditions_ before an XCM can even get executed, and so it's not just incoming XCM that needs to go through the `Barrier`; locally executed XCMs also need to go through the `Barrier` as well.

joepetrowski: Yeah, I had trouble with it before, see: https://github.com/paritytech/cumulus/pull/1169#pullrequestreview-942155751 (and Nacho's response)

parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
```diff
@@ -0,0 +1,83 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub mod account {
+	use frame_support::PalletId;
+	use sp_runtime::AccountId32;
+
+	/// Relay Chain treasury pallet id, used to convert into AccountId
+	pub const RELAY_TREASURY_PALL_ID: PalletId = PalletId(*b"py/trsry");
```
kianenigma, line 21: Someday these should go in a central repository (onchain) of all chain metadatas, so no one needs to hardcode them anymore -- discussed in the parachain summit as well.

kianenigma: At the minimum, can't we move things that are constant like this to `polkadot_core_primitives`?

parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
```diff
@@ -0,0 +1,83 @@
...
+	/// account used to temporarily deposit slashed imbalance before teleporting
+	pub const SLASHED_IMBALANCE_ACC_ID: AccountId32 = AccountId32::new([7u8; 32]);
+}
+
+pub mod currency {
+	use polkadot_core_primitives::Balance;
+	use polkadot_runtime_constants as constants;
+
+	/// The existential deposit. Set to 1/10 of its parent Relay Chain (v9200).
```
kianenigma, line 30:
```suggestion
	/// The existential deposit. Set to 1/10 of its parent Relay Chain.
```

Not sure if the version meant anything here?

parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
```diff
@@ -0,0 +1,83 @@
...
+	pub const EXISTENTIAL_DEPOSIT: Balance = constants::currency::EXISTENTIAL_DEPOSIT / 10;
+
+	pub const UNITS: Balance = constants::currency::UNITS;
+	pub const DOLLARS: Balance = constants::currency::DOLLARS;
+	pub const CENTS: Balance = constants::currency::CENTS;
+	pub const MILLICENTS: Balance = constants::currency::MILLICENTS;
+
+	pub const fn deposit(items: u32, bytes: u32) -> Balance {
+		// 1/100 of Polkadot v9200
```
kianenigma, line 39:
```suggestion
		// 1/100 of Polkadot. 
```

parachains/runtimes/collectives/collectives-polkadot/src/impls.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Implements `OnUnbalanced::on_unbalanced` to teleport slashed assets to relay chain treasury account.
+pub struct ToParentTreasury<TreasuryAcc, TempAcc, T, I = ()>(
+	PhantomData<(TreasuryAcc, TempAcc, T, I)>,
+);
+
+impl<TreasuryAcc, TempAcc, T, I: 'static> OnUnbalanced<NegativeImbalanceOf<T, I>>
+	for ToParentTreasury<TreasuryAcc, TempAcc, T, I>
+where
+	TreasuryAcc: Get<AccountIdOf<T>>,
+	TempAcc: Get<AccountIdOf<T>>,
```
kianenigma, line 51: If this is going to be `SLASHED_IMBALANCE_ACC_ID`, I don't think it is needed to make it a configurable generic anymore, but as you wish.

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -0,0 +1,779 @@
...
+	type ExistentialDeposit = ExistentialDeposit;
+	type AccountStore = System;
+	type WeightInfo = weights::pallet_balances::WeightInfo<Runtime>;
+	type MaxReserves = ConstU32<50>;
+	type ReserveIdentifier = [u8; 8];
+}
+
+parameter_types! {
+	/// Relay Chain `TransactionByteFee` / 10
+	pub const TransactionByteFee: Balance = 1 * MILLICENTS;
```
kianenigma, line 205: This is not really guaranteed, as I recall you have access to relay chain constant, perhaps it can be written better.

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -0,0 +1,779 @@
...
+	PartialEq,
+	Ord,
+	PartialOrd,
+	Encode,
+	Decode,
+	RuntimeDebug,
+	MaxEncodedLen,
+	scale_info::TypeInfo,
+)]
+pub enum ProxyType {
```
kianenigma, line 266: Just thinking out loud: re-writing all of this HIGHLY risky logic every time seems very dangerous to me. Ideally, we would have a base proxy builder that constructs the default `NonTransfer`, `Any`, `Cancel` etc, and we only extend it with the custom ones.

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -0,0 +1,779 @@
...
+	/// Can execute any call that does not transfer funds.
+	NonTransfer,
+	/// Proxy with the ability to reject time-delay proxy announcements.
+	CancelProxy,
+	/// Collator selection proxy. Can execute calls related to collator selection mechanism.
+	Collator,
+	/// Alliance proxy. Allows calls related to the Alliance.
+	Alliance,
+}
+impl Default for ProxyType {
```
kianenigma, line 278: Why? shouldn't default be actually the most restrained variant?

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -0,0 +1,779 @@
...
+}
+
+parameter_types! {
+	pub const PotId: PalletId = PalletId(*b"PotStake");
+	pub const SessionLength: BlockNumber = 6 * HOURS;
+	pub const ExecutiveBody: BodyId = BodyId::Executive;
+}
+
+/// We allow root and the Relay Chain council to execute privileged collator selection operations.
+pub type CollatorSelectionUpdateOrigin =
```
kianenigma, line 402: fwiw you are using the exact same thing for `xcmp_queue::ControllerOrigin`

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -0,0 +1,779 @@
...
+
+// Create the runtime by composing the FRAME pallets that were previously configured.
+construct_runtime!(
+	pub enum Runtime where
+		Block = Block,
+		NodeBlock = opaque::Block,
+		UncheckedExtrinsic = UncheckedExtrinsic,
+	{
+		// System support stuff.
+		System: frame_system::{Pallet, Call, Config, Storage, Event<T>} = 0,
```
kianenigma, line 492: my eyes! let's use the new syntax without needing to name all of the parts? 🫣

kianenigma (approved):
I have a bunch of comments that you probably want to address before going forward, but no blockers. LGTM

parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
joepetrowski: Yeah, @muharem and I discussed several options for this. We agreed that for now hardcode it to avoid waiting for a Polkadot release, but that these can go into crates similar to the Balances (ED, etc.) and be imported.

parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
joepetrowski: Just to reference when this value was taken.

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
joepetrowski: Kind of related to your comment on pre-defining common types. The default type is `Any` in Polkadot: https://github.com/paritytech/polkadot/blob/release-v0.9.27-1/runtime/polkadot/src/lib.rs#L1134L1138

However, I think it's hard to define "most restrained", because besides `Any` and `NonTransfer` they tend to be non-overlapping sets of call access.

parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
joepetrowski: Ah, I see what you mean :facepalm: . I wrote this comment in Statemint before `polkadot_runtime_constants` existed and we had just hardcoded it in. And it just survived the addition of this crate.

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
joepetrowski: Looks like it's not exported in `constants`. Perhaps a good one to add there to just import it.
https://github.com/paritytech/polkadot/blob/release-v0.9.27-1/runtime/polkadot/constants/src/lib.rs

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
joepetrowski: Also proxies allow addition/modification to less privileged ones, but not more. When creating an anon proxy you would almost always want `Any`, otherwise you could never spend from it.

bkchr: It must be the most permissive value: https://github.com/paritytech/substrate/blob/a1a9b475c354d873f91135ed5fa028ac9ef6a2c4/frame/proxy/src/lib.rs#L127

IMO, we should introduce some custom trait for this, where the function name already expresses on what should be returned.

paritytech-cicd-pr:
 The job name - test-linux-stable 
 The job logs - https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/1735685

joepetrowski:
bot merge

PR #1347, opened 2022-06-11
Split chain_spec to individual chain families

gilescope:
(Easiest to review this by commit.)

1st commit: No code changes, just pushing functions into separate files so that we can more sensibly handle more chains.
2nd commit: `cargo fmt`
3rd commit: Replace many `if`s with `matches`.


polkadot-parachain/src/command.rs
```diff
@@ -40,55 +40,54 @@ use sp_core::hexdisplay::HexDisplay;
...
-	fn is_statemine(&self) -> bool;
-	fn is_westmint(&self) -> bool;
-	fn is_contracts_rococo(&self) -> bool;
+enum Runtime {
+	Shell,
+	Seedling,
+	Statemint,
+	Statemine,
+	Westmint,
+	ContractsRococo,
```
gilescope, line 49: Maybe there should be a `Generic` enum arm here and if so we can remove the need to wrap this in an option.

gilescope: Yeah, that's a tad nicer I think.

polkadot-parachain/src/command.rs
```diff
@@ -40,54 +40,51 @@ use sp_core::hexdisplay::HexDisplay;
...
-trait IdentifyChain {
-	fn is_shell(&self) -> bool;
-	fn is_seedling(&self) -> bool;
-	fn is_statemint(&self) -> bool;
-	fn is_statemine(&self) -> bool;
-	fn is_westmint(&self) -> bool;
-	fn is_contracts_rococo(&self) -> bool;
+enum Runtime {
+	/// This is the default runtime (based on rococo)
+	Generic,
```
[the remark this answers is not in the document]
muharem: What if we just name it,  `DefaultRococoParachain`
So the enum itself will tell as whats expected for default

polkadot-parachain/src/chain_spec/contracts.rs
```diff
@@ -0,0 +1,279 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Cumulus is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
```
muharem, line 15: How about having some docs for new files on the file level?

polkadot-parachain/src/command.rs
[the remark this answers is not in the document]
gilescope: I'm not sure that we want to say rococo is the default. I just wanted to make it clear that historically the defaults came from rococo.

muharem (approved):

joepetrowski (approved):

joepetrowski:
bot merge

PR #1350, opened 2022-06-12
Co #11456: Match statement fallthrough for `benchmark` commands

ggwpez:
Currently each new benchmarking command in Substrate triggers a change in Cumulus.  
After this MR there is no need for a hard-companion anymore, since the default behaviour will return an error for "unsupported command".  

Useful for https://github.com/paritytech/substrate/pull/11456 since adding the `benchmark extrinsic` logic itself is highly runtime-dependent and (currently) not useful to have.


gilescope (approved):
LGTM

ggwpez:
bot merge

PR #1367, opened 2022-06-14
Companion for substrate#11618

hzy1919:
paritytech/substrate#11618


kianenigma (approved):

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -556,7 +557,7 @@ construct_runtime!(
 
 		// Monetary stuff.
 		Balances: pallet_balances::{Pallet, Call, Storage, Config<T>, Event<T>} = 10,
-		TransactionPayment: pallet_transaction_payment::{Pallet, Storage} = 11,
+		TransactionPayment: pallet_transaction_payment::{Pallet, Storage, Event<T>} = 11,
 		AssetTxPayment: pallet_asset_tx_payment::{Pallet} = 12,
```
joepetrowski, line 561: One question, if someone paid a transaction fee with an asset, would this still emit an event? I.e., should we also add `Event<T>` to `AssetTxPayment`?

I know it could be a bit beyond the scope of strictly a `pallet_transaction_payment` event, but I think it would be confusing for users to see events for one type of fee and not for another. Would it make sense to expand https://github.com/paritytech/substrate/pull/11618 to include an event for `pallet_asset_tx_payment` as well?

shawntabrizi: yes, sure. Lets do a follow up PR

shawntabrizi:
bot merge

parachains/runtimes/assets/statemine/src/lib.rs
hzy1919: good , I would like to continue the job.

PR #1373, opened 2022-06-15
Updated weights

gilescope:
(manually on BM2 from 14 june 11am GMT start. No one else was using BM2 at the time)


muharem (dismissed):

NachoPal (approved):

joepetrowski (approved):

hbulgarini (approved):

gilescope:
So generally weights the same with slight changes:

- Batch utils have gone back to expected values. 
- Slight increase from multisig pallet (`as_multi_threshold_1`),
- frame_system: set_metadata is heavier on the passed in params.
- Uniques has changed slightly (extra write due to CollectionMaxSupply when destroy called)

PR #1374, opened 2022-06-15
Cheaper uniques 10x direction fix

gilescope:
Fix for #1332


joepetrowski:
No, this was correct.

PR #1402, opened 2022-06-23
rename now shared node creation function

gilescope:
We're going to reuse the same node construction for as many common good parachains as we can (unless there's a particular reason not to).

### DIFF
--- a/polkadot-parachain/src/service.rs
+++ b/polkadot-parachain/src/service.rs
@@ -1165,8 +1165,9 @@ where
 	))
 }
 
-/// Start a statemint/statemine/westmint parachain node.
-pub async fn start_statemint_node<RuntimeApi, AuraId: AppKey>(
+/// Start an aura powered parachain node.
+/// (collective-polkadot and statemine/t use this)
+pub async fn start_generic_aura_node<RuntimeApi, AuraId: AppKey>(
 	parachain_config: Configuration,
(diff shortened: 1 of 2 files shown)


polkadot-parachain/src/service.rs
```diff
@@ -1165,8 +1165,9 @@ where
 	))
 }
 
-/// Start a statemint/statemine/westmint parachain node.
-pub async fn start_statemint_node<RuntimeApi, AuraId: AppKey>(
+/// Start an aura powered parachain node.
+/// (collective-polkadot and statemine/t use this)
```
muharem, line 1169: wonder if it worth on the api side to mention the clients.
hard to maintain and gets outdated.
meanwhile easy to check the usage/clients within one repo.

bkchr (approved):

muharem (approved):

polkadot-parachain/src/service.rs
gilescope: I want to make it clear that if you mess with this function you're going to be impacting parachains X, Y and Z. If we get to the point where we have lots of parachains using this and everyone knows to be uber-careful with this function then we can delete the comment. But for now I'd like to have it there so people don't have to figure this out for themselves.

gilescope:
bot merge

PR #1407, opened 2022-06-24
Less code in magic macros

gilescope:
Use the ConstU32 types and friends where possible to have the least amount of code inside `parameter_types` macros.

Interestingly SS58Prefix is defined as a u8 and used as a u16. 

(If there was a `ConstBalance` type then that would remove the need for half the remaining `parameter_types` macros.)


davxy (approved):
LGTM

joepetrowski (approved):

joepetrowski:
Can you fix tests (think just `fmt`), then I can merge it in.

gilescope:
Sure thing.

gilescope:
bot merge

PR #1411, opened 2022-06-27
Fix GHA hash

chevdor:


coderobe:
would be useful to know *why* these hashes have changed. if it is a version update, the comment should also be updated

coderobe:
oh i see, it is removing a leading v - all good then

coderobe (approved):

TriplEight (approved):

chevdor:
bot merge

PR #1420, opened 2022-06-30
Second part of the Docker CI fix

chevdor:
This part being missing changes to the previous fix.

### DIFF
--- a/.github/workflows/release-10_docker-manual.yml
+++ b/.github/workflows/release-10_docker-manual.yml
@@ -13,37 +13,29 @@ on:
         required: true
       prerelease:
         description: is prerelease
-        default: false
+        default: "false"
         required: true
 
 jobs:
-  prepare:
-    runs-on: ubuntu-latest
-    env:
-      TMP: tmp
-    steps:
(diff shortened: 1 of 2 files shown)


alvicsam (approved):

chevdor:
bot merge

.github/workflows/release-10_docker-manual.yml
```diff
@@ -13,25 +13,34 @@ on:
         required: true
       prerelease:
         description: is prerelease
-        default: false
+        default: "false"
```
sergejparity, line 16: Just curious why the double quotes needed here?

chevdor: I was curious as well :)
`true` is a boolean and considered as such unless you quote it. So using `false` will not fail but GH will NOT load the default and the user is left with the work. Using `"false"` however makes GH happy as it takes the **string** `false` and uses it as default.

chevdor: So without quoting, you get:
image

instead of:
image

PR #1424, opened 2022-07-04
Add trie migration rpc to polkadot-parachain

cheme:
This PR is adding state trie migration rpc to polkadot-parachain.
The rpc would be usefull for migration (we are testing statemine on versi and we miss it).
Not 100% sure if it should be in master cc @bkchr .


bkchr (approved):

bkchr:
If we require this, let's add the RPC.

cheme:
It is not strictly required but would avoid using a custom build to follow progress with the rpc.

kianenigma (approved):

cheme:
bot merge

PR #1426, opened 2022-07-05
Adding integration tests

NachoPal:
Integration tests for parachains


.gitignore
```diff
@@ -8,3 +8,4 @@ polkadot_argument_parsing
 **/chains/
 *.iml
 .env
```
chevdor, line 10: is this `.env` just a helper or required to run the tests?
If so, I would suggest the following and provide an example.

```suggestion
.env
!.env-example
```

parachains/integration-tests/statemine/xcm/2_ump.yml
```diff
@@ -0,0 +1,193 @@
...
+                    { v1: [ *ap_ksm_fungible ] }, # assets
+                    0, # feeAssetItem
+                    { unlimited: true } # weightLimit
+                  ]
+                  events:
+                    - name: polkadotXcm.Attempted
+                      attribute:
+                        type: XcmV2TraitsOutcome
+                        isError: true
+                        value: Barrier
```
chevdor, line 193:
```suggestion
                        value: Barrier

```

parachains/integration-tests/statemine/xcm/3_hrmp-open-channels.yml
```diff
@@ -0,0 +1,405 @@
...
+                      attributes:
+                        type: XcmV2TraitsOutcome
+                        isComplete: true
+                        value: 4,000,000
+                    - name: hrmp.OpenChannelAccepted
+                      chain: *relay_chain
+
+
+            
+
```
chevdor, line 405:
```suggestion
                      chain: *relay_chain

```

parachains/integration-tests/statemine/xcm/4_hrmp.yml
```diff
@@ -0,0 +1,350 @@
...
+                        }
+                      ]
+                    }
+                  ]
+                  events:
+                    - name: xcmpQueue.Fail
+                      chain: *assets_parachain
+                      attributes:
+                        type: XcmV2TraitsError
+                        value: FailedToTransactAsset
```
chevdor, line 350:
```suggestion
                        value: FailedToTransactAsset

```

parachains/integration-tests/statemine/xcm/1_dmp.yml
```diff
@@ -0,0 +1,250 @@
...
+                    ]
+          - name: Should keep the balance of the receiver
+            actions:
+              - asserts:
+                  equal:
+                    args:
+                      [
+                        $balance_ap_receiver_before,
+                        $balance_ap_receiver_after
+                      ]
```
chevdor, line 250:
```suggestion
                      ]

```

parachains/integration-tests/statemint/xcm/2_ump.yml
```diff
@@ -0,0 +1,193 @@
...
+                    { v1: [ *ap_ksm_fungible ] }, # assets
+                    0, # feeAssetItem
+                    { unlimited: true } # weightLimit
+                  ]
+                  events:
+                    - name: polkadotXcm.Attempted
+                      attribute:
+                        type: XcmV2TraitsOutcome
+                        isError: true
+                        value: Barrier
```
chevdor, line 193:
```suggestion
                        value: Barrier
                                 
```

parachains/integration-tests/statemint/xcm/3_hrmp-open-channels.yml
```diff
@@ -0,0 +1,400 @@
...
+                        value: Ok
+                    - name: polkadotXcm.Sent
+                    - name: ump.ExecutedUpward
+                      chain: *relay_chain
+                      attributes:
+                        type: XcmV2TraitsOutcome
+                        isComplete: true
+                        value: 4,000,000
+                    - name: hrmp.OpenChannelAccepted
+                      chain: *relay_chain
```
chevdor, line 400:
```suggestion
                      chain: *relay_chain

```

chevdor (approved):
LGTM beside the spacing issues

.gitignore
NachoPal: I dind't add the `.env`, only `bin`

chevdor: The question still stands :)
If the `.env` is not mandatory, we can skip this change.

gilescope:
Might be nice to mention in the readme how to run the tests.

NachoPal:
> Might be nice to mention in the readme how to run the tests.

Ok, I still have to create a script to run them easily. I'll add a mention in the README then.

bkchr:
Why isn't that done with zombienet? We don't want to have X different testing tools that no one maintains. If there is support missing for stuff in zombienet, then we need to add support. XCM aka sending transactions is a really "standard" feature that zombienet should support. So, please port these tests to normal zombienet tests. (I know that zombienet can just run normal polkadot-js scripts, if that isn't enough we need to make it easier to send xcm messages)

CC @pepoviola

NachoPal:
The tool was created to fulfil Statemine/t testing needs when zombienet was still in its very early stages and was something we couldn't use at all. Since zombienet became stable, we integrated it to the tool and now it can be used alongside to deploy the network.

Regarding your comment about "no one maintains", I do the same way Javier mantains zombienet, and the rest of the Common Good Parachains team can do it also + the community. I strongly believe in the virtues of the tool and how it eases writing tests quickly. The tool has its own npm package, so it can also be considered to be integrated directly in zombienet.

The tests have been in its own repository until now, and we have been using them already for moths, but now we need to add them here (it is just a bunch of yaml files) in our efforts to improve our release process. We need a CI job to run the tests.

PR #1433, opened 2022-07-07
Collectives: teleport slashed assets 

muharem:
Teleport slashed assets to the Polkadot treasury account.

### DIFF
--- a/parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
+++ b/parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+pub const POLKADOT_TREASURY_PALLET_INDEX: u8 = 19;
+
 pub mod currency {
 	use polkadot_core_primitives::Balance;
 	use polkadot_runtime_constants as constants;
--- a/parachains/runtimes/collectives/collectives-polkadot/src/impls.rs
+++ b/parachains/runtimes/collectives/collectives-polkadot/src/impls.rs
@@ -28,6 +28,7 @@ type NegativeImbalanceOf<T, I> = <<T as pallet_alliance::Config<I>>::Currency as
 type BalanceOf<T, I> = <<T as pallet_alliance::Config<I>>::Currency as Currency<
 	<T as frame_system::Config>::AccountId,
 >>::Balance;
+
 pub struct ToParentTreasury<T, I = ()>(PhantomData<(T, I)>);
 impl<T, I: 'static> OnUnbalanced<NegativeImbalanceOf<T, I>> for ToParentTreasury<T, I>
 where
--- a/parachains/runtimes/collectives/collectives-polkadot/src/impls.rs
+++ b/parachains/runtimes/collectives/collectives-polkadot/src/impls.rs
@@ -1,0 +1,50 @@
+// Copyright (C) 2021 Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::constants::POLKADOT_TREASURY_PALLET_INDEX;
+use frame_support::{
+	log,
+	traits::{Currency, Imbalance, OnUnbalanced, OriginTrait},
+};
+use sp_std::{boxed::Box, marker::PhantomData};
+use xcm::latest::{Fungibility, Junction, Parent};
+
+type NegativeImbalanceOf<T, I> = <<T as pallet_alliance::Config<I>>::Currency as Currency<
+	<T as frame_system::Config>::AccountId,
+>>::NegativeImbalance;
+
+type BalanceOf<T, I> = <<T as pallet_alliance::Config<I>>::Currency as Currency<
+	<T as frame_system::Config>::AccountId,
+>>::Balance;
+pub struct ToParentTreasury<T, I = ()>(PhantomData<(T, I)>);
+impl<T, I: 'static> OnUnbalanced<NegativeImbalanceOf<T, I>> for ToParentTreasury<T, I>
+where
+	T: pallet_xcm::Config + frame_system::Config + pallet_alliance::Config<I>,
+	BalanceOf<T, I>: Into<Fungibility>,
+{
+	fn on_unbalanced(amount: NegativeImbalanceOf<T, I>) {
+		let result = pallet_xcm::Pallet::<T>::teleport_assets(
+			<T as frame_system::Config>::Origin::root(),
+			Box::new(Parent.into()),
+			Box::new(Junction::PalletInstance(POLKADOT_TREASURY_PALLET_INDEX).into().into()),
+			Box::new((Parent, amount.peek()).into()),
+			0,
+		);
+		match result {
+			Err(err) => log::warn!("Failed to teleport slashed assets: {:?}", err),
+			_ => (),
+		};
+	}
+}
--- a/parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
+++ b/parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
@@ -38,10 +38,12 @@
 include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
 pub mod constants;
+pub mod impls;
 mod weights;
 pub mod xcm_config;
 
 use cumulus_pallet_parachain_system::RelayNumberStrictlyIncreases;
+use impls::ToParentTreasury;
 use sp_api::impl_runtime_apis;
 use sp_core::{crypto::KeyTypeId, OpaqueMetadata};
 use sp_runtime::{
@@ -500,7 +502,7 @@ impl pallet_alliance::Config for Runtime {
 		pallet_collective::EnsureProportionMoreThan<AccountId, AllianceCollective, 2, 3>,
 	>;
 	type Currency = Balances;
-	type Slashed = (); // TODO:COLLECTIVES add handler to send teleport to Relay Treasury
+	type Slashed = ToParentTreasury<Runtime>;
 	type InitializeMembers = AllianceMotion;
 	type MembershipChanged = AllianceMotion;
 	type IdentityVerifier = (); // Don't block accounts on identity criteria
--- a/parachains/common/src/impls.rs
+++ b/parachains/common/src/impls.rs
@@ -49,23 +49,6 @@ where
 	}
 }
 
-/*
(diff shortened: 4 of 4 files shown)


xlc:
This sounds like a bad idea that makes DoS relaychain much easier. Why not just sent to local treasury and have another method to teleport local treasury to parent treasury.

parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
```diff
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+pub const POLKADOT_TREASURY_PALLET_INDEX: u8 = 19;
```
gilescope, line 16: LGTM: https://github.com/paritytech/polkadot/blob/5a09bf62ec7be0e90dfae335a80459728a896d95/runtime/polkadot/src/lib.rs#L1429

gilescope (approved):
Nice work!

parachains/runtimes/collectives/collectives-polkadot/src/impls.rs
```diff
@@ -0,0 +1,51 @@
...
+
+pub struct ToParentTreasury<T, I = ()>(PhantomData<(T, I)>);
+impl<T, I: 'static> OnUnbalanced<NegativeImbalanceOf<T, I>> for ToParentTreasury<T, I>
+where
+	T: pallet_xcm::Config + frame_system::Config + pallet_alliance::Config<I>,
+	BalanceOf<T, I>: Into<Fungibility>,
+{
+	fn on_unbalanced(amount: NegativeImbalanceOf<T, I>) {
+		let result = pallet_xcm::Pallet::<T>::teleport_assets(
+			<T as frame_system::Config>::Origin::root(),
```
joepetrowski, line 40: `teleport_assets` will try to withdraw them from some account. `Root` will pass any kind of `ensure_origin` checks, but I don't think it will have the `amount` in an account, since it's still an imbalance. So I _think_ it will fail on `WithdrawAsset`. @KiChjang ?

We could put something like this in:
```rust
fn on_unbalanced(amount: NegativeImbalanceOf<T, I>) {
    <pallet_balances::Pallet<R>>::resolve_creating(some_new_temp_account, &amount);
    let result = pallet_xcm::Pallet::<T>::teleport_assets(
        some_new_temp_account,
        ...
    );
    ...
}
```

parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
joepetrowski: This index can change in a runtime upgrade (although we hope it won't). The `AccountId` is generated from the [module ID](https://github.com/paritytech/polkadot/blob/5a09bf62ec7be0e90dfae335a80459728a896d95/runtime/polkadot/src/lib.rs#L809), not the pallet index. So I wonder if it's better to derive the same `AccountId` from that?

joepetrowski:
> This sounds like a bad idea that makes DoS relaychain much easier.

How? The DoS protection is the fact that someone is getting slashed. And the slasher in this case is the Polkadot Alliance, which is semi-trusted. If the Alliance is DoSing the network then we have bigger problems.

> Why not just sent to local treasury and have another method to teleport local treasury to parent treasury.

1. Because there is no local treasury.
2. The Alliance members should not be in control of the tokens.
3. There should be a clear story about what happens when someone is slashed and where the tokens go. Putting them in limbo just adds confusion about the consequences of getting slashed and more work for others to sort it out later.

parachains/runtimes/collectives/collectives-polkadot/src/impls.rs
KiChjang: We first have to determine exactly which account it's withdrawing from, so let's do some investigation and look at the code in `pallet_xcm::do_teleport_assets` (because `teleport_assets` ultimately calls it):

https://github.com/paritytech/polkadot/blob/3b2b067d949c02ece20c76f1590066f2cd87b2b8/xcm/pallet-xcm/src/lib.rs#L825-L881

And there we see that there's a conversion from an `Origin` to a `MultiLocation` type in `T::ExecuteOrigin::ensure_origin()`, so now we'll need to see just how the collectives parachain defines `ExecuteXcmOrigin`:

https://github.com/paritytech/cumulus/blob/26301ad78a6dc2ac036517cfbd686bc27c139dc3/parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs#L172

So now we look at the code for `EnsureXcmOrigin`:

https://github.com/paritytech/polkadot/blob/3b2b067d949c02ece20c76f1590066f2cd87b2b8/xcm/xcm-builder/src/origin_conversion.rs#L233-L259

Normally we'd have to dig through more to know what the `Conversion` type is, but since we know in advance that the `Origin` is root, we see from the code that it gets converted to `Here`. Neat.

Now, we know what the root origin `MultiLocation` is, but we don't know which account it gets converted to yet, so we then turn our attention to the code for executing `WithdrawAsset` for clues:

https://github.com/paritytech/polkadot/blob/3b2b067d949c02ece20c76f1590066f2cd87b2b8/xcm/xcm-executor/src/lib.rs#L274-L282

Now we can clearly see that the actual asset withdrawal code is done in `Config::AssetTransactor::withdraw_asset`, so now we look at what the collectives parachain has configured for its `AssetTransactor` and look at its `withdraw_asset` implementation:

https://github.com/paritytech/cumulus/blob/26301ad78a6dc2ac036517cfbd686bc27c139dc3/parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs#L136

Aha, so it's a `CurrencyTransactor`, and it's defined here:

https://github.com/paritytech/cumulus/blob/26301ad78a6dc2ac036517cfbd686bc27c139dc3/parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs#L63-L75

And finally we see the code for `withdraw_asset` defined in `CurrencyAdapter`:

https://github.com/paritytech/polkadot/blob/3b2b067d949c02ece20c76f1590066f2cd87b2b8/xcm/xcm-builder/src/currency_adapter.rs#L155-L164

The line that we're most interested in now is the following:

```rust
		let who =
			AccountIdConverter::convert_ref(who).map_err(|()| Error::AccountIdConversionFailed)?;
```

... because it converts a `MultiLocation` into an account that we can actually withdraw balances from, in which case, the next question is  "what is `AccountIdConverter`?" Looking at the definition for `CurrencyAdapter`, we see that it's the 3rd type parameter, and in the collectives parachain runtime, it's defined as `LocationToAccountId`, which is the following:

https://github.com/paritytech/cumulus/blob/26301ad78a6dc2ac036517cfbd686bc27c139dc3/parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs#L51-L61

If we go to [`xcm-builder/src/location_conversion.rs`](https://github.com/paritytech/polkadot/blob/master/xcm/xcm-builder/src/location_conversion.rs), we'll see that none of the listed types here can help in converting a `Here` to a proper account ID, therefore, we can safely conclude that yes, `WithdrawAsset` will fail if the origin is root.

TL;DR: Joe is right in claiming that this code will fail during the execution of `WithdrawAsset`.

xlc:
I guess if slash operation is protected it will be fine. Still, having fun trying to figure out the correct weight for extrinsic that triggers slash.

By local treasury I mean just a local pallet account that is only accessible by root. Then relaychain governance can use XCM to teleport those funds back if needed. There is no need to have a treasury pallet.

The slash operation and amount is very clearly defined so I guess current approach is fine. However image if somehow resulting the slash amount to be less than ED, the cost of sending XCM is just not worth the teleport. This approach is unsafe as a generic way to handle slash. My only worry is in future, some new introduction of the slash mechanism could result unexpected complications.

joepetrowski:
> The slash operation and amount is very clearly defined so I guess current approach is fine. However image if somehow resulting the slash amount to be less than ED, the cost of sending XCM is just not worth the teleport. This approach is unsafe as a generic way to handle slash. My only worry is in future, some new introduction of the slash mechanism could result unexpected complications.

I don't think this is a huge concern. If slashes are less than the ED then they are not really that strong of a disincentive.

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -482,6 +484,9 @@ pub const MAX_ALLIES: u32 = 100;
 
 parameter_types! {
 	pub const AllyDeposit: Balance = 1_000 * UNITS; // 1,000 DOT bond to join as an Ally
+	// account used to temporary deposit slashed imbalance before teleporting
+	pub SlachedImbalanceAccId: AccountId = constants::account::SLASHED_IMBALANCE_ACC_ID.into();
```
joepetrowski, line 488:
```suggestion
	// account used to temporarily deposit slashed imbalance before teleporting
	pub SlashedImbalanceAccId: AccountId = constants::account::SLASHED_IMBALANCE_ACC_ID.into();
```

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -500,7 +505,7 @@ impl pallet_alliance::Config for Runtime {
 		pallet_collective::EnsureProportionMoreThan<AccountId, AllianceCollective, 2, 3>,
 	>;
 	type Currency = Balances;
-	type Slashed = (); // TODO:COLLECTIVES add handler to send teleport to Relay Treasury
+	type Slashed = ToParentTreasury<RelayTreasuryAccId, SlachedImbalanceAccId, Runtime>;
```
joepetrowski, line 508:
```suggestion
	type Slashed = ToParentTreasury<RelayTreasuryAccId, SlashedImbalanceAccId, Runtime>;
```

parachains/runtimes/collectives/collectives-polkadot/src/lib.rs
muharem: need to pay more attention to typos 🤪

parachains/runtimes/collectives/collectives-polkadot/src/impls.rs
muharem: fixed

muharem:
Integration tests for the use case - https://github.com/paritytech/parachains-integration-tests/pull/26
PR to allow teleport to the westend relay chain - https://github.com/paritytech/parachains-integration-tests/pull/26

parachains/runtimes/collectives/collectives-polkadot/src/impls.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (C) 2021 Parity Technologies (UK) Ltd.
 // SPDX-License-Identifier: Apache-2.0
 
-// Licensed under the Apache License, Version 2.0 (the "License");
+// &Licensed under the Apache License, Version 2.0 (the "License");
```
gilescope, line 4:
```suggestion
// Licensed under the Apache License, Version 2.0 (the "License");
```

parachains/runtimes/collectives/collectives-polkadot/src/constants.rs
```diff
@@ -13,6 +13,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+pub mod account {
+	use frame_support::PalletId;
+	use sp_runtime::AccountId32;
+
+	/// realy chain treasury pallet id, used to convert into AccountId
```
joepetrowski, line 20: Don't know how I missed this one.
```suggestion
	/// Relay Chain treasury pallet id, used to convert into AccountId
```

joepetrowski (approved):

PR #1438, opened 2022-07-11
Weights updated for alliance-bench PR for alliance PR

gilescope:
Reran the weights using the standard CI infrastructure including the new pallets.


joepetrowski (approved):

muharem:
@gilescope can you give more context on the PR description. 
I see some of weights increased a lot, some of them doubled.

gilescope:
Yeah some of the weights look bigger than expected. https://github.com/paritytech/cumulus/blob/3bb82e5126362f9cf3952c60b737c3b71b0a136b/parachains/runtimes/assets/statemint/src/weights/cumulus_pallet_xcmp_queue.rs#L50 is 3 million where as we're quoting 6 million here. But having bigger weights is just more secure (more expensive too). I suspect they will settle down on future re-runs, but these values are safe. (We will undoubtedly rerun the weights again before the production release so we can keep an eye on the changes.)

muharem (approved):

Issue #989, opened 2022-02-15
Enable collation with external relay chain full node


skunert:
As discussed in https://github.com/paritytech/cumulus/issues/545 we want to support running the collator relay chain node as its own process.
After https://github.com/paritytech/cumulus/pull/963 lands, we are in a position to run a parachain full node without running a relay chain full node internally. To run a proper parachain collator, we miss some of the subsystems usually running inside the full node. They do not offer an RPC interface, so I plan to start the relevant subsystems inside the collator.

## Relevant Subsystems
![CleanShot 2022-02-15 at 12 18 53](https://user-images.githubusercontent.com/5718005/154051817-9fb792b0-52e4-49a1-a52d-f085ec434a67.png)
After checking the code, I found that the following subsystems are involved:
- Availability Recovery
- Collation Generation
- Collator Protocol
- Network Bridge
- Runtime API
- Availability Store
- Chain API

The Availability store should be optional for now since the data we request from it will not be available inside the collator anyway and will then be requested from the Network Bridge.

## Implementation Plan
- Implement a custom `OverseerGen` that will replace unused Subsystem with dummies that drain channels
- Feed a client to the Runtime API subsystem that will use the RPC Client introduced in https://github.com/paritytech/cumulus/pull/963 to perform runtime calls

Related discussion in https://github.com/paritytech/polkadot/issues/4763

bkchr:
CC @drahnr maybe you have some input on the `OverseerGen` stuff :)

drahnr:
I think implementing `OverseerGen` for implementation of that is really all you need to do, besides assuring to `overseer_enable_ anyways`. Make sure to consume messages in the dummy subsystems or assert none are sent there. Ping me as needed.

gregzaitsev:
Hi, Sebastian! Is there any way to contribute to this issue? Looks like collators are almost finished, is that right?

skunert:
Hey @gregzaitsev, I have a branch where this works, some refactorings and polish needs to be done still.
So currently there is not much to contribute here, the feature will land soon.

drahnr:
@skunert I assume you are aware, a few weeks ago a change landed that splits the network bridge into two, incoming and outgoing messages, so the initial diagram might need some adjustments :)

PR #1462, opened 2022-07-21
Introduce RPC Worker

skunert:
This PR is the next step towards https://github.com/paritytech/cumulus/issues/989.

Currently the collator opens a new RPC subscription every time we request a notification stream from the `RelayChainRPCInterface`. 

After this PR is merged, we will have only three open notification subscriptionsvia RPC:
- finalized block headers
- imported block headers
- new best block headers

A worker is introduced that will poll these three subscriptions regularly and distribute the contained headers to all listening streams. Requesting a stream from the `RelayChainRPCInterface` will return a new channel and inform the worker.
Since the streams are only used in the collator, and we have complete control, we use bounded channels with a limit of 20 items.

### DIFF
--- a/client/relay-chain-rpc-interface/src/lib.rs
+++ b/client/relay-chain-rpc-interface/src/lib.rs
@@ -35,7 +35,7 @@ use std::pin::Pin;
 pub use url::Url;
 
 mod rpc_client;
-pub use rpc_client::{create_worker_client, RelayChainRPCClient};
+pub use rpc_client::{create_worker_client, RPCStreamWorker, RelayChainRPCClient};
 
 const TIMEOUT_IN_SECONDS: u64 = 6;
 
--- a/client/relay-chain-rpc-interface/src/lib.rs
+++ b/client/relay-chain-rpc-interface/src/lib.rs
@@ -35,25 +35,25 @@ use std::pin::Pin;
 pub use url::Url;
 
 mod rpc_client;
-pub use rpc_client::{create_worker_client, RPCStreamWorker, RelayChainRPCClient};
+pub use rpc_client::{create_worker_client, RelayChainRpcClient, RpcStreamWorker};
 
 const TIMEOUT_IN_SECONDS: u64 = 6;
 
-/// RelayChainRPCInterface is used to interact with a full node that is running locally
+/// RelayChainRpcInterface is used to interact with a full node that is running locally
 /// in the same process.
 #[derive(Clone)]
-pub struct RelayChainRPCInterface {
-	rpc_client: RelayChainRPCClient,
+pub struct RelayChainRpcInterface {
+	rpc_client: RelayChainRpcClient,
 }
 
-impl RelayChainRPCInterface {
-	pub fn new(rpc_client: RelayChainRPCClient) -> Self {
+impl RelayChainRpcInterface {
+	pub fn new(rpc_client: RelayChainRpcClient) -> Self {
 		Self { rpc_client }
 	}
 }
 
 #[async_trait]
-impl RelayChainInterface for RelayChainRPCInterface {
+impl RelayChainInterface for RelayChainRpcInterface {
 	async fn retrieve_dmq_contents(
 		&self,
 		para_id: ParaId,
--- a/client/relay-chain-rpc-interface/src/lib.rs
+++ b/client/relay-chain-rpc-interface/src/lib.rs
@@ -46,238 +46,11 @@ use sp_storage::StorageKey;
 use std::{pin::Pin, sync::Arc};
 
 pub use url::Url;
+mod rpc_client;
+pub use rpc_client::RelayChainRPCClient;
 
 const LOG_TARGET: &str = "relay-chain-rpc-interface";
 const TIMEOUT_IN_SECONDS: u64 = 6;
-
-/// Client that maps RPC methods and deserializes results
-#[derive(Clone)]
-struct RelayChainRPCClient {
-	/// Websocket client to make calls
-	ws_client: Arc<JsonRPCClient>,
-
-	/// Retry strategy that should be used for requests and subscriptions
-	retry_strategy: ExponentialBackoff,
-}
-
-impl RelayChainRPCClient {
-	pub async fn new(url: Url) -> RelayChainResult<Self> {
-		tracing::info!(target: LOG_TARGET, url = %url.to_string(), "Initializing RPC Client");
-		let ws_client = WsClientBuilder::default().build(url.as_str()).await?;
-
-		Ok(RelayChainRPCClient {
-			ws_client: Arc::new(ws_client),
-			retry_strategy: ExponentialBackoff::default(),
-		})
-	}
-
-	/// Call a call to `state_call` rpc method.
-	async fn call_remote_runtime_function<R: Decode>(
-		&self,
-		method_name: &str,
-		hash: PHash,
-		payload: Option<impl Encode>,
-	) -> RelayChainResult<R> {
-		let payload_bytes =
-			payload.map_or(sp_core::Bytes(Vec::new()), |v| sp_core::Bytes(v.encode()));
-		let params = rpc_params! {
-			method_name,
-			payload_bytes,
-			hash
-		};
-		let res = self
-			.request_tracing::<sp_core::Bytes, _>("state_call", params, |err| {
-				tracing::trace!(
-					target: LOG_TARGET,
-					%method_name,
-					%hash,
-					error = %err,
-					"Error during call to 'state_call'.",
-				);
-			})
-			.await?;
-		Decode::decode(&mut &*res.0).map_err(Into::into)
-	}
-
-	/// Subscribe to a notification stream via RPC
-	async fn subscribe<'a, R>(
-		&self,
-		sub_name: &'a str,
-		unsub_name: &'a str,
-		params: Option<ParamsSer<'a>>,
-	) -> RelayChainResult<Subscription<R>>
-	where
-		R: DeserializeOwned,
-	{
-		self.ws_client
-			.subscribe::<R>(sub_name, params, unsub_name)
-			.await
-			.map_err(|err| RelayChainError::RPCCallError(sub_name.to_string(), err))
-	}
-
-	/// Perform RPC request
-	async fn request<'a, R>(
-		&self,
-		method: &'a str,
-		params: Option<ParamsSer<'a>>,
-	) -> Result<R, RelayChainError>
-	where
-		R: DeserializeOwned + std::fmt::Debug,
-	{
-		self.request_tracing(
-			method,
-			params,
-			|e| tracing::trace!(target:LOG_TARGET, error = %e, %method, "Unable to complete RPC request"),
-		)
-		.await
-	}
-
-	/// Perform RPC request
-	async fn request_tracing<'a, R, OR>(
-		&self,
-		method: &'a str,
-		params: Option<ParamsSer<'a>>,
-		trace_error: OR,
-	) -> Result<R, RelayChainError>
-	where
-		R: DeserializeOwned + std::fmt::Debug,
-		OR: Fn(&jsonrpsee::core::Error),
-	{
-		retry_notify(
-			self.retry_strategy.clone(),
-			|| async {
-				self.ws_client.request(method, params.clone()).await.map_err(|err| match err {
-					JsonRpseeError::Transport(_) =>
-						backoff::Error::Transient { err, retry_after: None },
-					_ => backoff::Error::Permanent(err),
-				})
-			},
-			|error, dur| tracing::trace!(target: LOG_TARGET, %error, ?dur, "Encountered transport error, retrying."),
-		)
-		.await
-		.map_err(|err| {
-			trace_error(&err);
-			RelayChainError::RPCCallError(method.to_string(), err)})
-	}
-
-	async fn system_health(&self) -> Result<Health, RelayChainError> {
-		self.request("system_health", None).await
-	}
-
-	async fn state_get_read_proof(
-		&self,
-		storage_keys: Vec<StorageKey>,
-		at: Option<PHash>,
-	) -> Result<ReadProof<PHash>, RelayChainError> {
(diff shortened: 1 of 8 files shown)


davxy:
Really silly nit. Maybe we can stick to the notation where also the acronyms are usually written in camel case.
In other words `RPC` -> `Rpc`. Thus `RelayChainRPCClient` -> `RelayChainRpcClient`

Feel free to ignore me :-D

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -0,0 +1,454 @@
...
+	best_header_listeners: Vec<Sender<PHeader>>,
+
+	// Incoming notification subscriptions
+	rpc_imported_header_subscription: Subscription<PHeader>,
+	rpc_finalized_header_subscription: Subscription<PHeader>,
+	rpc_best_header_subscription: Subscription<PHeader>,
+}
+
+/// Entry point to create [`RelayChainRPCClient`] and [`RPCStreamWorker`];
+pub async fn create_worker_client(
```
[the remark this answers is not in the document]
skunert: Yes, it is expected that users of this method spawn the worker task immediately. Otherwise the streams will not work and all usages inside the collator need the streams to work. I will look into spawning directly.

As for the users of this, there will be other usage soon. Currently we immediately wrap the client in a `RelayChainRPCInterface`. In the future we will pass the client into the relay chain subsystems additionally.

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -0,0 +1,454 @@
...
+		retry_notify(
+			self.retry_strategy.clone(),
+			|| async {
+				self.ws_client.request(method, params.clone()).await.map_err(|err| match err {
+					JsonRpseeError::Transport(_) =>
+						backoff::Error::Transient { err, retry_after: None },
+					_ => backoff::Error::Permanent(err),
+				})
+			},
+			|error, dur| tracing::trace!(target: LOG_TARGET, %error, ?dur, "Encountered transport error, retrying."),
```
koute, line 287: Do we expect that this will trigger often? It should only trigger in cases like e.g. temporary disconnection, right? If so it seems to me that this is more of a `debug` log rather than a `trace` log. (But it's up to you; feel free to ignore this comment.)

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -0,0 +1,454 @@
...
+fn handle_event_distribution(
+	event: Option<Result<PHeader, JsonRpseeError>>,
+	senders: &mut Vec<Sender<PHeader>>,
+) -> Result<(), String> {
+	match event {
+		// If sending fails, we remove the sender from the list because the receiver was dropped
+		Some(Ok(header)) => {
+			senders.retain_mut(|e| e.try_send(header.clone()).is_ok());
+			Ok(())
+		},
```
koute, line 114: The `try_send` can also return an error when the sender's channel is full, not only when it was dropped. (Since it's a bounded channel.)

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -0,0 +1,454 @@
...
+pub enum NotificationRegisterMessage {
+	RegisterBestHeadListener(Sender<PHeader>),
+	RegisterImportListener(Sender<PHeader>),
+	RegisterFinalizationListener(Sender<PHeader>),
+}
+
+/// Worker that should be used in combination with [`RelayChainRPCClient`]. Must be polled to distribute header notifications to listeners.
+pub struct RPCStreamWorker {
+	// Communication channel with the RPC client
+	client_receiver: TracingUnboundedReceiver<NotificationRegisterMessage>,
```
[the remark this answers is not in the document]
skunert: I initially wanted to make it bounded, but the bounded sending requires `&mut self` which makes it super inconvenient in this case. Every stream getting method in the collator would need to take the mutable reference.

The channel is only used by the clients to request new notification streams, which are bounded. Also, the worker is started automatically and drains this channel when the client is constructed. The overall risk seems manageable to me in this case.

client/relay-chain-rpc-interface/src/rpc_client.rs
skunert: I just moved this client without modification for this PR. Since we consider all of the RPC collation stuff experimental and recommend running relay and para node on the same machine, I do not see this being triggered. 

If there were any severe networking problems, the current RPC collator node would crash anyway, as we have no way to deal with broken subscriptions. So I consider the whole error handling an area to be overhauled down the road.

client/relay-chain-rpc-interface/src/rpc_client.rs
skunert: I added a error print and do not remove the sender for now. Alternative would be to make the task fail.

client/relay-chain-rpc-interface/src/lib.rs
```diff
@@ -26,274 +25,35 @@ use cumulus_primitives_core::{
...
-
-	async fn parachain_host_session_index_for_child(
-		&self,
-		at: PHash,
-	) -> Result<SessionIndex, RelayChainError> {
-		self.call_remote_runtime_function("ParachainHost_session_index_for_child", at, None::<()>)
-			.await
-	}
+mod rpc_client;
+pub use rpc_client::{create_client_and_start_worker, RelayChainRpcClient, RpcStreamWorker};
```
davxy, line 38:
```suggestion
pub use rpc_client::{create_client_and_start_worker, RelayChainRpcClient};
```

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -0,0 +1,471 @@
...
+/// Worker messages to register new notification listeners
+#[derive(Clone, Debug)]
+pub enum NotificationRegisterMessage {
+	RegisterBestHeadListener(Sender<PHeader>),
+	RegisterImportListener(Sender<PHeader>),
+	RegisterFinalizationListener(Sender<PHeader>),
+}
+
+/// Worker that should be used in combination with [`RelayChainRpcClient`]. Must be polled to distribute header notifications to listeners.
+pub struct RpcStreamWorker {
```
davxy, line 76:
```suggestion
struct RpcStreamWorker {
```

Because it is not directly exposed anymore

davxy (commented):
LGTM, I've just added a suggestion to set the `RpcStreamWorker` as private

davxy (approved):

client/relay-chain-rpc-interface/src/rpc_client.rs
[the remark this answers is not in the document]
koute: > I initially wanted to make it bounded, but the bounded sending requires `&mut self` which makes it super inconvenient in this case. Every stream getting method in the collator would need to take the mutable reference.

FYI, you could always just use [the channel from `tokio`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html) which doesn't require a `&mut self` to send. (:

client/relay-chain-rpc-interface/src/rpc_client.rs
koute: Yeah, I think we need to think about what we should do if that happens and handle it appropriately, but considering this is still experimental and a work in progress I'm fine with it being as-is. But we'll need to revisit this issue before actually pushing this into production. Could you at least put like a TODO here as a reminder?

client/relay-chain-rpc-interface/src/rpc_client.rs
koute: Fair enough!

koute (approved):
The unbounded channel's still making me a little nervous, but I'll leave the decision as to what to do with it up to you. Otherwise LGTM.

client/relay-chain-rpc-interface/src/rpc_client.rs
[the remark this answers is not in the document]
skunert: Very good suggestion, was not aware of this difference. Switching to tokio then to be safe.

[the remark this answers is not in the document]
bkchr: Sorry for being late to the party. Not sure @koute why the usage of `unbounded` for the service to worker communication makes you nervous. I mean, that some is some well defined communication. We also don't want to drop any of this communication. We also don't need any backpressure here, otherwise @skunert could also just make the methods take `&mut self`. 

TLDR: I don't see any reason why we should not use unbounded here. If the worker is dropped the channel is closed and if we don't poll the unbounded receiver anymore, there is something broken anyway.

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -0,0 +1,478 @@
...
+
+	fn send_register_message_to_worker(
+		&self,
+		message: NotificationRegisterMessage,
+	) -> Result<(), RelayChainError> {
+		match &self.to_worker_channel {
+			Some(channel) => {
+				channel
+					.try_send(message)
+					.map_err(|e| RelayChainError::WorkerCommunicationError(e.to_string()))?;
```
bkchr, line 426: We should not really fail because of service to worker communication channel is full.

client/relay-chain-rpc-interface/src/rpc_client.rs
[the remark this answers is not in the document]
skunert: I think we should just default to bounded.

I just picked an unbounded one because the API was better for my usecase. The tokio channel solved this.

client/relay-chain-rpc-interface/src/rpc_client.rs
skunert: I want it to fail for now, since the worker should really be fast enough to drain this channel. It is only used to request another notification stream.

If it runs full, I need to fix it. Once this is used in production we can make it more defensive. As mentioned, there will be another error handling pass anyway.

client/relay-chain-rpc-interface/src/rpc_client.rs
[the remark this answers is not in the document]
bkchr: Bounded max sense when you need back pressure. Here you don't need any back pressure.

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -0,0 +1,478 @@
...
+	task_manager: &mut TaskManager,
+) -> RelayChainResult<RelayChainRpcClient> {
+	let mut client = RelayChainRpcClient::new(url).await?;
+	let best_head_stream = client.subscribe_new_best_heads().await?;
+	let finalized_head_stream = client.subscribe_finalized_heads().await?;
+	let imported_head_stream = client.subscribe_imported_heads().await?;
+
+	let (worker, sender) =
+		RpcStreamWorker::new(imported_head_stream, best_head_stream, finalized_head_stream);
+	client.set_worker_channel(sender);
```
bkchr, line 105: Why don't we first create the worker and then the client? Then we don't need to make the `sender` an `Option`.

client/relay-chain-rpc-interface/src/rpc_client.rs
bkchr: I get that you want to improve this later. However, the entire system here with worker/client is build in a way that both are working together in a tight way. Meaning, if the worker or the client stops the other end should also stop. Then both are also not required to have any kind of back pressure. You want that anything you send from the client side gets forwarded to the worker. Dropping here any data doesn't make any sense.

client/relay-chain-rpc-interface/src/rpc_client.rs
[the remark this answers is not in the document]
koute: Sorry, it's just that I've seen enough code already that used an unbounded channels and was also well defined and looked innocuous but ended up causing problems. And a failure mode where it fails explicitly is easier to debug *if* something were to be broken here. Which is why if we'd end up in the "something's broken anyway" situation I'd prefer to have it alert us to that fact, instead of potentially just silently chugging along in the worst case.

This is just a "just in case" measure; as you say, yeah, we don't strictly *need* this here, but I also don't see why not to do this; in normal circumstances when everything works as it should (that is - the messages are handled fast enough that no backpressure will ever be triggered) it should essentially be functionally equivalent to an unbound channel.

That said, it's not a huge deal, so if you feel strongly about not having a bounded channel here I'm fine with reverting it. :P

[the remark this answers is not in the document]
bkchr: Okay okay, I "give" up 😂 but we should at least increase the channel size to 1000. If something is broken, we still will see it.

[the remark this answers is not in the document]
bkchr: I mean it probably makes no difference, but we use almost everywhere in our code the futures channel and futures select. That is something I don't want to die on, but yeah.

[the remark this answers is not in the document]
koute: Indeed. But I don't really see us ever dropping the `tokio` dependency, and if it's an integral part of `tokio` we might as well use it *if* needed. It's not like it's some random unmaintained crate fresh off crates.io.

I even once tried to compare whether switching from the futures' channel to tokio's channel would make a difference performance-wise (for a few particularly busy channels in substrate), but in that particular case I found the difference to be negligible. Or in other words, performance-wise they're comparable, so it shouldn't really matter which one we use *in general* (not that this aspect matters here; it doesn't), except `tokio`'s channels *do* have a richer API in those cases where we can make use of it.

Anyway, again, this is not a huge deal so from my point of view so I'm fine with whichever way @skunert wants to go.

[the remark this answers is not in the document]
skunert: Lets stick to the tokio bounded channel in this case. Its more defensive here with no real downsides.

client/relay-chain-rpc-interface/src/rpc_client.rs
skunert: Worker took in the streams which required the client. I changed the methods to be static, so now we got rid of the `Option` 👍

PR #1470, opened 2022-07-26
make runtime release notes more relevant.

gilescope:
Going to create a draft runtime release and check that it's improved.


scripts/ci/changelog/templates/template.md.tera
```diff
@@ -19,17 +19,16 @@ This release contains the changes from `{{ env.REF1 }}` to `{{ env.REF2 }}`.
...
+    {% include "host_functions.md.tera" -%}
+    {% include "global_priority.md.tera" -%}
+    {% include "compiler.md.tera" -%}
+    {% include "migrations-db.md.tera" -%}
+    <!-- skipping runtime data for RELEASE_TYPE = {{ env.RELEASE_TYPE }} -->
 {% else %}
-{% include "migrations-runtime.md.tera" -%}
-
-{% include "runtimes.md.tera" -%}
+    {% include "compiler.md.tera" -%}
```
gilescope, line 29:
```suggestion
```
Given that runtimes are built inside docker the compiler version here is irrelevant so we should not emit it.

gilescope:
before we merge this we need to check carefully that we've not got too many additional new lines and add `-` where needed to remove surplus lines.

Now that context.json is output in the build artefacts we should be able to test this locally without having to trigger draft-releases. (thanks will for all your pointers)

PR #1473, opened 2022-07-27
slurp collectives digest to make appear in release notes

gilescope:
We need to slurp collectives if we want the chain to appear in the release notes.


gilescope:
Hmm, it's not too happy
```
D, [2022-07-27T10:49:07.616018 #1741] DEBUG -- : Building changelog with runtimes
jq: error: $srtool_polkadot_collectives_parachain is not defined at <top-level>, line 6:
          { name: "polkadot-collectives", data: $srtool_polkadot_collectives_parachain[0] },                                                
jq: 1 compile error
thread 'main' panicked at 'JSON parsing: Error("EOF while parsing a value", line: 1, column: 0)', src/wrapped_context.rs:40:53
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted (core dumped)
```

chevdor:
`srtool_polkadot_collectives_parachain` vs `srtool_polkadot_collectives`

chevdor (approved):

PR #1483, opened 2022-07-29
Adding TxVersion Check to signed params

mikiquantum:
Is there any underlaying reason to not have the TxVersion check as part of the signature payload in this particular test runtime? If not, then I would love to have it in, for consistency sake with most of runtime implementations.


bkchr (approved):

bkchr:
There is no reason, just missed. However, it is also not that important. But I'm okay with merging this.

PR #1500, opened 2022-08-04
Removed dependance on `rococo_parachain_runtime` for unrelated stuff

bkontur:
- removed duplicite `get_public_from_seed` vs `get_from_seed`
- changed import for AccountId/AuraId
- refactored stuff around reading runtime from chain_spec.id in `load_spec`

**Questions:**
- [x]  `ContractsRococoRuntimeExecutor` / `SeedlingRuntimeExecutor` / `RococoParachainRuntimeExecutor` are not used anywhere, could be removed?
- [x] there are another `fn get_from_seed` in Cumulus modules (1. `test/service` 2. `parachain-template`) 3. `polkadot-parachain`, is it worth to move it to `parachain-common`  or is it desired to have here multiple impls?


polkadot-parachain/src/command.rs
```diff
@@ -38,12 +38,12 @@ use sc_service::{
...
 use sp_runtime::traits::{AccountIdConversion, Block as BlockT};
-use std::net::SocketAddr;
+use std::{net::SocketAddr, path::PathBuf};
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 enum Runtime {
 	/// This is the default runtime (based on rococo)
-	Generic,
+	Default,
```
gilescope, line 46: we should derive Default and add #[default] here as that would be unsurprising.

polkadot-parachain/src/command.rs
```diff
@@ -97,23 +105,25 @@ fn runtime(id: &str) -> Runtime {
 	} else if id.starts_with("collectives-westend") {
 		Runtime::CollectivesWestend
 	} else {
-		Runtime::Generic
+		Runtime::Default
```
gilescope, line 108: this could then be Runtime::default() maybe? Idk that's totally up to you....

polkadot-parachain/src/command.rs
```diff
@@ -177,34 +190,33 @@ fn load_spec(id: &str) -> std::result::Result<Box<dyn ChainSpec>, String> {
...
-				),
+					Box::new(chain_spec::seedling::SeedlingChainSpec::from_json_file(path)?),
+				Runtime::ContractsRococo =>
+					Box::new(chain_spec::contracts::ContractsRococoChainSpec::from_json_file(path)?),
 				Runtime::Penpal(_para_id) =>
-					Box::new(chain_spec::penpal::PenpalChainSpec::from_json_file(path.into())?),
-				Runtime::Generic => Box::new(chain_spec),
+					Box::new(chain_spec::penpal::PenpalChainSpec::from_json_file(path)?),
+				Runtime::Default =>
+					Box::new(chain_spec::default::DefaultChainSpec::from_json_file(path)?),
```
gilescope, line 219: we can box after - that would be neater if the compiler lets us?

polkadot-parachain/src/chain_spec/penpal.rs
```diff
@@ -14,10 +14,11 @@
...
 // along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
 
-use crate::chain_spec::{get_account_id_from_seed, Extensions};
+use crate::chain_spec::{
+	get_account_id_from_seed, get_collator_keys_from_seed, Extensions, SAFE_XCM_VERSION,
+};
 use cumulus_primitives_core::ParaId;
-// use rococo_parachain_runtime::{AuraId};
-use crate::chain_spec::{get_collator_keys_from_seed, SAFE_XCM_VERSION};
+use parachains_common::{AccountId, AuraId};
```
gilescope, line 21: This is fine. Just watch out because AuraId is different for statemint than all the rest (for historical reasons).

polkadot-parachain/src/command.rs
bkontur: I like it, done

polkadot-parachain/src/chain_spec/penpal.rs
bkontur: yes, thank you, I saw that StatemintAuraId before

polkadot-parachain/src/command.rs
bkontur: yes, agree, would be neater, rust does not allow return multiple type from function, afaik the workaround is that dynamic dispatch with Box

bkontur:
> Looks helpful - thank you.

@gilescope 
thank you, and what about those two open questions/points from description?

polkadot-parachain/src/chain_spec/default.rs
```diff
@@ -0,0 +1,122 @@
...
+// Cumulus is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Default parachain ChainSpec based on testing rococo parachain runtime
+
```
muharem, line 18: I wonder what this means, can we explain more here?
Why defaults based on some specific chain? Should not be other around? 

Is it 'on testing rococo' or on 'test rococo'. If second, probably test can be even omitted.

polkadot-parachain/src/chain_spec/default.rs
```diff
@@ -0,0 +1,122 @@
+// Copyright 2019-2021 Parity Technologies (UK) Ltd.
```
muharem, line 1: 2022?

polkadot-parachain/src/chain_spec/default.rs
```diff
@@ -0,0 +1,122 @@
...
+				],
+				1000.into(),
+			)
+		},
+		Vec::new(),
+		None,
+		None,
+		None,
+		None,
+		Extensions { relay_chain: "westend".into(), para_id: 1000 },
```
muharem, line 63: after the package doc with rococo, westend here confuses even more.

polkadot-parachain/src/chain_spec/default.rs
```diff
@@ -0,0 +1,122 @@
...
+	)
+}
+
+pub fn staging_test_net() -> DefaultChainSpec {
```
muharem, line 67: this fn and the one above, seems like doing similar things, but the names are so different.
would be great to have some docs

polkadot-parachain/src/chain_spec/default.rs
```diff
@@ -0,0 +1,122 @@
...
+use polkadot_service::chain_spec::get_account_id_from_seed;
+use rococo_parachain_runtime::AuraId;
+use sc_chain_spec::ChainType;
+use sp_core::{crypto::UncheckedInto, sr25519};
+
+/// Specialized `ChainSpec` for the normal parachain runtime.
+pub type DefaultChainSpec =
+	sc_service::GenericChainSpec<rococo_parachain_runtime::GenesisConfig, Extensions>;
+
+pub fn get_chain_spec() -> DefaultChainSpec {
```
muharem, line 32: would be great to have some doc.
when you see it first time, its really hard to say what is meant for

polkadot-parachain/src/chain_spec/default.rs
muharem: this probably going to be redefined, how about setting it to some default values, like 
 { relay_chain: "undefined", para_id: 0 },

polkadot-parachain/src/chain_spec/default.rs
```diff
@@ -0,0 +1,122 @@
...
+	DefaultChainSpec::from_genesis(
+		"Local Testnet",
+		"local_testnet",
```
muharem, line 35: Same here, if its supposed to be overwrited, may be "undefined" ?

polkadot-parachain/src/command.rs
```diff
@@ -55,22 +56,30 @@ enum Runtime {
...
 	fn runtime(&self) -> Runtime {
-		runtime(self.id())
+		#[derive(Debug, serde::Deserialize)]
+		struct EmptyChainSpecWithId {
+			id: String,
+		}
+
+		let file = std::fs::File::open(self).expect("Failed to open file");
+		let reader = std::io::BufReader::new(file);
+		let chain_spec: EmptyChainSpecWithId = sp_serializer::from_reader(reader)
```
muharem, line 79: How we obtained the id before, why we got one more file read and parse?

muharem:
> ContractsRococoRuntimeExecutor / SeedlingRuntimeExecutor / RococoParachainRuntimeExecutor are not used anywhere, could be removed?

@bkontur if it compiles without, I would remove (:

> there are another fn get_from_seed in Cumulus modules (1. test/service 2. parachain-template) 3. polkadot-parachain, is it worth to move it to parachain-common or is it desired to have here multiple impls?

Just bare in mind `parachains/common` keeps common items for runtimes.

polkadot-parachain/src/command.rs
bkontur: old/actual approach, reads configuration path to GenericChainSpec based on testing/rococo (Genesis + Extentions), even if you have Statemint/Westmint/Whatever configuration file, e.g. once someone add new configuration which will use different Extentions (or maybe Geneis, not sure), it wont work, because serde will fail (unknown/missing attributes), and all of this is just because we need to know ID from the json file.

I was checking a possibilities to read chain_spec ignoring (genesis/extensions) but I did not find one.
Yess, now it does one more reading file, just ID from json. And then decides, what specific configuration to parse with anohter reading the file, so it wont fail on serde, so now it is more generic approach.

polkadot-parachain/src/chain_spec/default.rs
bkontur: well, to the other comments,
I did not touch any code (it stays like it was before), I just move it to `default.rs`,
main reason why I did this cleanup, was because lots of stuff depended on stuff from `rococo_parachain_runtime` (ChainSpec, AuraId, Genesis...)
```
pub type ChainSpec =
	sc_service::GenericChainSpec<rococo_parachain_runtime::GenesisConfig, Extensions>;
```

but thank you for your comments, now I can clean it little bit more :)

polkadot-parachain/src/command.rs
muharem: Thank you for explaining. 
Make sense.

polkadot-parachain/src/chain_spec/default.rs
bkontur: @muharem 
Thank you for your comments, I was not sure about default.rs, but after your comments I realized and just changed `default.rs` to `rococo_parachain.rs` and now it is align with others and also it mirrors `parachains/runtimes`, and makes much more sense,
so the "Default" word just stays in `command.rs - enum Runtime`

polkadot-parachain/src/chain_spec/default.rs
bkontur: yes, I think it was old copy/paste, but now when it is in rococo_parachain, this is changed to `rococo-local`

polkadot-parachain/src/chain_spec/default.rs
bkontur: as I said before, now everything is align to reflect parachains/runtimes structure, so all this Default/Generic stuff is changed to RococoParachain

polkadot-parachain/src/chain_spec/default.rs
bkontur: I dont now how is this cfg used, but I renamed it accordingly `staging_rococo_parachain_local`

polkadot-parachain/src/chain_spec/default.rs
bkontur: renamed to `"Rococo Parachain Local",`

bkontur:
> get_from_seed

those removed, but I did not touch `get_from_seed` better stays as it is now,
that function is not related to runtime I guess

bkontur:
> > ContractsRococoRuntimeExecutor / SeedlingRuntimeExecutor / RococoParachainRuntimeExecutor are not used anywhere, could be removed?
> 
> @bkontur if it compiles without, I would remove (:
> 
> > there are another fn get_from_seed in Cumulus modules (1. test/service 2. parachain-template) 3. polkadot-parachain, is it worth to move it to parachain-common or is it desired to have here multiple impls?
> 
> Just bare in mind `parachains/common` keeps common items for runtimes.

@muharem 

those removed, but I did not touch `get_from_seed` better stays as it is now,
that function is not related to runtime I guess

paritytech-cicd-pr:
 The job name - test-linux-stable 
 The job logs - https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/1744809

polkadot-parachain/src/chain_spec/default.rs
bkontur: @muharem 
because of this change: westend -> rococo-local, now this test is failing: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/1744809 ,

need to find a way to fix it clearly

bkontur: @muharem @gilescope 
this is working: https://github.com/paritytech/cumulus/pull/1500/commits/54c448f4b0978072c97e02dd2de6352a9eddd695#diff-5dcee2b3279312fb4434bec01df2816aeae5d1e593e2118b74706e1a2c905c83R99-R100

polkadot-parachain/src/chain_spec/rococo_parachain.rs
```diff
@@ -0,0 +1,121 @@
...
+
+// Cumulus is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! ChainSpecs dedicated to Rococo parachain setups (for testing and example purposes)
```
muharem, line 17: Much clear to me

muharem (approved):

PR #1515, opened 2022-08-08
ci: update extrinsic ordering check to use reference binary

coderobe:
same as we recently did on the polkadot side


EgorPopelyaev (approved):

chevdor:
We should keep the existing GH Workflow. I lined up a new PR: https://github.com/paritytech/cumulus/pull/1528

coderobe:
done in https://github.com/paritytech/cumulus/pull/1528

PR #1517, opened 2022-08-08
Align versions for starters runtimes

bkontur:


parachains/runtimes/starters/seedling/src/lib.rs
```diff
@@ -65,7 +65,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 9270,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 7,
```
NachoPal, line 68:
```suggestion
	transaction_version: 1,
```
Do not bump `transaction_version`

parachains/runtimes/starters/shell/src/lib.rs
```diff
@@ -69,10 +69,10 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("shell"),
 	impl_name: create_runtime_str!("shell"),
 	authoring_version: 1,
-	spec_version: 2,
+	spec_version: 9270,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 7,
```
NachoPal, line 75:
```suggestion
	transaction_version: 1,
```
Do not bump `transaction_version`

parachains/runtimes/testing/rococo-parachain/src/lib.rs
```diff
@@ -97,7 +97,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 9270,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 7,
```
NachoPal, line 100:
```suggestion
	transaction_version: 1,
```
Do not bump `transaction_version`

NachoPal (commented):
You forgot to bump `penpal` (only the `spec_version`)

bkchr:
Why bump them? They are thought as some sort of think that people copy as a starter. They don't really need these version numbers?

NachoPal:
> Why bump them? They are thought as some sort of think that people copy as a starter. They don't really need these version numbers?

To indicate they are based on substrate and polkadot `v0.9.27`

bkontur:
@bkchr @gilescope @NachoPal 
before this PR, 
`starters/seedling` had just bumped spec_version and not Cargo.toml version
`starters/shell` was not bumped neither spec_version nor Cargo.toml version
`testing/rococo-parachain` had bumped both spec_version and Cargo.toml version
`starters/shell` was not bumped neither spec_version nor Cargo.toml version

so it is kind of really unclear, what should be bumped or not,
I would maybe prefer to do this in one way and not partially.

so, how do we want it finally?
bump **spec_version** and **Cargo.toml** for every runtime in release or no?

chevdor:
I see no reason to bump those.

First of all, the runtimes did not change.

Second, this is not a one shot bump. If we start bumping this now, we need to keep bumping everytime, all those runtimes, that did not change, for both the release branch and master. This is a lot of work for no benefit at all.

Those runtimes were NOT bumped on purpose, they were not forgotten :)

bkontur:
@chevdor 
so, there is no rule like "when spec_version is changed, we should also change Cargo.toml version" or something like that?

ok, no problem, I can close it

NachoPal:
I am fine if you guys do not want to bump those `spec_versions`, it is not critical, but the runtimes do change. They depend on different substrate and polkadot versions, therefore they are different runtimes.

I see benefit at least for bumping the testing runtimes (`penpal` and `rococo-parachain`). It is not the same testing interactions with a `v0.9.22` testing parachain or a `v0.9.27`.

NachoPal (approved):

PR #1524, opened 2022-08-09
Backport from release-parachains-v9270

bkontur:


NachoPal (approved):

chevdor:
Also the B0 label does not see to be correct as we backport at least one B1 PR.

PR #1528, opened 2022-08-10
Extrinsic ordering v2

chevdor:
This PR adds a new GH workflow using 2 binaries locally.
It compares the metadata changes between the reference binary and the one under test.
Despite the renaming, it does not change the [existing workflow](https://github.com/paritytech/cumulus/actions/workflows/extrinsic-ordering-check-from-bin.yml) which why I would favor this PR over #1515 

A test run can be seen [here](https://github.com/chevdor/cumulus/actions/runs/2831228118).


EgorPopelyaev (approved):

.github/workflows/release-21_extrinsic-ordering-check-from-two.yml
```diff
@@ -0,0 +1,124 @@
...
+      REF_URL: ${{github.event.inputs.reference_binary_url}}
+      BIN_REF: polkadot-parachain-ref
+      BIN_URL: ${{github.event.inputs.binary_url}}
+      BIN_BASE: polkadot-parachain
+      TMP: ./tmp
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - category: assets
```
coderobe, line 31: the category key doesn't seem to be used anywhere

.github/workflows/release-21_extrinsic-ordering-check-from-two.yml
```diff
@@ -0,0 +1,124 @@
...
+            runtime: statemine
+            local: statemine-local
```
coderobe, line 33: this could be deduplicated i think

coderobe (approved):

.github/workflows/release-21_extrinsic-ordering-check-from-two.yml
chevdor: I am not sure GH supports using the matrix within itself before it is defined.

.github/workflows/release-21_extrinsic-ordering-check-from-two.yml
chevdor: That could be removed indeed, it comes from the other workflow and documents the sub-folder where this runtime comes from. So besides this "documentation" there is no need today and I don't even really see a need later.

chevdor:
rebased master

.github/workflows/release-21_extrinsic-ordering-check-from-two.yml
coderobe: it could do the same it does on the polkadot side and add the -local where needed

chevdor:
bot rebase

chevdor:
bot rebase

PR #1533, opened 2022-08-10
[ci] change labels

alvicsam:
Close https://github.com/paritytech/ci_cd/issues/569


the-right-joyce:
image
`C1-low 📌` should be replaced by: `C1-low`

alvicsam:
Good catch, thanks!

the-right-joyce (approved):

rcny (approved):

scripts/ci/github/check_labels.sh
```diff
@@ -18,24 +18,21 @@ ensure_labels() {
...
 # Must be an ordered list of priorities, lowest first
 priority_labels=(
-  'C1-low 📌'
-  'C3-medium 📣'
-  'C7-high ❗️'
-  'C9-critical ‼️'
+  'C1-low'
+  'C3-medium'
+  'C7-high'
+  'C9-critical'
```
sergejparity, line 29:
```suggestion
  'C7-critical'
```
Looks like this should be only 'C7-critical' as per table https://www.notion.so/paritytechnologies/3e034a92359149199a33cfe85d17f611?v=54434c7bb30b4952a98280d1ff839249

sergejparity (changes requested):

scripts/ci/github/check_labels.sh
alvicsam: Thanks, also fixed `C7-high` -> `C5-high`

sergejparity (approved):

alvicsam:
bot rebase

coderobe (approved):

chevdor:
This PR should NOT be merged until other PRs are ready, especially around the release notes.

It brings major changes in the meaning of critical labels:
- B1
- D1 (major change)
- D2
- D3

More over, one part is missing.
If B7 is removed, what replaces it ?
We have 'B1-note_worthy' which is part of it but we miss the runtime part we talked about.

chevdor (changes requested):
See comments

the-right-joyce:
for now: no changes on the D-Labels, they will come in a separate batch
additionally: please, add a rule to enforce the user to use a X-labels if B1-note_worthy is tagged

chevdor (approved):
LGTM, before merging though:
- [ ] please make sure to clearly communicate to the Cumulus team that `B7` now becomes `B1` + `X1`
- [ ] we need a PR for the release notes that updates the logic from `B7` to `B1`+`X1`

chevdor:
Please do NOT merge until https://github.com/paritytech/cumulus/pull/1543 is also mergeable and ideally we merge both at the same time.

scripts/ci/github/check_labels.sh
```diff
@@ -46,6 +52,16 @@ else
   exit 1
 fi
 
+if has_label  "$repo" "$pr" 'B1-note_worthy'; then
```
chevdor, line 55: It would be better to check for the code only: `B1` rather than the label string: `B1-note_worthy`.
Here everything will collapse if we ever rename `B1-note_worthy` to something like `B1-Interesting ℹ️ ` (for whatever other text).

I would suggest adding a function such as:
```
get_code() {
    [[ $1 =~ ^([A-Z][0-9]).* ]] && echo "${BASH_REMATCH[1]}"
}
```
to extract the label "code" and compare only those label codes.

chevdor (changes requested):

scripts/ci/github/check_labels.sh
chevdor: It may be also convenient to handle a list of tuples with the code only and the full string:
```
for i in B1,B1-note_worthy X1,X1-whatever; do
    IFS=","
    set -- $i
    echo $1 and $2;
done
```

alvicsam: I guess adding the `get_code()` function requires some refactor. IMO it's not a good option to make this bash-script (or `lib.sh`) more complicated. At this moment I would suggest merging it as is and create an issue for the OpsTooling team to make a proper service for label checks for all 3 repos (cumulus, polkadot and substrate)

chevdor: ok, that works too

chevdor (approved):

Issue #1520, opened 2022-08-09
Tests in the `master` branch are broken


alvicsam:
https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/1728689

test name: `polkadot-parachain::polkadot_argument_parsing polkadot_argument_parsing`

<details><summary>Logs</summary>

```bash
--- STDOUT:                                        polkadot-parachain::polkadot_argument_parsing polkadot_argument_parsing ---
running 1 test
test polkadot_argument_parsing ... FAILED
failures:
failures:
    polkadot_argument_parsing
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 21.03s
--- STDERR:                                        polkadot-parachain::polkadot_argument_parsing polkadot_argument_parsing ---
2022-08-09 07:27:36 Polkadot parachain    
2022-08-09 07:27:36 ✌️  version 0.9.230-ed57c8aca96    
2022-08-09 07:27:36 ❤️  by Parity Technologies <admin@parity.io>, 2017-2022    
2022-08-09 07:27:36 📋 Chain specification: Local Testnet    
2022-08-09 07:27:36 🏷  Node name: foolish-oatmeal-0384    
2022-08-09 07:27:36 👤 Role: FULL    
2022-08-09 07:27:36 💾 Database: RocksDb at polkadot_argument_parsing/chains/local_testnet/db/full    
2022-08-09 07:27:36 ⛓  Native runtime: test-parachain-9230 (test-parachain-0.tx1.au1)    
2022-08-09 07:27:38 Running in --dev mode, RPC CORS has been disabled.    
2022-08-09 07:27:38 Parachain id: Id(1000)    
2022-08-09 07:27:38 Parachain Account: 5Ec4AhPZk8STuex8Wsi9TwDtJQxKqzPJRCH7348Xtcs9vZLJ    
2022-08-09 07:27:38 Parachain genesis state: 0x0000000000000000000000000000000000000000000000000000000000000000009b2a97568aa5bdad4c8204f44a150951fbea02000d76de717094fb804d64bf0303170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c11131400    
2022-08-09 07:27:38 Is collating: no    
2022-08-09 07:27:43 [Parachain] 🔨 Initializing Genesis block/state (state: 0x9b2a…bf03, header-hash: 0x4bf5…471b)    
[...]

PR #1534, opened 2022-08-10
[Fix] flaky node tests

ruseinov:
Fixes https://github.com/paritytech/cumulus/issues/1520

NOTE: This needs to be further refined as we can't wait for blocks on collators, hence for now it's just a timeout after getting the WS url. We shall see how to improve this in later iterations.

### DIFF
--- a/polkadot-parachain/Cargo.toml
+++ b/polkadot-parachain/Cargo.toml
@@ -92,6 +92,11 @@ substrate-build-script-utils = { git = "https://github.com/paritytech/substrate"
 assert_cmd = "2.0"
 nix = "0.24"
 tempfile = "3.3.0"
+tokio = { version = "1.17.0", features = ["macros", "time", "parking_lot"] }
+wait-timeout = "0.2"
+remote-externalities = { git = "https://github.com/paritytech/substrate", branch = "master" }
+node-primitives = { git = "https://github.com/paritytech/substrate", branch = "master" }
+
 
 [features]
 default = []
--- a/polkadot-parachain/Cargo.toml
+++ b/polkadot-parachain/Cargo.toml
@@ -94,9 +94,6 @@ nix = "0.24"
 tempfile = "3.3.0"
 tokio = { version = "1.19.2", features = ["macros", "time", "parking_lot"] }
 wait-timeout = "0.2"
-remote-externalities = { git = "https://github.com/paritytech/substrate", branch = "master" }
-node-primitives = { git = "https://github.com/paritytech/substrate", branch = "master" }
-
 
 [features]
 default = []
--- a/polkadot-parachain/Cargo.toml
+++ b/polkadot-parachain/Cargo.toml
@@ -92,7 +92,7 @@ substrate-build-script-utils = { git = "https://github.com/paritytech/substrate"
 assert_cmd = "2.0"
 nix = "0.24"
 tempfile = "3.3.0"
-tokio = { version = "1.17.0", features = ["macros", "time", "parking_lot"] }
+tokio = { version = "1.19.2", features = ["macros", "time", "parking_lot"] }
 wait-timeout = "0.2"
 remote-externalities = { git = "https://github.com/paritytech/substrate", branch = "master" }
 node-primitives = { git = "https://github.com/paritytech/substrate", branch = "master" }
--- a/polkadot-parachain/tests/common.rs
+++ b/polkadot-parachain/tests/common.rs
@@ -14,25 +14,159 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate.  If not, see <http://www.gnu.org/licenses/>.
 
+#![cfg(unix)]
+
+use assert_cmd::cargo::cargo_bin;
+use nix::{
+	sys::signal::{kill, Signal, Signal::SIGINT},
+	unistd::Pid,
+};
+use node_primitives::Block;
+use remote_externalities::rpc_api;
 use std::{
-	process::{Child, ExitStatus},
-	thread,
+	io::{BufRead, BufReader, Read},
+	ops::{Deref, DerefMut},
+	path::Path,
+	process::{self, Child, Command, ExitStatus},
 	time::Duration,
 };
+use tokio::time::timeout;
 
-/// Wait for the given `child` the given ammount of `secs`.
+/// Wait for the given `child` the given number of `secs`.
 ///
 /// Returns the `Some(exit status)` or `None` if the process did not finish in the given time.
-pub fn wait_for(child: &mut Child, secs: usize) -> Option<ExitStatus> {
-	for _ in 0..secs {
-		match child.try_wait().unwrap() {
-			Some(status) => return Some(status),
-			None => thread::sleep(Duration::from_secs(1)),
+pub fn wait_for(child: &mut Child, secs: u64) -> Result<ExitStatus, ()> {
+	let result = wait_timeout::ChildExt::wait_timeout(child, Duration::from_secs(5.min(secs)))
+		.map_err(|_| ())?;
+	if let Some(exit_status) = result {
+		Ok(exit_status)
+	} else {
+		if secs > 5 {
+			eprintln!("Child process taking over 5 seconds to exit gracefully");
+			let result = wait_timeout::ChildExt::wait_timeout(child, Duration::from_secs(secs - 5))
+				.map_err(|_| ())?;
+			if let Some(exit_status) = result {
+				return Ok(exit_status)
+			}
 		}
+		eprintln!("Took too long to exit (> {} seconds). Killing...", secs);
+		let _ = child.kill();
+		child.wait().unwrap();
+		Err(())
 	}
-	eprintln!("Took to long to exit. Killing...");
-	let _ = child.kill();
-	child.wait().unwrap();
+}
 
-	None
+/// Wait for at least n blocks to be finalized within a specified time.
+pub async fn wait_n_finalized_blocks(
+	n: usize,
+	timeout_secs: u64,
+	url: &str,
+) -> Result<(), tokio::time::error::Elapsed> {
+	timeout(Duration::from_secs(timeout_secs), wait_n_finalized_blocks_from(n, url)).await
+}
+
+/// Wait for at least n blocks to be finalized from a specified node
+pub async fn wait_n_finalized_blocks_from(n: usize, url: &str) {
+	let mut built_blocks = std::collections::HashSet::new();
+	let mut interval = tokio::time::interval(Duration::from_secs(2));
+
+	loop {
+		if let Ok(block) = rpc_api::get_finalized_head::<Block, _>(url.to_string()).await {
+			built_blocks.insert(block);
+			if built_blocks.len() > n {
+				break
+			}
+		};
+		interval.tick().await;
+	}
+}
+
+/// Run the node for a while (3 blocks)
+pub async fn run_node_for_a_while(base_path: &Path, args: &[&str], signal: Signal) {
+	let mut cmd = Command::new(cargo_bin("polkadot-parachain"))
+		.stdout(process::Stdio::piped())
+		.stderr(process::Stdio::piped())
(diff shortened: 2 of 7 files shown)


polkadot-parachain/tests/purge_chain_works.rs
```diff
@@ -15,59 +15,39 @@
...
+		.arg(base_dir.path())
+		.arg("-y")
+		.status()
+		.unwrap();
+	assert!(status.success());
 
-		// Make sure that the `parachain_local_testnet` chain folder exists, but the `db` is deleted.
-		assert!(base_path.path().join("chains/local_testnet").exists());
-		assert!(!base_path.path().join("chains/local_testnet/db/full").exists());
-		// assert!(base_path.path().join("polkadot/chains/dev").exists());
```
bkontur, line 70: @ruseinov 
I would maybe remove those lines, if not needed anymore

polkadot-parachain/tests/running_the_node_and_interrupt.rs
```diff
@@ -31,24 +32,10 @@ fn running_the_node_works_and_can_be_interrupted() {
...
-			"the process must exit gracefully after signal {}",
-			signal,
-		);
-	}
-
-	run_command_and_kill(SIGINT);
-	run_command_and_kill(SIGTERM);
+	let base_dir = tempdir().expect("could not create a temp dir");
+
+	let args = &["--", "--dev"];
```
bkontur, line 37: @ruseinov 
what about those original args `"-d", "interrupt_test", "--", "--dev"`, they are not needed?

polkadot-parachain/Cargo.toml
```diff
@@ -92,6 +92,11 @@ substrate-build-script-utils = { git = "https://github.com/paritytech/substrate"
 assert_cmd = "2.0"
 nix = "0.24"
 tempfile = "3.3.0"
+tokio = { version = "1.17.0", features = ["macros", "time", "parking_lot"] }
```
bkontur, line 95: @ruseinov 
other cumulus tokio deps, are `tokio = { version = "1.19.2" ...`, 
could we have it here the same?

polkadot-parachain/tests/running_the_node_and_interrupt.rs
ruseinov: Nope, please look at the base_dir logic.

polkadot-parachain/tests/purge_chain_works.rs
ruseinov: I do not want to change what was already there, simply adopting a new method of running the node. I think we should leave cleanups for any follow up PRs as I don't have any context as to why those were left commented out.

polkadot-parachain/Cargo.toml
ruseinov: It's a dev-dependency for tests, I don't believe that it matters much, but sure!

bkontur (approved):
lgtm

polkadot-parachain/Cargo.toml
```diff
@@ -92,6 +92,11 @@ substrate-build-script-utils = { git = "https://github.com/paritytech/substrate"
 assert_cmd = "2.0"
 nix = "0.24"
 tempfile = "3.3.0"
+tokio = { version = "1.19.2", features = ["macros", "time", "parking_lot"] }
+wait-timeout = "0.2"
+remote-externalities = { git = "https://github.com/paritytech/substrate", branch = "master" }
+node-primitives = { git = "https://github.com/paritytech/substrate", branch = "master" }
+
```
bkchr, line 99:
```suggestion
```

polkadot-parachain/Cargo.toml
```diff
@@ -92,6 +92,11 @@ substrate-build-script-utils = { git = "https://github.com/paritytech/substrate"
 assert_cmd = "2.0"
 nix = "0.24"
 tempfile = "3.3.0"
+tokio = { version = "1.19.2", features = ["macros", "time", "parking_lot"] }
+wait-timeout = "0.2"
+remote-externalities = { git = "https://github.com/paritytech/substrate", branch = "master" }
+node-primitives = { git = "https://github.com/paritytech/substrate", branch = "master" }
```
bkchr, line 98: Please remove. You should not use these crates here.

polkadot-parachain/tests/common.rs
```diff
@@ -14,25 +14,145 @@
...
+/// Wait for the given `child` the given number of `secs`.
 ///
 /// Returns the `Some(exit status)` or `None` if the process did not finish in the given time.
-pub fn wait_for(child: &mut Child, secs: usize) -> Option<ExitStatus> {
-	for _ in 0..secs {
-		match child.try_wait().unwrap() {
-			Some(status) => return Some(status),
-			None => thread::sleep(Duration::from_secs(1)),
+pub fn wait_for(child: &mut Child, secs: u64) -> Result<ExitStatus, ()> {
+	let result = wait_timeout::ChildExt::wait_timeout(child, Duration::from_secs(5.min(secs)))
```
gilescope, line 39: do you want `max` there? if secs is 30 it's going to choose 5 at the moment...

polkadot-parachain/tests/common.rs
```diff
@@ -14,25 +14,145 @@
...
+			built_blocks.insert(block);
+			if built_blocks.len() > n {
+				break
+			}
+		};
+		interval.tick().await;
+	}
+}
+
+/// Run the node for a while (3 blocks)
```
bkchr, line 84:
```suggestion
/// Run the node for a while (~3 blocks)
```

polkadot-parachain/tests/common.rs
```diff
@@ -14,25 +14,145 @@
...
+		.unwrap();
+
+	let stderr = cmd.stderr.take().unwrap();
+
+	let mut child = KillChildOnDrop(cmd);
+
+	let (ws_url, _) = find_ws_url_from_output(stderr);
+
+	// Let it produce some blocks.
+	let _ = wait_n_finalized_blocks(3, 30, &ws_url).await;
```
bkchr, line 102:
```suggestion
	let _ = wait_n_finalized_blocks(3, 120, &ws_url).await;
```

CI sometimes take more time.

polkadot-parachain/tests/common.rs
```diff
@@ -14,25 +14,145 @@
...
+			if built_blocks.len() > n {
+				break
+			}
+		};
+		interval.tick().await;
+	}
+}
+
+/// Run the node for a while (3 blocks)
+pub async fn run_node_for_a_while(base_path: &Path, args: &[&str], signal: Signal) {
```
ggwpez, line 85: We already have such functions in Substrate.  
Do you think it makes sense to generalize them and use here and in Polkadot?

ruseinov: It might, but right now we need to fix the flaky test that breaks all our builds. I'd prefer that to be done sooner rather than later. For generalising we shall create a separate issue I believe.

ggwpez: Okay mentioned here https://github.com/paritytech/substrate/issues/12003

ruseinov:
@ggwpez @bkchr I can't seem to get WS extraction to work, it takes so long for the chain to start up (at least on my local) that it never happens.

ggwpez:
> I can't seem to get WS extraction to work, it takes so long for the chain to start up (at least on my local) that it never happens.

Always run the tests with `--release`, otherwise it takes forever :laughing:

ruseinov:
> > I can't seem to get WS extraction to work, it takes so long for the chain to start up (at least on my local) that it never happens.
> 
> Always run the tests with `--release`, otherwise it takes forever 😆

Yeah, but it does not work on the CI as well.. I'm using nextest with release same as in the CI.

ruseinov:
> > > I can't seem to get WS extraction to work, it takes so long for the chain to start up (at least on my local) that it never happens.
> > 
> > 
> > Always run the tests with `--release`, otherwise it takes forever 😆
> 
> Yeah, but it does not work on the CI as well.. I'm using nextest with release same as in the CI.

Something's wrong, I've introduced a 30 seconds sleep before starting to read stderr to see if that helps the issue + printing all the lines. But it does not seem to help :/

ruseinov:
here's what I get:

```
2022-08-10 16:41:23 Polkadot parachain
2022-08-10 16:41:23 ✌️  version 0.9.230-unknown
2022-08-10 16:41:23 ❤️  by Parity Technologies <admin@parity.io>, 2017-2022
2022-08-10 16:41:23 📋 Chain specification: Local Testnet
2022-08-10 16:41:23 🏷  Node name: gentle-point-6281
2022-08-10 16:41:23 👤 Role: FULL
2022-08-10 16:41:23 💾 Database: RocksDb at /home/roman/.local/share/polkadot-parachain/chains/local_testnet/db/full
2022-08-10 16:41:23 ⛓  Native runtime: test-parachain-9230 (test-parachain-0.tx1.au1)
2022-08-10 16:41:24 Failed to remove the file used for the disk benchmark: No such file or directory (os error 2)
Error: Service(Client(Backend("IO error: While lock file: /home/roman/.local/share/polkadot-parachain/chains/local_testnet/db/full/LOCK: Resource temporarily unavailable")))
2022-08-10 16:41:25 Running in --dev mode, RPC CORS has been disabled.
2022-08-10 16:41:25 Parachain id: Id(1000)
2022-08-10 16:41:25 Parachain Account: 5Ec4AhPZk8STuex8Wsi9TwDtJQxKqzPJRCH7348Xtcs9vZLJ
2022-08-10 16:41:25 Parachain genesis state: 0x00000000000000000000000000000000000000000000000000000000000000000066bc141a7e79562e6bb2f09c0864773917b10ec038ea4e4e6881b15a5bfbb36b03170a2e7597b7b7e3d84c05391d139a62b157e78786d8c082f29dcf4c11131400
2022-08-10 16:41:25 Is collating: no
```

ruseinov:
@ggwpez it seems like this might be the culprit: 

```
2022-08-10 16:41:24 Failed to remove the file used for the disk benchmark: No such file or directory (os error 2)
Error: Service(Client(Backend("IO error: While lock file: /home/roman/.local/share/polkadot-parachain/chains/local_testnet/db/full/LOCK: Resource temporarily unavailable")))
```

ruseinov:
@ggwpez I think I know why, it's because the base_path has to go before the other args..

ggwpez:
Sorry but how is the collator binary supposed to produce any blocks?  
The `run_node_for_a_while` function waits for three blocks.

ruseinov:
> Sorry but how is the collator binary supposed to produce any blocks? The `run_node_for_a_while` function waits for three blocks.

Yes, I'm fixing it now. Because waiting for the WS url is enough for us to start killing the node. I'll just leave the timeout in.

gilescope:
Really we should move https://github.com/paritytech/substrate/blob/7d233c2446b5a60662400a0a4bcfb78bb3b79ff7/bin/node/cli/tests/common.rs to be in a lib that we can reference here rather than copying the code. Maybe in substrate test-utils ? I would not hold up this PR for that but I think that would be nice (as maybe polkadot repo might want to use the same stuff too).

ggwpez:
@gilescope https://github.com/paritytech/substrate/issues/12003

ruseinov:
> Really we should move https://github.com/paritytech/substrate/blob/7d233c2446b5a60662400a0a4bcfb78bb3b79ff7/bin/node/cli/tests/common.rs to be in a lib that we can reference here rather than copying the code. Maybe in substrate test-utils ? I would not hold up this PR for that but I think that would be nice (as maybe polkadot repo might want to use the same stuff too).

Yeah, we should definitely do that, but it'll need some tweaks due to some fundamental differences between the node types.

polkadot-parachain/tests/common.rs
ruseinov: Alright, I'll do 120

gilescope (approved):

PR #1538, opened 2022-08-10
contracts: Add experimental xcm chain extension

athei:
This adds XCM support to contracts by plugging in [this](https://github.com/paritytech/pallet-contracts-xcm/tree/master/runtime) chain extension. This is highly experimental but we want to push it out early so downstream teams can start experimenting using a testnet.

### DIFF
--- a/parachains/runtimes/contracts/contracts-rococo/Cargo.toml
+++ b/parachains/runtimes/contracts/contracts-rococo/Cargo.toml
@@ -52,6 +52,7 @@ pallet-sudo = { git = "https://github.com/paritytech/substrate", default-feature
 pallet-contracts = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 pallet-contracts-primitives = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 pallet-contracts-rpc-runtime-api = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
+pallet-contracts-xcm = { git = "https://github.com/paritytech/pallet-contracts-xcm", default-features = false, branch = "master" }
 
 # Polkadot
 kusama-runtime-constants = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "master" }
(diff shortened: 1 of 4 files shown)


cmichi (approved):

parachains/runtimes/contracts/contracts-rococo/src/contracts.rs
```diff
@@ -12,6 +12,7 @@ use pallet_contracts::{
 	weights::{SubstrateWeight, WeightInfo},
 	Config, DefaultAddressGenerator, DefaultContractAccessWeight, Frame, Schedule,
 };
+use pallet_contracts_xcm::{Config as XcmConfig, Extension as XcmExtension};
```
HCastano, line 15:
```suggestion
use pallet_contracts_xcm::{Config as ContractsXcmConfig, Extension as XcmExtension};
```

Don't think we want to conflate this with being the "general" XCM config

HCastano (approved):

parachains/runtimes/contracts/contracts-rococo/src/contracts.rs
athei: We are already in the contracts module. I think there is even a clippy lint that annoys you when your type is prefixed with the module name.

parachains/runtimes/contracts/contracts-rococo/Cargo.toml
```diff
@@ -52,6 +52,7 @@ pallet-sudo = { git = "https://github.com/paritytech/substrate", default-feature
 pallet-contracts = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 pallet-contracts-primitives = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 pallet-contracts-rpc-runtime-api = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
+pallet-contracts-xcm = { git = "https://github.com/paritytech/pallet-contracts-xcm", default-features = false, branch = "master" }
```
bkchr, line 55: This will make people creating releases on branches unhappy. Because they will need to add a branch to this repo and change the branches to the appropriate Substrate/Polkadot branches.

This will also make merging stuff that breaks the code in the repo hard. (assuming something in Substrate changed)

athei: What else can we do? Move that stuff into the polkadot repo?

bkchr: Or into the Cumulus repo :see_no_evil:

athei:
Okay I will need to merge it into polkadot then.

PR #1546, opened 2022-08-15
Easy testing of release note changes

gilescope:
By having example context file for runtime and client releases,
 we can now quickly see the results of any release note template changes we do.


chevdor:
I don't really mind merging that in but I think the usefullness is limited IMO.

The contexts are generated per release and many things can influence their content, one being rather current are the labels.
Do you plan on updating the reference contexts as labels change for instance ?

When I work on the templates, I usually load a local `.env` file (similar to what is your scripts) and generate fresh contexts.

As a matter of fact, you could easily do the same: if you run the scripts manually, you will get local contexts and they will be git-ignored by default.

coderobe:
I think it makes sense to have this, it makes prototyping the release notes easier without having to know the arcane combination of tools to generate it yourself.
This would also be nice to have on the polkadot side, IMO

coderobe (approved):

chevdor:
Ok, so who will take care of keeping the contexts up to date ? :)

chevdor:
bot rebase

gilescope:
When you run it and the context is too out of date to test what you're focused on then you can just grab the contexts generated by the lastest draft release.

chevdor:
I am no fan of storing release artifacts in the repo.
I would prefer a script that fetches the freshest artifacts from GH than having the artifacts themselves in the repo as they will very likely become obsolete and useless very soon.

gilescope:
contexts are too big and need to be got from somewhere...

PR #1556, opened 2022-08-19
Add new blank issue template

chevdor:
This is to help having an option to create a new issue without having to search for the mini link down the list.

## Before:
image

## After:
image


bkchr (approved):

gilescope:
Looks better - I agree it seemed a bit strange before.

Issue #432, opened 2021-05-10
Remove "dead" leaves


bkchr:
When proposing new PoV blocks with a collator, it can happen that we are "stuck" as the relay chain for whatever reason doesn't include our blocks. After X blocks the collator will start to fail importing new blocks, because we have imported too many leaves.

Using https://github.com/paritytech/substrate/pull/8533 we will be able to delete "dead" leaves.

We will require some criterias to decide wether a leave should be seen as "dead":

1. A block was not seconded, means we can directly remove it.
2. We have produced block X for relay chain block Y. Now we start producing for Y + N, means we should be able to remove all blocks we have produced/imported for Y.

xlc:
I hope this can be done before Beyond the End of the Century 😅

bkchr:
Maybe :P

xlc:
One of our testnet RPC node crashed again due to this and looks like the only solution is to purge db and resync.

davxy:
I think that could be a good idea to generalize the leaves pruning feature to Substrate and not limit it to Cumulus.

The idea is to allow the user to limit the lifetime of chain leaves before getting pruned.

This leaves pruning strategy can thus be enabled where it is more appropriate, for example to solve the current observed behavior in cumulus where new blocks are dropped forever because we've saturated the allowed leaves per height (32).

For the pruning criteria I see three options
1. allow the user to specify the pruning criteria (e.g. via some closure) < Flexibility over simplicity of usage >
2. we decide what is the pruning criteria (e.g. leaf lifetime) < Simplicity of usage over flexibility >
3. allow the user to specify the pruning criteria but also provide a default implementation (e.g. leaf lifetime)

---

### Pruning criteria using lifetime.

When a new leaf is added to the chain then we are going to start some kind of timer in order to limit the leaf lifetime.

The timer for a leaf 'L' is stopped if a child C is added to L, in this case the timer is restarted for the new leaf C.

If a leaf L lifetime is over, then we are going to remove all its parents down to one of the following nodes
(assuming F is the last finalized block):

1. If there are forks between L and F, then we stop pruning nodes as soon as we reach one fork point.
    Example:
    ````
    F - B0 - B1
         \ - B2 - L
    ````
    If lifetime of 'L' is over, then we remove 'L' and 'B2'. 
    We stop at 'B0' since there is another fork depending on it.

2. If there are no forks between L and F, then we stop pruning at F.

skunert:
[...]

PR #1559, opened 2022-08-19
Limit number of blocks per level (2nd attempt)

davxy:
Closes https://github.com/paritytech/cumulus/issues/432

---

This PR prevents the `StateDbError::TooManySiblingBlocks` error from being triggered by eagerly removing leaves from the backend on block import and before the error condition is met.

---

⚠ SUPERSEEDS: https://github.com/paritytech/cumulus/pull/1544

The main difference is this is a portable solution WRT the leaves order read from the backend.


client/consensus/relay-chain/src/import_queue.rs
```diff
@@ -101,25 +103,27 @@ where
...
 	<Client as ProvideRuntimeApi<Block>>::Api: BlockBuilderApi<Block>,
 	CIDP: CreateInherentDataProviders<Block, ()> + 'static,
+	BE: sc_client_api::Backend<Block> + 'static,
 {
 	let verifier = Verifier::new(client, create_inherent_data_providers);
 
 	Ok(BasicQueue::new(
 		verifier,
-		Box::new(cumulus_client_consensus_common::ParachainBlockImport::new(block_import)),
+		Box::new(ParachainBlockImport::new(block_import, backend, LeavesLevelLimit::None)),
```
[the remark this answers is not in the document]
davxy: Ty. It was wrong

client/consensus/common/src/lib.rs
```diff
@@ -69,46 +81,173 @@ impl<B: BlockT> ParachainConsensus<B> for Box<dyn ParachainConsensus<B> + Send +
...
+const CLEANUP_THRESHOLD: usize = 64;
+
+impl<Block: BlockT, BE: Backend<Block>> LeavesLevelMonitor<Block, BE> {
+	fn update(&mut self, hash: BlockHash<Block>, parent_hash: BlockHash<Block>) {
+		self.import_counter += 1;
+		self.leaves_cache.insert(hash, self.import_counter);
+		self.leaves_cache.remove(&parent_hash);
+	}
+
+	fn check(&mut self, number: NumberFor<Block>) {
```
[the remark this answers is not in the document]
davxy: You are absolutely right.

Indeed - don't know why - but I was strangely convinced that the `MAX_BLOCKS_PER_LEVEL` in the overlay was only concerning the leaves. And this is 100% wrong.

https://github.com/paritytech/substrate/blob/28f0cb18c96fb82f810f1d9be69fb609fbb7217e/client/state-db/src/noncanonical.rs#L292-L294

This complicates the things a bit.  I open a discussion at the end of the PR to talk this.

davxy:
Following https://github.com/paritytech/cumulus/pull/1559#discussion_r951126469, let's recap the problems to be solved:
1. I have to  fetch the list of blocks at the level of the block that I'm importing.
2. if this list length is >= the threshold then I then have to choose which one to remove.
3. The block to remove may not be a leaf (e.g. because what you've just said). So I have to also find all its descendants and prune everything.

I don't see a direct support from the Backend to solve the issues, let's see if we can somehow work around the limitations.

1. This is probably the most annoying. Is there a easy and possibly low cost way to get all blocks at one level?
2. The choice will be again driven by the "age". Probably the best thing to do is to consider the age of the most recent descendant of the blocks under analysis (for example if a block has a very old leaf and a very fresh one, then we consider the block age equal to the most fresh one).
3. Once we have all the leaves of the chosen block, we can remove all its descendants (starting from the leaves) by using the  `Backend::remove_leaf_block`.

A block descendants can be recovered using the leaves can be found using the `sp_blockchain::three_route` (is expensive... but also a very sporadic operation).

So the major blocker here is how to get the blocks list at one level?

I'm open to more simple suggestions :-)

bkchr:
> * issue 1. This is probably the most annoying. Is there already an easy and possibly low cost way to get all blocks at one level?

`HashMap<BlockNumber, Vec<BlockHash>>` :see_no_evil: 

Pruning will be easy when we do pruning on finality. We will just need to `hash_map.remove(last_pruned..last_finalized)`
We would just keep this list locally here, as we observe all imported blocks anyway.

> * issue 2. The choice will be again driven by the "age". Probably the best thing to do is: for each block at the same height of the imported one its "age" is set as the age of its most recent descendant leaf (for example if a block is the parent a very old leaf and a very fresh one, then we consider the block age equal to the most fresh leaf).

Sounds good to me.

davxy:
> > * issue 1. This is probably the most annoying. Is there already an easy and possibly low cost way to get all blocks at one level?
> 
> `HashMap<BlockNumber, Vec<BlockHash>>` 🙈

👍 ok so I will maintain this metadata locally here. I was hoping for something already in place hidden very well somewhere 😀  Ty

bkchr:
The noncanonical overlay already stores the same mapping. However, I don't see any problem of keeping it here locally as well.

polkadot-parachain/src/service.rs
```diff
@@ -1123,7 +1150,7 @@ where
 
 	Ok(BasicQueue::new(
 		verifier,
-		Box::new(ParachainBlockImport::new(client)),
+		Box::new(ParachainBlockImport::new(client, backend, LeavesLevelLimit::Default)),
```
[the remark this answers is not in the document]
davxy: Yes a limit below the default one is used only for the tests. I can create a new constructor

client/consensus/common/src/lib.rs
```diff
@@ -69,46 +81,173 @@ impl<B: BlockT> ParachainConsensus<B> for Box<dyn ParachainConsensus<B> + Send +
...
+	// Map between block hash and age.
+	leaves_cache: HashMap<BlockHash<Block>, u64>,
+	// Backend reference to remove leaves on level saturation.
+	backend: Arc<BE>,
+}
+
+// Threshold after which we are going to cleanup our internal leaves cache by removing hashes that
+// doesn't belong to leaf blocks anymore.
+// This is a farly arbitrary value that can be changed in the future without breaking anything.
+const CLEANUP_THRESHOLD: usize = 64;
```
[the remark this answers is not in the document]
davxy: In order to prevent adding another trait bound to get the finality notifications I opted out for a simpler and cheaper solution that in the end gives the same net result.

I just check if the last finalized block is changed since the last cleanup, if it is, I proceed to the cleanup.

Since in normal circumstances the finalization proceeds approximately at the same pace as the block authoring/import. I also opted out to don't attempt the cleanup on every imported block. But to check if there is something to cleanup every `CLEANUP_THRESHOLD` imported blocks.
(**on averaege** I prefer to cleanup a bunch of information in one shot instead of cleaning one block at a time)

Overall this prevents passing an extra param or adding an extra bound for a type implementing `BlockchainEvents` (that propagates all the way up as the `Blockchain` generic I've added).

davxy:
You may notice that the "proper" initialization (i.e. using the backend information) of the `LevelMonitor` instance is postponed to the first imported block (i.e. the first call to `LevelMonitor::enforce_limit`.

I would preferred to do that in the `LevelMonitor::new()` method directly, but that imply adding the bound `BE: Backend`.
That is not a big matter, is just that it then propagates to the `ParachainBlockImport` as well and all the way up.

In the end, to prevent modifying too much unrelated stuff, I've opted out for this lazy initialization (I hope it is not too disturbing :-D ).
In practice, the end result is the same.

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,247 @@
...
+					if leaf_idx >= candidate_fresher_leaf_idx ||
+						route.common_block().hash == best_hash ||
+						route.enacted().iter().any(|entry| entry.hash == best_hash)
+					{
+						leaves_routes.clear();
+						break
+					}
+
+					if leaves_routes.is_empty() {
+						fresher_leaf_idx = leaf_idx;
```
skunert, line 194: Not entirely sure, but don't we also need to update `fresher_leaf_idx` when we come across `leaf_idx > fresher_leaf_idx`?

skunert (commented):
Looks good, just some nits

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,247 @@
...
+						log::warn!(
+							target: "parachain",
+							"Unable getting route from {} to {}: {}",
+							blk_hash, leaf_hash, e,
+						);
+						continue
+					},
+				};
+				if route.retracted().is_empty() {
+					if leaf_idx >= candidate_fresher_leaf_idx ||
```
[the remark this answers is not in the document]
davxy: Yeah. I think so :-)

davxy:
@skunert I just updated the procedure so that it is more clear and more optimized (along with some more comments).

The basic idea here is:
1. for each block find out what is its younger/fresher descendant leaf. Let's say this becomes the block "freshness".
2. remove the candidate that is less fresh.

I hope that with the new modifications/comments the code is more clear.

client/consensus/common/src/level_monitor.rs
davxy: Code has been just updated to be more readable and optimized. Let me know if it is clear now

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,269 @@
...
+		// one. Should be considered that in normal conditions the number of blocks to remove is 0
+		// or 1, it is not worth to complicate the code too much.
+		let remove_count = level.len() - self.level_limit + 1;
+		(0..remove_count).for_each(|_| self.remove_block(number, &leaves));
+	}
+
+	/// This will search for the less fresh block and removes it along with all its descendants.
+	/// Leaves should have already been ordered by "freshness" (older first).
+	fn remove_block(&mut self, number: NumberFor<Block>, leaves: &[Block::Hash]) {
+		let mut candidate_fresher_leaf_idx = usize::MAX;
```
skunert, line 159: super nit: maybe rename to `candidate_freshest_leaf_idx`?

skunert:
Thanks for the improvements, I think its more clear now!

skunert (approved):

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,269 @@
...
+	/// This will search for the less fresh block and removes it along with all its descendants.
+	/// Leaves should have already been ordered by "freshness" (older first).
+	fn remove_block(&mut self, number: NumberFor<Block>, leaves: &[Block::Hash]) {
+		let mut candidate_freshest_leaf_idx = usize::MAX;
+		let mut candidate_freshest_route = None;
+
+		let blockchain = self.backend.blockchain();
+		let best_hash = blockchain.info().best_hash;
+
+		for blk_hash in self.levels.entry(number).or_default().iter() {
```
bkchr, line 165: Shouldn't we also not start here with the oldest block? If we change `levels` to `HashMap<Hash, Vec<>>` we could get the ordering oldest to freshest for free.

bkchr (changes requested):
I think that there some bugs around removing blocks on the requested level.

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,269 @@
...
+		let mut counter_max = 0;
+		let finalized = self.backend.blockchain().info().finalized_hash;
+
+		for leaf in self.backend.blockchain().leaves().unwrap_or_default() {
+			let route = sp_blockchain::tree_route(self.backend.blockchain(), finalized, leaf)
+				.expect("Route from finalized to leaf should be available; qed");
+			if !route.retracted().is_empty() {
+				continue
+			}
+			std::iter::once(route.common_block()).chain(route.enacted()).for_each(|elem| {
```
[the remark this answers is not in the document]
davxy: You are right. Common block is always the finalized one. I moved the insertion of that one out of the loop

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,269 @@
...
+		self.import_counter += 1;
+
+		if self.import_counter % CLEANUP_THRESHOLD == 0 {
+			// Do cleanup once in a while, we are allowed to have some obsolete information.
+			let finalized_num = self.backend.blockchain().info().finalized_number;
+			let delta: u64 =
+				finalized_num.saturating_sub(self.lowest_level).unique_saturated_into();
+
+			for i in 0..delta {
+				let number = finalized_num + i.unique_saturated_into();
```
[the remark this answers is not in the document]
davxy: `Range` implements `Iterator` only if its `Item` associated type implements `Step`:
https://doc.rust-lang.org/std/ops/struct.Range.html#impl-Iterator

But unfortunately our generic `Number` is not bound by `Step`.

Nor can be manually implemented because it requires the unstable feature `step_trait`:
https://github.com/rust-lang/rust/issues/42168

Step is only directly implemented by Rust std/core for primitive numeric types (https://doc.rust-lang.org/core/iter/trait.Step.html#implementors).

If there is a workaround or if I miss something please correct me.

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,269 @@
...
+
+		let mut leaves = self.backend.blockchain().leaves().unwrap_or_default();
+		// Sort leaves by freshness (most recent first).
+		leaves.sort_unstable_by(|a, b| self.freshness.get(a).cmp(&self.freshness.get(b)));
+
+		// This may not be the most efficient way to remove **multiple** entries, but is the clear
+		// one. Should be considered that in normal conditions the number of blocks to remove is 0
+		// or 1, it is not worth to complicate the code too much.
+		let remove_count = level.len() - self.level_limit + 1;
+		(0..remove_count).for_each(|_| self.remove_block(number, &leaves));
```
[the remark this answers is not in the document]
davxy: Removing one single block is the "normal" case.

I've added this because it may happen that we require to delete more than one block in the following cases:
1. the limit is changed  to something smaller. In this case the DB was previously pruned using the old bigger limit.
   The first time this code runs it will remove more than one block to enforce the limit.
2. The first time this code is enacted. This is a special case of 1 (with previous "bigger" limit equal to infinite).

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	/// This will search for the less fresh block and removes it along with all its descendants.
+	/// Leaves should have already been ordered by "freshness" (older first).
+	fn remove_block(&mut self, number: NumberFor<Block>, leaves: &[Block::Hash]) {
+		let mut candidate_freshest_leaf_idx = usize::MAX;
+		let mut candidate_freshest_route = None;
```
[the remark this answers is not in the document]
davxy: :-D

Let me first try to explain better the idea here:
1.Given a set of candidate blocks at the same level we want to choose one (THE target)
2. Each of these candidates may have a bunch of leaves  (if not, the candidate block itself is the leaf).
3. For each candidate we select the freshest leaf. 
    Let's say that this freshest leaf "freshness" value becomes the candidate freshness.
4. The target is chosen to be the one that is less fresh (according to 3 criteria).

Now, regarding the two variables. These are used **only** to optimize some steps.

Usage:
- `candidate_freshest_leaf_idx`: if during the loop we find another block (at the same level of the current candidate) with a leaf that is fresher (or equally fresh) than `candidate_freshest_leaf_idx` than we can early exit processing that block since for sure it is not a candidate (because it has a leaf that is fresher than the current candidate).
- `candidate_freshest_route`:  route to candidate freshest leaf. This is just to avoid getting the route to this leaf again on children removal [here]()

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,269 @@
...
+			});
+		};
+
+		remove_route(candidate_freshest_route);
+
+		let to_skip = leaves.len() - candidate_freshest_leaf_idx;
+		leaves.iter().rev().skip(to_skip).for_each(|leaf_hash| {
+			match sp_blockchain::tree_route(blockchain, candidate_hash, *leaf_hash) {
+				Ok(route) if route.retracted().is_empty() => {
+					remove_route(route);
```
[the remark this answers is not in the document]
davxy: > None of this guarantees that you remove a block at level number.

Why? If there are not bugs somewhere in the backend this code should end up removing something 

---

The strategy in my implementation is:

> The freshness of each block at level `number` is equal to the freshness of its freshest leaf 

Sorry about all this freshness words :-D 

I've chosen this strategy because the sum may end up choosing not the best candidate (IMO).

For example lets assume I have this situation (attached the blocks freshness):

```
Level:   |   1   |   2   |   3   |   4   |
            
            A(1)----B(2)----C(3)----G(7)
              \
               +----D(4)----E(5)
                     \
                      +-----F(6)
```

If we want to remove something at level 2
- The "sum" strategy suggests to remove B (SUM @ D = 15, SUM for B = 12)
- The strategy I've implemented suggests to remove D  (fresher leaf @ D = 6, fresher leaf @ B = 7)


---

Note that I've updated the code and I've isolated the removal code from the lookup code.
In this way we can eventually change the target search code easily

client/consensus/common/src/level_monitor.rs
davxy: This is true. If you use `HashMap<Hash, Vec<>>` you already find the entries within the vector sorted by arrival order.

But the point of my overall strategy is not to remove the older element at a given level. But to remove the one according to the search strategy I've described in my other comments.

So having it already sorted is not super helpful.

I went for an `HashMap<Hash, HashSet<>>` because the HashSet is faster for removal of elements at random indices.

davxy:
@arkpar 

A Substrate PR you've worked introduced the `MAX_BLOCKS_PER_LEVEL` limit in the backend (https://github.com/paritytech/substrate/pull/8494)

After that someone experienced stalled parachain nodes because of level saturation (https://github.com/paritytech/substrate/issues/10476) in Cumulus we've decided to prevent the error to be triggered by checking if the number of blocks at the given level is greater than a configurable threshold (by default equal to `MAX_BLOCKS_PER_LEVEL`) and to eventually eagerly remove on of the blocks (chosen via some criteria) in order to make space for the fresher one.

Now before merging this sort of "workaround" we would like to know if maybe the problem can be directly (and easily) resolved in Substrate. Is the `MAX_BLOCKS_PER_LEVEL` limit really necessary (I understand that this is a defensive mechanism) or can it can be incremented over a value that will never be realistically reached?

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,325 @@
...
+
+		let level = match self.levels.get(&number) {
+			Some(level) => level,
+			None => return None,
+		};
+
+		for blk_hash in level.iter().filter(|hash| **hash != best_hash) {
+			// Search for the fresher leaf information for this block
+			let candidate_info =
+				leaves.iter().enumerate().rev().find_map(|(leaf_idx, leaf_hash)| {
```
bkchr, line 205: We should filter out leaves that we already successfully found with `route.retracted().is_empty()`.

And why not start with the oldest leaves? Wouldn't that make more sense?

bkchr (approved):
Thank you!

This latest iteration looks much better than the ones before! Much easier to read and to understand!

I left some comments on things that should be improved, but overall logic is looking good.

Nice pr! :)

bkchr:
And sorry for the delay :see_no_evil:

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,325 @@
...
+	/// No explicit limit, however a limit may be implicitly imposed by the backend implementation.
+	None,
+	/// Custom value.
+	Some(usize),
+}
+
+// Support structure to constrain the number of leaves at each level.
+pub struct LevelMonitor<Block: BlockT, BE> {
+	// Max number of leaves for each level.
+	pub level_limit: usize,
```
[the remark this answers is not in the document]
davxy: Some fields are read in the `tests` module assertions.
But only some of them and indeed they don't have to be `pub`, for them a `pub(crate)` would suffice.

client/consensus/common/src/level_monitor.rs
```diff
@@ -0,0 +1,325 @@
...
+		// Sort leaves by freshness (less fresh first).
+		leaves.sort_unstable_by(|a, b| self.freshness.get(a).cmp(&self.freshness.get(b)));
+
+		// This may not be the most efficient way to remove **multiple** entries, but is the easy
+		// one :-). Should be considered that in "normal" conditions the number of blocks to remove
+		// is 0 or 1, it is not worth to complicate the code too much. One condition that may
+		// trigger multiple removals (2+) is if we restart the node using an existing db and a
+		// smaller limit wrt the one previously used.
+		let remove_count = level_len - self.level_limit + 1;
+		(0..remove_count).all(|_| {
```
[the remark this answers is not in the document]
davxy: Within the iteration the only thing that could fail is `find_target()`.
If it fails once then it will fail for all the other attempts as well.
I've used `all` to stop as soon as it fails the first time

client/consensus/common/src/level_monitor.rs
davxy: > We should filter out leaves that we already successfully found with route.retracted().is_empty().

Good catch

> And why not start with the oldest leaves? Wouldn't that make more sense?

We start from the fresher leaves to stop as soon as we find one descending from the node under processing (we are looking for its fresher descendant).

bkchr: > We start from the fresher leaves to stop as soon as we find one descending from the node under processing (we are looking for its fresher descendant).

Yeah I got this, my question was more, wouldn't it be faster to directly start with the oldest (less freshest leave)? I mean yes, it always depends on how the blocks on the level are ordered in relation to the leaves. However, in theory we should be able to find the oldest leave directly for the first block, which on the other way around, as you are currently doing it, wouldn't be possible.

arkpar:
> Now before merging this sort of "workaround" we would like to know if maybe the problem can be directly (and easily) resolved in Substrate. Is the MAX_BLOCKS_PER_LEVEL limit really necessary (I understand that this is a defensive mechanism) or can it can be incremented over a value that will never be realistically reached?

Each imported block that is not yet finalized requires memory and disk resources, so we can't just add them indefinitely. In addition to that, there are a few places in the substrate code that expect the number of active leaves to be reasonable.

IMO proper solution would be to just stop producing blocks if they are not being included, requiring manual intervention to resume the process. Or at least introduce some kind of exponential delay between attempts to produce a block.

davxy:
Finally I think I've found the REAL issue here 🫠

The storage backend stores the blocks using the post seal hash as the key (as it should be).
In the cache I was using the pre seal one instead (i.e. BlockImportParams::header.hash() was returning the pre seal one since the seal has been already removed at this point).
  
I should use `import_params.post_hash()` instead 🙈

https://github.com/paritytech/cumulus/blob/3cd5d313e20e00dbd361ae719a296e949ab912c3/client/consensus/common/src/lib.rs#L138-L139

---

The test was not detecting this since without any `post_digest` entry in the `BlockImportParams` we have post == pre hash.

To exercise the fix I added a dummy entry (variant `DigestItem::Other`) during test block construction.
(this may be useful to catch other bugs of this kind in the future)

https://github.com/paritytech/cumulus/blob/3cd5d313e20e00dbd361ae719a296e949ab912c3/client/consensus/common/src/tests.rs#L137-L153

Now the old code fails with the same errors as seen in *Versi* testnet.

---

I'm going to deploy this code in *Versi* and then we're going to deploy *Malus* validators to see how it behaves this time

davxy:
# Update

Parachain import queue and authoring engine construction functions where modified to take a generic block import type bounded by `BlockImport` trait and not directly the `ParachainBlockImport` type (this is as it was before recent https://github.com/paritytech/cumulus/pull/1782)

The generic block import has been additionally bounded by a `ParachainBlockImportMarker` marker.

---

The `ParachainBlockImportMarker` is currently only implemented by `ParachainBlockImport` itself, but in the future may be implemented by any other block import types wrapping `ParachainBlockImport` for test purposes.

But the main motivation for this marker is to not leak the details about `ParachainBlockImport` all over the laces where it is used. (e.g. that internally it is requiring a `Backend` instance).

If we are not using a marker trait then every place currently touched by a `ParachainBlockImport<Block, BlockImport, Backend>` should take an additional `Backend` generic. And this is a detail I prefer to avoid and keeps the overall design more clean.

---

As a side node we may want to rename:
- `struct` `ParachainBlockImport` to `ParachainBasicBlockImport`
- `trait` `ParachainBlockImportMarker` to `ParachainBlockImport`

---

For the moment the code still contains `XXX` logs for testing

client/consensus/common/src/lib.rs
```diff
@@ -74,47 +83,100 @@ impl<B: BlockT> ParachainConsensus<B> for Box<dyn ParachainConsensus<B> + Send +
...
+		}
+
+		Ok(res)
 	}
 }
+
+/// Marker trait denoting a block import type that fits the parachain requirements.
+pub trait ParachainBlockImportMarker {}
+
+impl<B: BlockT, BI, BE> ParachainBlockImportMarker for ParachainBlockImport<B, BI, BE> {}
```
bkchr, line 182: This should probably be sealed.

client/consensus/common/src/lib.rs
```diff
@@ -74,47 +83,100 @@ impl<B: BlockT> ParachainConsensus<B> for Box<dyn ParachainConsensus<B> + Send +
 /// This is used to set `block_import_params.fork_choice` to `false` as long as the block origin is
 /// not `NetworkInitialSync`. The best block for parachains is determined by the relay chain. Meaning
 /// we will update the best block, as it is included by the relay-chain.
-pub struct ParachainBlockImport<I>(I);
+pub struct ParachainBlockImport<Block: BlockT, BI, BE> {
+	inner: BI,
+	monitor: Option<SharedData<LevelMonitor<Block, BE>>>,
```
bkchr, line 88: Using a `Mutex` here should be much simpler. We don't need this data being locked from somewhere else.

davxy: (Correct me if I overlooked something or I didn't get your suggestion)

What if the block import method is called concurrently by two different components (e.g. by the authoring and import queue)?

If I release the lock between the read (before inner import) and the write of the cache (after inner import), a race condition may happen.

Assume this scenario
- we have a monitor limit set to something less than the hard limit in the backend (lets say this is infinite for the sake of this example).
- two components sharing the block import object.
- both of them import a block
- both of them read that the num of elements per level is `limit-1`.
- they don't prune nothing from
- in the end they will both add the imported block to the cache
- resulting in a num of blocks per level equal to `limit+1`.

Obviously (in this scenario) **if the monitor limit is equal to the hard limit of the backend** (32) then the second `inner.block_import` will fail with the `TooManySiblingBlocks` error (thus the second element is not imported in the cache) but we don't want this error to be triggered.

client/consensus/common/src/lib.rs
davxy: Ok. I'll make it sealed.

What do you think about renaming:
```
    struct ParachainBlockImport to ParachainBasicBlockImport
    trait ParachainBlockImportMarker to ParachainBlockImport
```

And then use a "supertrait" for `ParachainBlockImport` trait:

```
trait ParachainBlockImport<...>: BlockImport<...> {}
```

---

As I said the advantage of doing so would be that the users can do:

```
BI: ParachainBlockImport<B>
```

While currently (with the simple marker) they should do:

```
BI: BlockImport<B> + ParachainBlockImportMarker
```

client/consensus/common/src/lib.rs
bkchr: Yeah, good points :+1: I didn't thought of importing a build block and importing a block from the network.

client/consensus/common/src/lib.rs
bkchr: Supertrait sounds like a good idea. The naming. IDK. Naming is hard :P "Basic" sounds weird TBH.

sandreim:
I tested this PR on Versi and I can confirm it fixes the issue.

davxy:
> I tested this PR on Versi and I can confirm it fixes the issue.

@sandreim awesome!

I'm working on an improvement that prevents the removal of blocks that were "available" according to the relay chain. This should be a bit safer, but it requires some rework and design considerations...

Maybe, if this already improves the experience and is required asap, we can merge this *"as is"* and add the improvement I'm working on in a separate PR?

---

Nevertheless, as said, I'm working on this improvement that **soon™** will land.

Here is a draft PR https://github.com/paritytech/cumulus/pull/1864 I already opened to share some ideas and gain suggestions...

Mostly is the implementation of some of the ideas me and @bkchr have talked about

I'm experimenting with it

I'm going thus continuing this conversation on the draft PR

davxy:
(PS: Obviously before merge I'm going to remove the `XXX` comments :-)

client/consensus/common/src/lib.rs
bkchr: This was not done yet or @davxy?

bkchr:
> Maybe, if this already improves the experience and is required asap, we can merge this _"as is"_ and add the improvement I'm working on in a separate PR?

I would be okay with doing this :)

client/consensus/common/src/lib.rs
davxy: No, not yet. I would be happy to do this before merge but I've been held back by the *naming dilemma*.

To me as the general trait to import blocks is `BlockImport` makes sense to rename the parachain one as `ParachainBlockImport`.

But then we have to find a proper name for the provided implementation :-D
Indeed `BasicParachainBlockImport` sounds weird to me as well but I can't find something better...

We have to find a proper prefix or leave it as is (and use a marker, that in the end is not so bad)

bkchr:
```
struct ParachainBlockImport;

trait ParachainBlockImportMarker
```

Still sounds good to me.

bkchr:
@davxy what is the status with merging this earlier?

davxy:
> @davxy what is the status with merging this earlier?

Confident to merge this very soon. I just need to add a small safety net

sandreim:
bot rebase

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2141846

bkchr:
@davxy can we merge this?

davxy:
bot merge

PR #1567, opened 2022-08-23
shell and seedling did not build

gilescope:
(even though CI was green). Time for more CI. These additional CI builds only trigger if the relevant runtimes have been altered in any way.

The commented out of shell should make that build fail proving that the CI now picks this up.


alvicsam:
Why not run them in Gitlab? I think it's going to be quicker than on GitHub instances. Gitlab pipelines also have possibility to run jobs only if some [files were changed](https://docs.gitlab.com/ee/ci/yaml/index.html#ruleschanges).

chevdor:
I agree with @alvicsam. @giles and I had a chat and I also mentioned that I would **not** use `srtool` here and enjoy as much caching as we possibly can. 

The goal here is not to build a release runtime but only check if this/these runtime(s) are buildable and catch breakages as early as possible.

The open point was whether it is worth trying to be smart about **what** runtime needs a rebuild or rely on our caching and rebuild all the runtimes everytime but I think I know what  @alvicsam's take will be 😄

muharem (approved):

ruseinov (approved):

parachains/runtimes/starters/shell/Cargo.toml
```diff
@@ -74,6 +74,7 @@ std = [
 	"cumulus-primitives-core/std",
 	"cumulus-primitives-utility/std",
 	"parachain-info/std",
+#	"parachains-common/std",
```
bkchr, line 77:
```suggestion
```

bkontur, line 77: @gilescope 
it is not needed here? because dependency `parachains-common = { path = "../../../common", default-features = false }` is there on the line 42

gilescope, line 77:
```suggestion
	"parachains-common/std",
```

gilescope: Have uncommented as we're not testing the CI changes in this PR any more.

gilescope:
bot merge

PR #1576, opened 2022-08-29
Replace the label checks using bash script by `ruled-labels`

chevdor:
Instead of having the label checking rules coded as [`bash script`](https://github.com/paritytech/cumulus/blob/master/scripts/ci/github/check_labels.sh), this PR introduce the same rules that we currently have in a [`ruled-labels`](https://github.com/chevdor/ruled_labels) `specs.yaml` as well as the matching tests in `tests.yaml`.

TODOs:
- [x] validate the rules (correctness and completeness) - cc @the-right-joyce 
- [x] fix the the new script to fetch the labels - cc @alvicsam 
- [x] Decide about usage and which Docker image to build / use
- [x] add a new CI job with failure allowed to check the labels using the new method
- [x] add a CI job when `specs.yaml` changes to re-run the tests from `tests.yaml`

The equivalent rules from [this script](https://github.com/paritytech/cumulus/blob/master/scripts/ci/github/check_labels.sh) can be found [here](https://github.com/paritytech/cumulus/blob/wk-220829-ruled_labels/scripts/ci/ruled_labels/specs.yaml#L202-L220).


the-right-joyce:
I'd change the priority check to: 

- name: Release need criticality 
    when: !one_of [ B1 ]
    spec:
      require: !one_of [ C* ]

chevdor:
@alvicsam I added a [simple GH Workflow](https://github.com/paritytech/cumulus/pull/1576/files#diff-61ca6bac3ca9a62c3504e72696151a7371bffd3ba4edb9e46b4d5a7c757f34e2R1-R25) to run all the tests if anything in the ruled_labels folder changes.

That helps ensuring that:
- the specs are valid
- the test specs are valid
- all the tests pass

For now, this is using the latest published image under my account. We can swap to another image once available.
The workflow is fast and usually takes under 20s.

the-right-joyce:
I would not make those here mandatory:

```
 - name: Require a single issue status (S) label
    id: single_s
    tags: [ issue ]
    spec:
      when: !some_of  [ S* ]
      require: !one_of [ S* ]

  - name: Require a single Urgency status (U) label
    id: single_u
    tags: [ issue ]
    spec:
      when: !some_of  [ U* ]
      require: !one_of [ U* ]
```

If someone chooses to use S* and U* then we should check if they're using only one. But I'd not make them mandatory, as everyone has their own way of managing their issues in their projects and they should not be obligated to use the S* and U* labels if they don't want to/use it already through their boards.


```
 - name: Release need priorities
    id: require_one_p_when_b1
    tags: [ PR ]
    when: !one_of [ B1 ]
    spec:
      require: !one_of [ P* ]
```

We don't have priority labels.

```
 - name: Release need topics
    id: b_needs_t
    tags: [ PR ]
    when: !one_of [ B1 ]
    spec:
      require: !some_of [ T* ]
```

Also there I would not make mandatory yet. We said when someone adds B1 they need to add an X* label, bc those are shown in the release notes. The T* labels are used for human interaction only.

chevdor:
> I would not make those here mandatory:

They are not, look at the `when: !some_of  [ S* ]`

> We don't have priority labels.

We have the C* labels currently

gilescope:
replace bash with rust. Thumbs up.

chevdor:
bot rebase

chevdor:
bot rebase

chevdor:
This is now all working:
image

image

I had to disable the check for Priorities since those labels do not exist in the repo.

.github/workflows/check-labels-new.yml
```diff
@@ -15,12 +15,19 @@ jobs:
...
-          IMAGE: chevdor/ruled-labels:latest
-        run: |
-          docker pull $IMAGE
-          docker run --rm $IMAGE --version
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          override: true
+          components: ruled-labels
```
chevdor: I think this option will be much longer (we need to build and install ruled-labels) everytime, compared to using a (pre-built) docker image. Did you check the timing ?

chevdor: It may however still be fast is the cache works good. It would be good to check though since checking the labels ideally should be very fast.

alvicsam (approved):

the-right-joyce:
@ParityReleases please review

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2069394

sergejparity (approved):

the-right-joyce (approved):

coderobe (approved):

PR #1579, opened 2022-08-30
[Feature] Limit collectives teleports to DOT

ruseinov:
related to https://github.com/paritytech/cumulus/pull/1346#discussion_r941534026

### DIFF
--- a/pallets/xcm/src/lib.rs
+++ b/pallets/xcm/src/lib.rs
@@ -192,8 +192,8 @@ where
 	}
 }
 
-/// Accepts an asset if it is a native asset of a particular type.
-pub struct ConcreteNativeAsset<AssetId>(PhantomData<AssetId>);
+/// Accepts an asset if it is a native asset from a particular `MultiLocation`.
+pub struct ConcreteNativeAssetFrom<Location>(PhantomData<Location>);
 impl<AssetId: Get<MultiLocation>> FilterAssetLocation for ConcreteNativeAsset<AssetId> {
 	fn filter_asset_location(asset: &MultiAsset, origin: &MultiLocation) -> bool {
 		log::trace!(target: "xcm::filter_asset_location",
--- a/pallets/xcm/src/lib.rs
+++ b/pallets/xcm/src/lib.rs
@@ -24,15 +24,16 @@ use codec::{Decode, DecodeLimit, Encode};
 use cumulus_primitives_core::{
 	relay_chain::BlockNumber as RelayBlockNumber, DmpMessageHandler, ParaId,
 };
-use frame_support::dispatch::Weight;
+use frame_support::{dispatch::Weight, traits::Get};
 pub use pallet::*;
 use scale_info::TypeInfo;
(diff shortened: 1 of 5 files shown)


pallets/xcm/src/lib.rs
```diff
@@ -190,3 +191,14 @@ where
 		_ => Err(BadOrigin),
 	}
 }
+
+/// Accepts an asset if it is a native asset of a particular type.
+pub struct ConcreteNativeAsset<AssetId>(PhantomData<AssetId>);
```
KiChjang, line 196:
```suggestion
/// Accepts an asset if it is a native asset from a particular `MultiLocation`.
pub struct ConcreteNativeAssetFrom<Location>(PhantomData<AssetId>);
```

KiChjang: Also, what was the reason to put this struct inside of the XCM pallet?

ruseinov: For potential reuse across parachain runtimes, it also has dependencies on xcm, so that makes sense. As discussed with @joepetrowski

KiChjang: No it doesn't. `DenyThenTry` lives in `parachains/common/src/xcm_config.rs`; why isn't this struct also living there instead?

joepetrowski: Yeah, this should definitely go where Keith suggested, as part of the parachains' runtime config

ruseinov: Will fix that

ruseinov:
@joepetrowski @KiChjang all done

KiChjang (approved):

gilescope (approved):
That's better. Limited to just the parent's currency.

parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -140,7 +140,7 @@ impl xcm_executor::Config for XcmConfig {
 	// Collectives does not recognize a reserve location for any asset. Users must teleport DOT
 	// where allowed (e.g. with the Relay Chain).
 	type IsReserve = ();
-	type IsTeleporter = NativeAsset; // <- should be enough to allow teleportation of DOT
+	type IsTeleporter = ConcreteNativeAssetFrom<DotLocation>; // <- should be enough to allow teleportation of DOT
```
joepetrowski, line 143:
```suggestion
	// Only allow teleportation of DOT.
	type IsTeleporter = ConcreteNativeAssetFrom<DotLocation>;
```

chevdor, line 143: Let me suggest adding one more `/` so it shows up in the doc.

```suggestion
	/// Only allow teleportation of DOT.
	type IsTeleporter = ConcreteNativeAssetFrom<DotLocation>; 
```

chevdor (approved):

muharem (approved):

ruseinov:
bot merge

PR #1581, opened 2022-08-30
Companion for Weight v1.5

shawntabrizi:
Substrate: https://github.com/paritytech/substrate/pull/12138
Polkadot: https://github.com/paritytech/polkadot/pull/5943

### DIFF
--- a/pallets/collator-selection/src/weights.rs
+++ b/pallets/collator-selection/src/weights.rs
@@ -41,26 +41,37 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	fn set_invulnerables(b: u32) -> Weight {
 		Weight::from_ref_time(18_563_000 as RefTimeWeight)
 			// Standard Error: 0
-			.saturating_add((68_000 as RefTimeWeight).saturating_mul(b as RefTimeWeight))
+			.saturating_add(
+				Weight::from_ref_time(68_000 as RefTimeWeight)
+					.scalar_saturating_mul(b as RefTimeWeight),
+			)
 			.saturating_add(T::DbWeight::get().writes(1 as RefTimeWeight))
 	}
 	fn set_desired_candidates() -> Weight {
-		(16_363_000 as RefTimeWeight).saturating_add(T::DbWeight::get().writes(1 as RefTimeWeight))
+		Weight::from_ref_time(16_363_000 as RefTimeWeight)
+			.saturating_add(T::DbWeight::get().writes(1 as RefTimeWeight))
 	}
 	fn set_candidacy_bond() -> Weight {
-		(16_840_000 as RefTimeWeight).saturating_add(T::DbWeight::get().writes(1 as RefTimeWeight))
+		Weight::from_ref_time(16_840_000 as RefTimeWeight)
+			.saturating_add(T::DbWeight::get().writes(1 as RefTimeWeight))
 	}
 	fn register_as_candidate(c: u32) -> Weight {
 		Weight::from_ref_time(71_196_000 as RefTimeWeight)
 			// Standard Error: 0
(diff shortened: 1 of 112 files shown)


kianenigma (approved):

parachains/common/src/lib.rs
```diff
@@ -92,7 +92,7 @@ mod constants {
 	pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
 
 	/// We allow for 0.5 seconds of compute with a 6 second average block time.
-	pub const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND / 2;
+	pub const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND.scalar_saturating_mul(2);
```
joepetrowski, line 95: Why did you change divided by 2 to multiplied by 2?

joepetrowski: Until async backing is in, our limit is still half a second AFAIK.

test/runtime/src/lib.rs
```diff
@@ -138,7 +138,7 @@ const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
 /// by  Operational  extrinsics.
 const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(75);
 /// We allow for .5 seconds of compute with a 12 second average block time.
-const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND / 2;
+const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND.scalar_saturating_mul(2);
```
joepetrowski, line 141: ditto

joepetrowski (changes requested):

parachains/common/src/lib.rs
shawntabrizi: yeah bug, sorry

joepetrowski (approved):

pallets/collator-selection/src/weights.rs
```diff
@@ -39,91 +39,143 @@ pub trait WeightInfo {
...
-			.saturating_add(T::DbWeight::get().writes((2 as Weight).saturating_mul(c as Weight)))
+			.saturating_add(
+				Weight::from_ref_time(151_952_000 as RefTimeWeight)
+					.scalar_saturating_mul(c as RefTimeWeight),
+			)
+			.saturating_add(
+				T::DbWeight::get().reads((1 as RefTimeWeight).saturating_mul(r as RefTimeWeight)),
+			)
+			.saturating_add(
+				T::DbWeight::get().reads((2 as RefTimeWeight).saturating_mul(c as RefTimeWeight)),
```
ggwpez, line 99: We could think about a DLS for the weight annotations in the long run.  
This is starting to look really unreadable - at least for humans.

pallets/xcmp-queue/src/tests.rs
```diff
@@ -73,15 +83,20 @@ fn handle_invalid_data() {
...
+		);
 	});
 }
 
 #[test]
 fn service_overweight_unknown() {
 	new_test_ext().execute_with(|| {
 		assert_noop!(
-			XcmpQueue::service_overweight(Origin::root(), 0, 1000),
+			XcmpQueue::service_overweight(Origin::root(), 0, Weight::from_ref_time(1000)),
```
ggwpez, line 99: It would be better to use `WEIGHTS_PER_NANOS` in these cases instead of 1000. Then no refactoring will be required in the future.

ggwpez (approved):

pallets/collator-selection/src/weights.rs
shawntabrizi: i think making the weight file structure into a rust macro is not such a bad idea

PR #1585, opened 2022-08-31
Enable collation via RPC relay chain node

### LINKED ISSUE #989: Enable collation with external relay chain full node

skunert:
closes #989 

## Goal
This PR adds support for building blocks for a parachain without running an in-process relay chain node. Until now it was only possible to run a full node. The feature is still considered experimental, but is ready to test.
I recommend running the relay chain full node on the same machine as the collator.

**Usage:** `polkadot-parachain --alice --collator --relay-chain-rpc-url <rpc-websocket-url>`

## Internals
The parachain block production relies on some subsystems of polkadot. With this PR, we spawn a minimal polkadot node that has no runtime itself and receives chain related updates from the external full node via RPC. However, we spawn necessary subsystems for collation:
- Collation Generation Subsystem
- Runtime Api Subsystem
- Collator Protocol Subsystem
- Availability Recovery Subsystem
- Network Bridge Tx Subsystem
- Network Bridge Rx subsystem
- Availability Store
- Availability Distribution
- Chain API

The other subsystems are replaced by dummy subsystems that only drain channels of incoming messages. The subsystems that are running fetch their data off an instance of the newly added `BlockChainRpcClient`, which takes the place of the usual substrate client.


client/minimal-relay-chain-node/src/collator_overseer.rs
```diff
@@ -0,0 +1,246 @@
...
+			i = imports.next() => {
+				match i {
+					Some(header) => {
+		let block_info = BlockInfo { hash: header.hash(), parent_hash: header.parent_hash, number: header.number };
+						handle.block_imported(block_info).await;
+					}
+					None => break,
+				}
+			},
+			complete => break,
```
davxy, line 243:
```suggestion
```
Maybe this is not strictly required as the loop terminates as soon as one of the two futures yields `None`

client/minimal-relay-chain-node/src/collator_overseer.rs
```diff
@@ -0,0 +1,246 @@
...
+				// let forward = forward.fuse();
+				let overseer_fut = overseer.run().fuse();
+
+				pin_mut!(overseer_fut);
+				pin_mut!(forward);
+
+				select! {
+					_ = forward => (),
+					_ = overseer_fut => (),
+					complete => (),
```
davxy, line 203:
```suggestion
```
I don't think this branch is necessary, as the select terminates as soon as one of the two completes.

client/minimal-relay-chain-node/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+
+# cumulus deps
+cumulus-relay-chain-interface = { path = "../relay-chain-interface" }
+cumulus-relay-chain-rpc-interface = { path = "../relay-chain-rpc-interface" }
+cumulus-primitives-core = { path = "../../primitives/core" }
+
+lru = "0.7"
+tracing = "0.1.25"
+async-trait = "0.1.52"
+futures = { version = "0.3.1", features = ["compat"] }
```
davxy, line 45:
```suggestion
futures = "0.3.24"
```
1. Align to the version already used by the other crates (even though semver was already implicitly bumping it to 0.3.24)
2. Doesn't seem that we're using features from futures v0.1 (https://docs.rs/futures/0.3.24/futures/compat/index.html).

Note: second point will generate a different `Cargo.lock`, that will require to be updated as well

client/minimal-relay-chain-node/src/network.rs
```diff
@@ -0,0 +1,420 @@
...
+use crate::BlockChainRpcClient;
+
+pub(crate) struct BuildCollatorNetworkParams<'a> {
+	/// The service configuration.
+	pub config: &'a Configuration,
+	/// A shared client returned by `new_full_parts`.
+	pub client: Arc<BlockChainRpcClient>,
+	/// A handle for spawning tasks.
+	pub spawn_handle: SpawnTaskHandle,
+
```
davxy, line 42: I know that this is not exported out of crate and the meaning is super obvious. But the other entries have doc :-) Just to maintain the coherence

davxy (commented):
Overall LGTM.

My review mostly consists of a bunch of pedantic nitpicks :-D 

The most important part is probably the question wrt the `RelayChainRcInterface` construction.

Another question: "in process" relay chain will be removed in the future?
(I refer to this match branch: https://github.com/skunert/cumulus/blob/3c06eceb0050802a955f1478eb8f72a321de73c5/parachain-template/node/src/service.rs#L202)

Additionally maybe we should consider the idea to increment at least minor
versions to the crates that expose the additional features (e.g. : `parachain-template-node` and `polkadot-parachain`)

polkadot-parachain/src/cli.rs
```diff
@@ -83,7 +83,7 @@ pub struct Cli {
 	pub no_hardware_benchmarks: bool,
 
 	/// Relay chain arguments
-	#[clap(raw = true, conflicts_with = "relay-chain-rpc-url")]
+	#[clap(raw = true)]
```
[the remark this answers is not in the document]
skunert: This should not conflict anymore. The relay chain arguments that users pass influence the `polkadot_config` that goes into the minimal node. So we allow the user to specify extra options like bootnodes that should be passed to the minimal relay chain node.

skunert:
> Another question: "in process" relay chain will be removed in the future?

It will not be removed in the foreseeable future. It is the gold standard implementation and battle tested.

skunert:
Summary of recent changes to this PR:
- Add `ChainApiSubsystem`, `AvailabilityDistribution` and `AvailabilityStore` subsystems to enable pov-recovery
- Modify rust integration test and zombienet tests to have some RPC-connected nodes
- minor cleanup (log and format issues in pov-recovery)

skunert:
bot rebase

davxy (approved):
This is huge. And beside some small nit, looks very good!

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
```diff
@@ -0,0 +1,490 @@
...
+		self.rpc_client
+			.get_finalized_heads_stream()
+			.await
+			.expect("imported_headers_stream")
+			.boxed()
+	}
+}
+
+fn block_local<T>(fut: impl Future<Output = T>) -> T {
+	let tokio_handle = tokio::runtime::Handle::current();
```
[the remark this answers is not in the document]
skunert: I think there is not much to win here by using the `SpawnTaskHandle`. The task manager seems to be mostly designed towards longer running worker tasks. We would need to create channels for the return value of the rpc call since `spawn_blocking` returns `()`. Would be fine but the current solution is much simpler.

parachain-template/node/src/service.rs
```diff
@@ -178,7 +180,18 @@ async fn build_relay_chain_interface(
...
+				polkadot_config,
+				collator_pair.clone(),
+				Arc::new(BlockChainRpcClient::new(client.clone())),
+			)
+			.await?;
+			task_manager.add_child(collator_node.task_manager);
+			Ok((
+				Arc::new(RelayChainRpcInterface::new(client, collator_node.overseer_handle)),
+				Some(collator_pair),
+			))
```
bkchr, line 194: Probably worth to put all of this into an internal function as we have it for the in process relay chain?

test/service/src/cli.rs
```diff
@@ -41,7 +41,7 @@ pub struct TestCollatorCli {
 	pub parachain_id: u32,
 
 	/// Relay chain arguments
-	#[clap(raw = true, conflicts_with = "relay-chain-rpc-url")]
+	#[clap(raw = true)]
```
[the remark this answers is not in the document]
bkchr: Yeah or maybe this important for testing?

test/service/src/main.rs
```diff
@@ -78,7 +78,8 @@ fn main() -> Result<(), sc_cli::Error> {
 			})
 		},
 		None => {
-			let mut builder = sc_cli::LoggerBuilder::new("");
+			let log_filters = cli.run.normalize().log_filters();
+			let mut builder = sc_cli::LoggerBuilder::new(log_filters.unwrap_or(String::new()));
```
bkchr, line 82:
```suggestion
			let mut builder = sc_cli::LoggerBuilder::new(log_filters.unwrap_or_default());
```

zombienet_tests/0002-pov_recovery.toml
```diff
@@ -33,12 +33,30 @@ cumulus_based = true
...
+  #
+  # They will need to recover the pov blocks through availability recovery.
+  [[parachains.collators]]
+  name = "one"
+  validator = true
+  image = "{{COL_IMAGE}}"
+  command = "test-parachain"
+  args = ["--reserved-only", "--reserved-nodes {{'charlie'|zombie('multiAddress')}}", "-lcumulus-pov-recovery=debug", "--use-null-consensus", "--disable-block-announcements", "--relay-chain-rpc-url {{'alice'|zombie('wsUri')}}", "-- --bootnodes {{'alice'|zombie('multiAddress')}}"]
+
+  # run two as parachain full node that is only connected to dave
```
bkchr, line 56:
```suggestion
  # run two as parachain full node that is only connected to charlie
```

client/relay-chain-minimal-node/src/network.rs
```diff
@@ -0,0 +1,420 @@
...
+		_number: NumberFor<Block>,
+		_justifications: Justifications,
+	) {
+	}
+
+	/// Polls for actions to perform on the network.
+	///
+	/// This method should behave in a way similar to `Future::poll`. It can register the current
+	/// task and notify later when more actions are ready to be polled. To continue the comparison,
+	/// it is as if this method always returned `Poll::Pending`.
```
bkchr, line 70: No need to copy the docs here from the trait.

client/relay-chain-minimal-node/src/network.rs
```diff
@@ -0,0 +1,420 @@
...
+	pub config: &'a Configuration,
+	/// A shared client returned by `new_full_parts`.
+	pub client: Arc<BlockChainRpcClient>,
+	/// A handle for spawning tasks.
+	pub spawn_handle: SpawnTaskHandle,
+	/// Genesis hash
+	pub genesis_hash: Hash,
+}
+
+struct DummyImportQueue;
```
bkchr, line 46: Could also be moved down to the end of the file to have a clean separation between the dummy stuff and the real implementation.

client/relay-chain-minimal-node/src/network.rs
```diff
@@ -0,0 +1,420 @@
...
+		futures::select! {
+			// Answer incoming RPC requests.
+
+			// The network worker has done something. Nothing special to do, but could be
+			// used in the future to perform actions in response of things that happened on
+			// the network.
+			_ = (&mut network).fuse() => {}
+		}
+	}
+}
```
bkchr, line 175:
```suggestion
```

The `NetworkWorker` is already a future and a future "fires once" and then you should not poll it anymore.

client/relay-chain-minimal-node/src/network.rs
```diff
@@ -0,0 +1,420 @@
...
+		state_request_protocol_config,
+		warp_sync_protocol_config: None,
+		light_client_request_protocol_config,
+		request_response_protocol_configs: Vec::new(),
+	};
+
+	let network_mut = sc_network::NetworkWorker::new(network_params)?;
+	let network = network_mut.service().clone();
+
+	let future = build_network_collator_future(network_mut);
```
bkchr, line 130:
```suggestion
	let network = sc_network::NetworkWorker::new(network_params)?;
	let network_service = network_mut.service().clone();
```

client/relay-chain-minimal-node/src/network.rs
```diff
@@ -0,0 +1,420 @@
...
+			Some(Box::new(move |fut| {
+				spawn_handle.spawn("libp2p-node", Some("networking"), fut);
+			}))
+		},
+		transactions_handler_executor: {
+			let spawn_handle = Clone::clone(&spawn_handle);
+			Box::new(move |fut| {
+				spawn_handle.spawn("network-transactions-handler", Some("networking"), fut);
+			})
+		},
```
bkchr, line 111: With the latest upstream changes this shouldn't be required anymore.

client/relay-chain-minimal-node/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+		let registry = config.prometheus_config.as_ref().map(|cfg| &cfg.registry);
+		TaskManager::new(config.tokio_handle.clone(), registry)?
+	};
+
+	let prometheus_registry = config.prometheus_registry().cloned();
+
+	let genesis_hash = relay_chain_rpc_client
+		.block_get_hash(Some(0))
+		.await
+		.expect("Crash here if no genesis is available")
```
bkchr, line 116:
```suggestion
		.expect("Genesis block hash is always available; qed")
```

client/relay-chain-minimal-node/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+/// - RuntimeApi
+#[sc_tracing::logging::prefix_logs_with("Relaychain")]
+pub async fn new_minimal_relay_chain(
+	mut config: Configuration,
+	collator_pair: CollatorPair,
+	relay_chain_rpc_client: Arc<BlockChainRpcClient>,
+) -> Result<NewMinimalNode, RelayChainError> {
+	let role = config.role.clone();
+
+	// Use the given RPC node as bootnode, since we do not have a chain spec with valid boot nodes
```
bkchr, line 102: I'm okay with this. However, we should probably somewhere up in the stack returning an error if someone passes arguments to the relay chain node, which we consume, but ignore.

client/relay-chain-minimal-node/src/collator_overseer.rs
```diff
@@ -0,0 +1,278 @@
...
+	pub available_data_req_receiver: IncomingRequestReceiver<AvailableDataFetchingRequest>,
+	/// Prometheus registry, commonly used for production systems, less so for test.
+	pub registry: Option<&'a Registry>,
+	/// Task spawner to be used throughout the overseer and the APIs it provides.
+	pub spawner: sc_service::SpawnTaskHandle,
+	/// Determines the behavior of the collator.
+	pub collator_pair: CollatorPair,
+	/// Request response protocols
+	pub req_protocol_names: ReqProtocolNames,
+	pub peer_set_protocol_names: PeerSetProtocolNames,
```
bkchr, line 84: Missing docs

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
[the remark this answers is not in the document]
bkchr: I don't think that we need to overcomplicate this here, as @skunert said.

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
```diff
@@ -0,0 +1,490 @@
...
+	async fn dmq_contents(
+		&self,
+		at: Hash,
+		recipient: cumulus_primitives_core::ParaId,
+	) -> Result<
+		Vec<cumulus_primitives_core::InboundDownwardMessage<polkadot_core_primitives::BlockNumber>>,
+		sp_api::ApiError,
+	> {
+		let result = self.rpc_client.parachain_host_dmq_contents(recipient, at).await?;
+		Ok(result)
```
bkchr, line 224:
```suggestion
		Ok(self.rpc_client.parachain_host_dmq_contents(recipient, at).await?)
```

Stylistic nit :P Could be done here everywhere.

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
```diff
@@ -0,0 +1,490 @@
...
+			.boxed()
+	}
+
+	pub async fn finality_notification_stream_async(
+		&self,
+	) -> Pin<Box<dyn Stream<Item = Header> + Send>> {
+		self.rpc_client
+			.get_finalized_heads_stream()
+			.await
+			.expect("imported_headers_stream")
```
bkchr, line 400: Why can they not fail? Better expects please

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
```diff
@@ -0,0 +1,490 @@
...
+	fn info(&self) -> Info<Block> {
+		let best_header = block_local(self.rpc_client.chain_get_header(None))
+			.expect("get_header")
+			.unwrap();
+		let genesis_hash = block_local(self.rpc_client.chain_get_head(Some(0))).expect("get_head");
+		let finalized_head =
+			block_local(self.rpc_client.chain_get_finalized_head()).expect("get_head");
+		let finalized_header = block_local(self.rpc_client.chain_get_header(Some(finalized_head)))
+			.expect("get_head")
+			.unwrap();
```
bkchr, line 438: No unwraps and better expect messages please.

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
```diff
@@ -0,0 +1,490 @@
...
+				} else {
+					false
+				}
+			},
+		};
+
+		match exists {
+			true => Ok(sc_client_api::blockchain::BlockStatus::InChain),
+			false => Ok(sc_client_api::blockchain::BlockStatus::Unknown),
+		}
```
bkchr, line 470:
```suggestion
		if exists {
			Ok(sc_client_api::blockchain::BlockStatus::InChain) } else {
			Ok(sc_client_api::blockchain::BlockStatus::Unknown)
		}
```

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
```diff
@@ -0,0 +1,490 @@
...
+			.boxed()
+	}
+}
+
+fn block_local<T>(fut: impl Future<Output = T>) -> T {
+	let tokio_handle = tokio::runtime::Handle::current();
+	tokio::task::block_in_place(|| tokio_handle.block_on(fut))
+}
+
+impl HeaderBackend<Block> for BlockChainRpcClient {
```
bkchr, line 410: Is any of these functions ever called?

bkchr: If yes, where again?

bkchr (approved):
I like!

Overall a very simple implementation, based on all the preparations you have done :)

I mainly left some nitpicks here and there, but nothing serious ;)

client/relay-chain-minimal-node/src/network.rs
```diff
@@ -0,0 +1,420 @@
...
+
+	let (network_start_tx, network_start_rx) = futures::channel::oneshot::channel();
+
+	// The network worker is responsible for gathering all network messages and processing
+	// them. This is quite a heavy task, and at the time of the writing of this comment it
+	// frequently happens that this future takes several seconds or in some situations
+	// even more than a minute until it has processed its entire queue. This is clearly an
+	// issue, and ideally we would like to fix the network future to take as little time as
+	// possible, but we also take the extra harm-prevention measure to execute the networking
+	// future using `spawn_blocking`.
```
koute, line 140: So the processing of the queue can take more than a minute, and it spends most of this time *actually* burning CPU time instead of waiting? If so that does sound quite abnormal to me.

zombienet_tests/0002-pov_recovery.toml
```diff
@@ -33,12 +33,30 @@ cumulus_based = true
...
+  [[parachains.collators]]
+  name = "two"
+  validator = false
+  image = "{{COL_IMAGE}}"
+  command = "test-parachain"
+  args = ["--reserved-only", "--reserved-nodes {{'charlie'|zombie('multiAddress')}}", "-lcumulus-pov-recovery=debug","--use-null-consensus", "--disable-block-announcements", "--relay-chain-rpc-url {{'bob'|zombie('wsUri')}}"]
```
koute, line 62: Wouldn't it be slightly more real-world if these parachain nodes would connect to non-validator relay nodes as their relay chain RPC nodes?

koute (approved):
Looks mostly good to me!

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
bkchr: Haha. 1000 ways to obfuscate your if :D

client/relay-chain-minimal-node/src/collator_overseer.rs
```diff
@@ -0,0 +1,278 @@
...
+			availability_config,
+			registry,
+			spawner,
+			collator_pair,
+			req_protocol_names,
+			peer_set_protocol_names,
+			parachains_db,
+			pov_req_receiver,
+			chunk_req_receiver,
+		}: CollatorOverseerGenArgs<'a>,
```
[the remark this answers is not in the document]
bkchr: I have done this multiple times here in the code base. In the end it is just to name the parameters. 

Though, I never checked how the signature looks like in rust docs, but I would assume that they don't show the unpacking there.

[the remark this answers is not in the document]
koute: Well, it's fine by me either way. (:

client/relay-chain-minimal-node/src/network.rs
skunert: I copied this part of the code from [here](https://github.com/paritytech/substrate/blob/61b9a4d1a8a9bf39c1d89a8dd02f82785c10860c/client/service/src/builder.rs#L931), I did not investigate this specific issue. Could very well be that this is outdated.

client/relay-chain-minimal-node/src/lib.rs
skunert: Since I am passing the config into the polkadot code, it is a bit hard to analyze which fields exactly are used and which are unused. I added a warning message to the cli that prints a message when users provide `--relay-chain-rpc-url` together with raw relay chain arguments.

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
skunert: I removed some of the expect messages here, some are left however since `HeaderBackend` methods are not expected to fail currently and we can not handle request errors anyway in this version.

client/relay-chain-minimal-node/src/blockchain_rpc_client.rs
skunert: Mostly in protocol and authority-discovery. Not all are called however, but in this case it was easy enough to implement them. If you prefer I can make  the uncalled ones `unimplemented!`.

**network protocol**
https://github.com/paritytech/substrate/blob/61b9a4d1a8a9bf39c1d89a8dd02f82785c10860c/client/network/src/protocol.rs#L278
https://github.com/paritytech/substrate/blob/61b9a4d1a8a9bf39c1d89a8dd02f82785c10860c/client/network/src/protocol.rs#L370
https://github.com/paritytech/substrate/blob/61b9a4d1a8a9bf39c1d89a8dd02f82785c10860c/client/network/src/protocol.rs#L740
https://github.com/paritytech/substrate/blob/61b9a4d1a8a9bf39c1d89a8dd02f82785c10860c/client/network/src/protocol.rs#L844

**authority-discovery**
https://github.com/paritytech/substrate/blob/61b9a4d1a8a9bf39c1d89a8dd02f82785c10860c/client/authority-discovery/src/worker.rs#L381
https://github.com/paritytech/substrate/blob/61b9a4d1a8a9bf39c1d89a8dd02f82785c10860c/client/authority-discovery/src/worker.rs#L601

client/relay-chain-minimal-node/src/network.rs
koute: I see. Okay, then you can leave it as-is.

From what I can see @tomaka added this over two years ago, so it's probably worth revisit.

tomaka: I'm not following networking code changes, but given the lack of huge changes I think that this is still accurate.
I don't know what you have in mind when you say "burning" CPU, but yes we're talking about several seconds of CPU cycles to run `poll()` once.

tomaka:
I just wanted to mention as a heads up that you should be able to relatively easily embed smoldot:

https://docs.rs/smoldot-light/latest/smoldot_light/
https://github.com/paritytech/smoldot/blob/main/bin/light-base/examples/basic.rs

This doesn't need to be done over a socket (as it feels a bit stupid to open a socket between two parts of the same program) but simply by calling functions of smoldot.
The API is only about sending/receiving JSON-RPC requests/responses/notifications, and the actual "meaningful" communication is only done in the context of the JSON-RPC API.

skunert:
@paritytech/ci I think this is ready for review. Can someone take a look?

skunert:
I reverted the pov-recovery test back to the state on master. It is flaky and there seem to be issues with it that are unrelated to this PR, should be a separate investigation. Issue: https://github.com/paritytech/cumulus/issues/1744

Polkadot-Forum:

https://forum.polkadot.network/t/polkadot-release-analysis/1026/2

PR #1601, opened 2022-09-06
Companion for paritytech/substrate#12183

KiChjang:
Companion for paritytech/substrate#12183


pallets/xcmp-queue/src/lib.rs
```diff
@@ -831,7 +831,7 @@ impl<T: Config> Pallet<T> {
 				if shuffle_index < status.len() {
 					weight_available +=
 						(max_weight - weight_available) / (weight_restrict_decay.ref_time() + 1);
-					if weight_available + threshold_weight > max_weight {
+					if (weight_available + threshold_weight).all_gt(max_weight) {
```
shawntabrizi, line 834: first glance, this should be `any_gt`?

Usually it is "any" when comparing against a max weight

shawntabrizi (approved):
consider my comment, but assume this will be merged with correct logic

pallets/xcmp-queue/src/lib.rs
KiChjang: Ah, I might have confused myself about what's on the LHS and what's on the RHS, lemme double check.

KiChjang: Ok, I see what's happening here, the statement following this conditional attempts to reassign `weight_available` to `max_weight`, and so my thought process was that the reassignment should only happen when `(weight_available + threshold_weight)` is _strictly greater_ than `max_weight`, otherwise we may end up with a `weight_available` that has a component lower than the previous `max_weight`.

KiChjang: Hmm okay, I get the logic here, if the LHS ever gets larger than the RHS, then we cap the available weight at the max, so yes, it should be `any_gt`.

```suggestion
					if (weight_available + threshold_weight).any_gt(max_weight) {
```

andresilva (approved):

PR #1636, opened 2022-09-18
updated integration tests

gilescope:
Integration test changes:

- xcm weights are no longer all fixed (can we assert they're in a range? https://github.com/paritytech/parachains-integration-tests/issues/35 )
- default port changes so that polkadot is not on the same ports as kusama. (at some point we're going to be running both at once doing integration tests between them and if they're on the same ports we can't do that so let's get into good conventions early on.)
- typo fixes.
- change from `*amount` to `1,000,000,000,000` as it was expecting the commas when it checked the assertion now. (Can / should we make it less picky @NachoPal ? )

(I can break these up into smaller PRs on request)

### DIFF
--- a/parachains/integration-tests/statemine/config.toml
+++ b/parachains/integration-tests/statemine/config.toml
@@ -5,7 +5,7 @@ chain = "kusama-local"
 
 	[[relaychain.nodes]]
 	name = "alice"
-	ws_port = 9900
+	ws_port = 8800
 	validator = true
 
 	[[relaychain.nodes]]
@@ -27,7 +27,7 @@ cumulus_based = true
 
 	[[parachains.collators]]
 	name = "collator1"
-	ws_port = 9910
+	ws_port = 8810
 	command = "./bin/polkadot-parachain"
 
 	[[parachains.collators]]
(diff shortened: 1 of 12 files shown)


gilescope:
u64::MAX should not come up in this PR. That should already be in the release branch. Will require investigation.

NachoPal:
> xcm weights are no longer all fixed (can we assert they're in a range? https://github.com/paritytech/parachains-integration-tests/issues/35 )

Instead of commenting the expected weight values, can you add the weights values you are getting? I get your point of adding a range, but right now XCM weights is not something that changes for every release. If you take a look at Statemine, the XCM values where added there some releases ago, and never changed since then.

I think for now (until ranges are in place) it is more useful if we add the values instead of commenting them out. If we silent them we might miss potential issues during the next releases.

NachoPal:
> change from *amount to 1,000,000,000,000 as it was expecting the commas when it checked the assertion now. (Can

I think it was fixed, maybe you are not using the last version. There is a published version here: https://www.npmjs.com/package/@parity/parachains-integration-tests

NachoPal:
Also, do you mind creating a new `release-v9290-v0.9.29` branch in https://github.com/paritytech/parachains-integration-tests with the changes? I usually run the tests there in a new release branch and when happy I open a PR here with the changes. As last step, I merge the new release branch into `master` to update it to the last version.

gilescope:
Cool. will do that. I think the xcm weights will be re-run every release from now on (aside from the hardcoded ones).

gilescope:
Will open again against the parachain-integration-tests repo.

PR #1640, opened 2022-09-19
Revert "[Fix] Weight calculations for Wild -> affects teleports"

joepetrowski:
Reverts paritytech/cumulus#1639

Based on https://github.com/paritytech/cumulus/pull/1639#discussion_r974569345


KiChjang (approved):

ruseinov:
@KiChjang I also assumed that XCM pallet needs to be modified to use definite, it was my temp solution for overpaying.
Good to know I was right about the former. I wasn’t actually sure what we should do with that coefficient 
I’d like to make this change in XCM pallet.

ruseinov (approved):

ruseinov:
bot merge

PR #1643, opened 2022-09-20
set xcm weights template

gilescope:
Xcm benchmarks use a separate template. (We should have been calling this)

### DIFF
--- a/scripts/benchmarks-ci.sh
+++ b/scripts/benchmarks-ci.sh
@@ -22,27 +22,52 @@ if [[ $runtimeName == "statemint" ]] || [[ $runtimeName == "statemine" ]] || [[ 
 		pallet_uniques
 		cumulus_pallet_xcmp_queue
 		frame_system
+	)
+	xcm_pallets=(
 		pallet_xcm_benchmarks::generic
 		pallet_xcm_benchmarks::fungible
 	)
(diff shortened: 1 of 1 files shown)


scripts/benchmarks-ci.sh
```diff
@@ -22,28 +22,53 @@ if [[ $runtimeName == "statemint" ]] || [[ $runtimeName == "statemine" ]] || [[
...
 	exit 1
 fi
 
+for pallet in ${xcm_pallets[@]}
+do
+	# a little hack for xcm benchmarks
+	output_file="${pallet//::/_}"
+  if [[ "$pallet" == *"xcm"* ]]; then
+		output_file="xcm/$output_file"
+  fi
```
muharem, line 57: this probably not necessary, they all xcm.
and the loop below also can be simplified

muharem (approved):

scripts/benchmarks-ci.sh
gilescope: actually let's lean into that and use that for setting the template.

ruseinov (approved):

ruseinov:
Looks good, maybe a manual run of just one bench without repetitions to see that it writes to a correct place and with correct format?

PR #1654, opened 2022-09-21
Ensure that inherents are first and unsigned

bkchr:


pallets/parachain-system/src/validate_block/tests.rs
```diff
@@ -251,3 +251,40 @@ fn check_inherent_fails_on_validate_block_as_expected() {
...
+				"check_inherents_are_unsigned_and_before_all_other_extrinsics",
+				"--",
+				"--nocapture",
+			])
+			.env("RUN_TEST", "1")
+			.output()
+			.expect("Runs the test");
+		assert!(output.status.success());
+
+		assert!(dbg!(String::from_utf8(output.stderr).unwrap())
```
pepyakin, line 287:
```suggestion
		assert!(String::from_utf8(output.stderr).unwrap()
```

pepyakin (approved):

kianenigma (approved):

BustaNit (approved):

pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -118,6 +118,11 @@ where
 	let inherent_data = block
 		.extrinsics()
 		.iter()
+		// Inherents are at the front of the block and are unsigned.
+		//
+		// If `is_signed` is returning `None`, we keep it safe and assume that it is "signed".
+		// We are searching for unsigned transactions anyway.
+		.take_while(|e| !e.is_signed().unwrap_or(true))
```
mustermeiszer, line 125: @bkchr just to clarify if I get the comments correctly. 

This `take_while(..)` will include the extrinsic if `is_signed()` returns `None`, while the docs say to keep it safe and assumes the extrinsic is signed, which I understand as rejecting the extrinsic at that point instead of including it. Hence, I am confused.

bkchr: There is a `!` at the front. So, we only `take` extrinsics until we find the first signed one.

mustermeiszer: Thanks, missed that!

Polkadot-Forum:

https://forum.polkadot.network/t/security-vulnerability-executing-unauthorized-xcm-messages/553/1

PR #1658, opened 2022-09-21
Fix failing weights jobs

EgorPopelyaev:
This small fix adds current time in millis to the name of the branch with weights to make it uniq and avoid issue when the re-ran job fails because the branch with the same name already exist


gilescope (approved):
LGTM

muharem (approved):

Issue #533, opened 2021-07-12
Weights for parachain-system


xlc:

bkchr:
@shawntabrizi anyone that could work on this? 

Especially https://github.com/paritytech/cumulus/blob/ad5dd362b451ce2f31a862f2f118bb92984ebebc/pallets/parachain-system/src/lib.rs#L291

bkchr:
@ggwpez could you look into this please?

ggwpez:
If `set_validation_data` is called from inherents, then it it part of the `BlockExecutionWeight`. But benchmarks should be added nonetheless. I will take a look.

PR #1709, opened 2022-09-30
Benchmark `pallet-parachain-system`

ggwpez:
WIP Closes https://github.com/paritytech/cumulus/issues/533


bkchr:
> * Generating proofs is `std`-only and cannot happen within the benchmarks.

You can also generate them in `no_std`. `sp_trie::generate_trie_proof` for example :P We probably need to generate the proof for the benchmarking, otherwise the number of messages is sort of hard coded?  

> * Must the `AbridgedHostConfiguration` be injected by the concrete runtime or are constants fine? Note that these must be compile-time available for benchmarking. The constants are currently taken from the `cumulus-test-relay-sproof-builder`.

We should probably find out the impact of different `AbridgedHostConfiguration` on `on_finalize`. I mean it looks a little bit complicated, as the inherent sets the value and that influences how the weight of `on_finalize` that we normally should already account for in `on_initialize`. We could probably cheat and make the inherent use more weight, so that the weight we could not account for in `on_initialize` is accounted in the inherent. Writing this makes me realize how bad the current approach is :P The host configuration can any way only change one per session. We should probably change the inherent to take this into account. Aka only when the configuration will change, we will put it into the inherent and then it would be the last thing of `on_finalize` where we write the new config to the state.

nbaztec:
Since it's been long enough perhaps we can merge this PR with hard coded proofs for now? It'll be better than not having any weight on these extrinsics for the time being.

PR #1710, opened 2022-09-30
XCMv3 fixes

bkontur:


parachain-template/node/src/service.rs
```diff
@@ -488,13 +487,14 @@ pub async fn start_parachain_node(
 					create_inherent_data_providers: move |_, (relay_parent, validation_data)| {
 						let relay_chain_interface = relay_chain_interface.clone();
 						async move {
-							let parachain_inherent =
```
bkontur, line 491: @KiChjang
`parachain_inherent` is not used anywhere, so I just temporary commented out it,
maybe it could be removed, I dont know

parachain-template/runtime/src/xcm_config.rs
```diff
@@ -180,7 +180,7 @@ impl xcm_executor::Config for XcmConfig {
 	type AssetTrap = PolkadotXcm;
 	type AssetClaims = PolkadotXcm;
 	type SubscriptionService = PolkadotXcm;
-	type PalletInstancesInfo = ();
+	type PalletInstancesInfo = super::AllPalletsWithSystem;
```
KiChjang, line 183: Import this at the very top of the file instead:

```
use super::AllPalletsWithSystem;
```

bkontur: fixed

KiChjang (approved):

PR #1712, opened 2022-09-30
Westend Collectives chain spec

muharem:
The chain spec based on previous Westend Collectives chain spec, the only changes are:
- parachain id changed from 1002 to 1001 to align with future Polkadot Collectives 
- latest (v9290) runtime wasm binary 
- nonraw spec file (human readable) format


gilescope (approved):
Looks good. Start and end bytes match what I was expecting of the runtime.

joepetrowski (approved):

muharem:
Bot merge

bkchr:
@muharem the chain spec you added is a not a raw chain spec.

muharem:
> @muharem the chain spec you added is a not a raw chain spec.

yes, the idea was to make it human readable.
should I upload the raw one? can you explain please why we wanna store a raw one?

bkchr:
The human readable chain spec requires the native runtime to be loaded. When you change the native runtime the chain spec can not be loaded anymore. Using the human readable chain spec on the same commit is fine, but the moment something changes it starts to fail to load the chain spec (which is the case on master).

muharem:
> The human readable chain spec requires the native runtime to be loaded. When you change the native runtime the chain spec can not be loaded anymore. Using the human readable chain spec on the same commit is fine, but the moment something changes it starts to fail to load the chain spec (which is the case on master).

https://github.com/paritytech/cumulus/pull/2131

PR #1716, opened 2022-10-03
rename polkadot-parachain package to avoid name clash

gilescope:
There's a polkdot-parachain crate in polkadot. We can rename the bin crate to be a different package while keeping the bin name the same.

This tides up the lock file so that it does not have to specify a version number to distinguish between the two.


ruseinov (approved):

gilescope:
Please wait till Chevor gives the nod before we merge this.

chevdor (approved):
I checked in our `rel` scripts and I don't think we use this value so we should be fine with that change.

bkchr:
bot rebase

bkchr:
bot rebase

bkchr (approved):

bkchr:
bot merge

PR #1722, opened 2022-10-03
Migrate Weights properly to v2

KiChjang:
Since migration to weights v2, we need to ensure that any pallets that store `Weight` structs or contain an extrinsic with a `Weight` parameter to migrate properly.

### DIFF
--- a/pallets/dmp-queue/src/migration.rs
+++ b/pallets/dmp-queue/src/migration.rs
@@ -17,7 +17,11 @@
 //! A module that is responsible for migration of storage.
 
 use crate::{Config, Pallet, Store};
-use frame_support::{pallet_prelude::*, traits::StorageVersion, weights::{constants::WEIGHT_PER_MILLIS, Weight}};
+use frame_support::{
+	pallet_prelude::*,
+	traits::StorageVersion,
+	weights::{constants::WEIGHT_PER_MILLIS, Weight},
+};
 use xcm::latest::Weight as XcmWeight;
 
 /// The current storage version.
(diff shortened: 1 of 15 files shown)


pallets/dmp-queue/src/migration.rs
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+    }
+}
+
+/// Migrates `QueueConfigData` from v1 (using only reference time weights) to v2 (with
+/// 2D weights).
+///
+/// NOTE: Only use this function if you know what you're doing. Default to using
+/// `migrate_to_latest`.
```
sam0x17, line 61: maybe an additional sentence about the scenario where one would _want_ to call this just as a sanity check for people who know what they're doing ;)

KiChjang: Honestly, the situation where I think it would make sense to call `migrate_to_v1` directly is during unit testing, otherwise I really don't see why anyone would call this method on its own.

pallets/xcmp-queue/src/migration.rs
```diff
@@ -54,37 +56,35 @@ mod v0 {
...
-/// Uses the `Default` implementation of `QueueConfigData` to choose a value for
-/// `xcmp_max_individual_weight`.
+/// Migrates `QueueConfigData` from v1 (using only reference time weights) to v2 (with
+/// 2D weights).
 ///
 /// NOTE: Only use this function if you know what you're doing. Default to using
 /// `migrate_to_latest`.
-pub fn migrate_to_v1<T: Config>() -> Weight {
-	let translate = |pre: v0::QueueConfigData| -> super::QueueConfigData {
+pub fn migrate_to_v2<T: Config>() -> Weight {
```
ruseinov, line 72: nit: why is this not in the module to be consistent with other migration definitions?

ruseinov (approved):
seems good to me

ggwpez (approved):
Should work but it would be a bit easier for testing if it was in a `OnRuntimeUpgrade`.

KiChjang:
bot merge

crystalin:
@bkchr This includes a runtime upgrade it seems.
It shouldn't be considered `D1-low` and should have the migration tag !
:)

bkchr:
@crystalin you are right, but we pinging me and not @KiChjang who created and labeled the pr? 😉

PR #1730, opened 2022-10-05
Add protoc

chevdor:
This PR adds protoc for the rustdoc workflow

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - name: Install tooling
         run: |
-          apt-get install -y protobuf
+          apt-get install -y protobuf-compiler
           protoc --version
 
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -11,6 +11,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - name: Install tooling
+        run: |
+          apt-get install -y protobuf
```
alvicsam, line 16:
```suggestion
          apt-get install -y protobuf-compiler
```

I'm not sure but the [official doc](https://grpc.io/docs/protoc-installation/) says that the package has a slightly different name

chevdor: Totally, this is also what I did in https://github.com/paritytech/srtool/pull/50/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R35 🤦 and confirmed fixing the issues.

alvicsam (approved):

EgorPopelyaev (approved):

bkontur (approved):

PR #1562, opened 2022-08-22
[ci] add auto cargo-fmt

alvicsam:
The PR adds `cargo fmt` check in the beginning of the pipeline. If the check fails, the code will be formatted automatically and the pipeline is going to be restarted.

cc https://github.com/paritytech/ci_cd/issues/462

PR #1731, opened 2022-10-05
fmt

chevdor:
Fix fmt issues


bkchr (approved):

bkchr:
What happened to the formatting check action?

ggwpez (approved):

chevdor:
The fmt check was removed from GH a while back via https://github.com/paritytech/cumulus/pull/1562/files and added to Gitlab. @alvicsam it looks like there may be an issue in the Gitlab CI: it does not pickup fmt issues.

chevdor:
I opened an issue to track the CI problem: https://github.com/paritytech/cumulus/issues/1732

alvicsam:
It's my fault, I removed auto-fmt because of the forum post and forgot to return the GHA back

Issue #1732, opened 2022-10-05
Rustfmt checks are no longer running


chevdor:
The fmt check was removed from GH a while back via https://github.com/paritytech/cumulus/pull/1562/files and added to Gitlab. 

@alvicsam it looks like there may be an issue in the Gitlab CI: it does not pickup fmt issues.

PR #1733, opened 2022-10-05
[ci] add fmt check

alvicsam:
Close https://github.com/paritytech/cumulus/issues/1732

### DIFF
--- a/.github/workflows/fmt-check.yml
+++ b/.github/workflows/fmt-check.yml
@@ -1,0 +1,40 @@
+name: Quick check Formatting
+
+on:
+  push:
+    branches:
+      - "*"
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+
+jobs:
+  quick_check:
+    strategy:
+      matrix:
+        os: ["ubuntu-latest"]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Install Rust nightly toolchain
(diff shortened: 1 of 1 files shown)


.github/workflows/fmt-check.yml
```diff
@@ -0,0 +1,40 @@
+name: Quick check Formatting
```
chevdor, line 1: I would suggest aligning the `name` and the filename.
This workflow used to be called `quick-check.yml`.

chevdor (changes requested):

.github/workflows/fmt-check.yml
alvicsam: But it's not clear from the name "quick check" what the action does. Maybe `quick-fmt-check` then?

radupopa2010 (approved):

.github/workflows/fmt-check.yml
```diff
@@ -0,0 +1,40 @@
+name: Rustfmt check
+
+on:
+  push:
+    branches:
+      - "*"
```
sergejparity, line 6: Do we really want to trigger GHA on every push to every branch?

sergejparity (approved):

chevdor (approved):

chevdor:
bot merge

coderobe (approved):

PR #1738, opened 2022-10-06
Dependency not needed

gilescope:
When building the `substrate-parachain-template`, the `polkadot-cli` dependency contributed 1 min build time on my machine but isn't actually needed. In that workspace it was only used by this crate so hopefully this will knock a cool 1 min of the build time on my fast machine.
(It won't help cumulus repo build times as 3-4 other things in this repo do depend on `polkadot-cli` - I'm assuming they do so for good reason...)

Before one after the other:

image

After both compiling at once:

image

### DIFF
--- a/client/relay-chain-inprocess-interface/src/lib.rs
+++ b/client/relay-chain-inprocess-interface/src/lib.rs
@@ -31,7 +31,6 @@ use polkadot_client::{ClientHandle, ExecuteWithClient, FullBackend};
 use polkadot_service::{
 	AuxStore, BabeApi, CollatorPair, Configuration, Handle, NewFull, TaskManager,
 };
-use sc_cli::SubstrateCli;
 use sc_client_api::{
 	blockchain::BlockStatus, Backend, BlockchainEvents, HeaderBackend, ImportNotifications,
 	StorageProof, UsageProvider,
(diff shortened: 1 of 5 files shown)


NukeManDan:
Bravo. And a greener network to boot with less energy wasted on unneeded compilation on _thousands_ of parachain builders machines.

client/relay-chain-inprocess-interface/src/lib.rs
```diff
@@ -363,8 +362,8 @@ pub fn build_inprocess_relay_chain(
 ) -> RelayChainResult<(Arc<(dyn RelayChainInterface + 'static)>, Option<CollatorPair>)> {
 	// This is essentially a hack, but we want to ensure that we send the correct node version
```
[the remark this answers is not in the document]
gilescope: Good point. Updated.

gilescope:
Hang on there Dan, it's still compiling all the same things. But given that the machine gets done faster you're right, it probably is a bit more efficient.

muharem (approved):

bkontur (approved):
lgtm

skunert (approved):

client/relay-chain-inprocess-interface/src/lib.rs
```diff
@@ -361,11 +360,6 @@ pub fn build_inprocess_relay_chain(
 	task_manager: &mut TaskManager,
 	hwbench: Option<sc_sysinfo::HwBench>,
 ) -> RelayChainResult<(Arc<(dyn RelayChainInterface + 'static)>, Option<CollatorPair>)> {
-	// This is essentially a hack, but we want to ensure that we send the correct node version
-	// to the telemetry.
-	polkadot_config.impl_version = polkadot_cli::Cli::impl_version();
-	polkadot_config.impl_name = polkadot_cli::Cli::impl_name();
```
bkchr, line 367: Did you read these comments? This wasn't done without any reason.

bkchr (changes requested):

client/relay-chain-inprocess-interface/src/lib.rs
gilescope: Need the relay-chain-inprocess-interface to expose its version number. That can be done as a separate polkadot PR.

PR #1743, opened 2022-10-07
Statemine state migration runtime changes.

cheme:
This PR contains runtime change to switch state mine to a hybrid trie state.

Note that the accounts need to be change to the rights KEY_ROOT_MIG_CONTROLLER and KEY_MIG_CONTROLLER that will be responsible for the migration. CC @PierreBesson

### DIFF
--- a/parachains/runtimes/assets/statemine/Cargo.toml
+++ b/parachains/runtimes/assets/statemine/Cargo.toml
@@ -44,6 +44,7 @@ sp-session = { git = "https://github.com/paritytech/substrate", default-features
 sp-std = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 sp-transaction-pool = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 sp-version = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
+pallet-state-trie-migration = { git = "https://github.com/paritytech/substrate", default-features = false, branch = "master" }
 
 # Polkadot
 kusama-runtime-constants = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "master" }
@@ -143,6 +144,7 @@ std = [
 	"pallet-multisig/std",
 	"pallet-proxy/std",
 	"pallet-session/std",
+	"pallet-state-trie-migration/std",
 	"pallet-timestamp/std",
 	"pallet-transaction-payment-rpc-runtime-api/std",
 	"pallet-transaction-payment/std",
--- a/parachains/runtimes/assets/statemine/src/lib.rs
+++ b/parachains/runtimes/assets/statemine/src/lib.rs
@@ -49,7 +49,7 @@ use frame_support::{
 	construct_runtime,
 	dispatch::DispatchClass,
 	parameter_types,
-	traits::{AsEnsureOriginWithArg, EitherOfDiverse, InstanceFilter},
+	traits::{AsEnsureOriginWithArg, EitherOfDiverse, InstanceFilter, SortedMembers},
 	weights::{ConstantMultiplier, Weight},
 	PalletId, RuntimeDebug,
 };
@@ -87,11 +87,11 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("statemine"),
 	impl_name: create_runtime_str!("statemine"),
 	authoring_version: 1,
-	spec_version: 9300,
+	spec_version: 9301,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 8,
-	state_version: 0,
+	state_version: 1,
 };
 
 /// The version information used to identify this runtime when compiled natively.
@@ -596,6 +596,8 @@ construct_runtime!(
 		// The main stage.
 		Assets: pallet_assets::{Pallet, Call, Storage, Event<T>} = 50,
 		Uniques: pallet_uniques::{Pallet, Call, Storage, Event<T>} = 51,
+
+		StateTrieMigration: pallet_state_trie_migration = 52,
 	}
 );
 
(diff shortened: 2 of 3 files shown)


parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -979,3 +981,68 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+		vec![KEY_ROOT_MIG_CONTROLLER.into()]
+	}
+}
+
+impl SortedMembers<AccountId> for MigController {
+	fn sorted_members() -> Vec<AccountId> {
+		// hardcoded key of controller for manual migration
+		vec![KEY_MIG_CONTROLLER.into()]
+	}
+}
```
bkchr, line 1034:
```suggestion
frame_support::ord_parameter_types! {
    pub const KEY_MIG_CONTROLLER = hex!("INSERT_HEX_STRING");
    pub const KEY_ROOT_MIG_CONTROLLER = hex!("INSERT_HEX_STRING");
}

const KEY_ROOT_MIG_CONTROLLER: [u8; 32] = [
	82, 188, 113, 193, 236, 165, 53, 55, 73, 84, 45, 253, 240, 175, 151, 191, 118, 79, 156, 47, 68,
	232, 96, 205, 72, 95, 28, 216, 100, 0, 246, 73,
];

const KEY_MIG_CONTROLLER: [u8; 32] = [
	82, 188, 113, 193, 236, 165, 53, 55, 73, 84, 45, 253, 240, 175, 151, 191, 118, 79, 156, 47, 68,
	232, 96, 205, 72, 95, 28, 216, 100, 0, 246, 73,
];

impl SortedMembers<AccountId> for RootMigController {
	fn sorted_members() -> Vec<AccountId> {
		// hardcoded key of controller for manual migration
		vec![KEY_ROOT_MIG_CONTROLLER.into()]
	}
}

impl SortedMembers<AccountId> for MigController {
	fn sorted_members() -> Vec<AccountId> {
		// hardcoded key of controller for manual migration
		vec![KEY_MIG_CONTROLLER.into()]
	}
}
```

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -87,11 +87,11 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("statemine"),
 	impl_name: create_runtime_str!("statemine"),
 	authoring_version: 1,
-	spec_version: 9300,
+	spec_version: 9301,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 8,
-	state_version: 0,
+	state_version: 1,
```
bkchr, line 94: We should feature gate this for now. 

As the GCP is still deploying the chain on Rococo for testing.

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -596,6 +596,8 @@ construct_runtime!(
 		// The main stage.
 		Assets: pallet_assets::{Pallet, Call, Storage, Event<T>} = 50,
 		Uniques: pallet_uniques::{Pallet, Call, Storage, Event<T>} = 51,
+
+		StateTrieMigration: pallet_state_trie_migration = 52,
```
bkchr, line 600: Should also be feature gated.

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -979,3 +996,47 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+	use sp_core::crypto::Ss58Codec;
+	let acc =
+		AccountId::from_ss58check("5DwBmEFPXRESyEam5SsQF1zbWSCn2kCjyLW51hJHXe9vW4xs").unwrap();
+	//panic!("{:x?}", acc);
+	assert_eq!(acc, MigController::sorted_members()[0]);
+	let acc =
+		AccountId::from_ss58check("5DwBmEFPXRESyEam5SsQF1zbWSCn2kCjyLW51hJHXe9vW4xs").unwrap();
+	assert_eq!(acc, RootMigController::sorted_members()[0]);
+	//panic!("{:x?}", acc);
+}
```
bkchr, line 1042: Do we need this?

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -87,11 +87,11 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("statemine"),
 	impl_name: create_runtime_str!("statemine"),
 	authoring_version: 1,
-	spec_version: 9300,
+	spec_version: 9301,
```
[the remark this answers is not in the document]
bkchr: Yeah, use the same version for now.

bkchr (approved):

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -596,6 +610,9 @@ construct_runtime!(
 		// The main stage.
 		Assets: pallet_assets::{Pallet, Call, Storage, Event<T>} = 50,
 		Uniques: pallet_uniques::{Pallet, Call, Storage, Event<T>} = 51,
+
+		#[cfg(feature = "state-trie-migration")]
+		StateTrieMigration: pallet_state_trie_migration = 52,
```
joepetrowski, line 615: Just for tidyness I'd leave the 50s clear for new pallets related to the core functionality of Statemine. Could this be `5` and just go with the system stuff?
```suggestion
		StateTrieMigration: pallet_state_trie_migration = 70,
```

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -82,6 +82,20 @@ impl_opaque_keys! {
 	}
 }
 
+#[cfg(feature = "state-trie-migration")]
+#[sp_version::runtime_version]
+pub const VERSION: RuntimeVersion = RuntimeVersion {
```
joepetrowski, line 87: DQ but why does this introduce a whole new `VERSION`?

cheme: The sp_version::runtime_version do not accept `cfg(feature...` I remember trying to do the macro change for it in the past, but since this is just rather `temporary` code, having a bit of duplication here looks fine to me.

parachains/runtimes/assets/statemine/src/lib.rs
cheme: I use it to ensure the controller account is correct (from the ss58 one).
I did check again why I did it in the first place, turns out Ss58Codec is behind 'full_crypto' feature and 'from_ss58check' is behind std.
I don't really see why from_ss58check could not be no_std, just would be a bit more work to switch it.

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -979,3 +996,47 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+}
+
+#[cfg(feature = "state-trie-migration")]
+impl pallet_state_trie_migration::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Currency = Balances;
+	type SignedDepositPerItem = MigrationSignedDepositPerItem;
+	type SignedDepositBase = MigrationSignedDepositBase;
+	// An origin that can control the whole pallet: should be Root, or a part of your council.
+	type ControlOrigin = frame_system::EnsureSignedBy<RootMigController, AccountId>;
```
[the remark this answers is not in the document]
cheme: @bckchr So the idea is to remove `RootMigController` and use an xcm message instead to start, stop migration?

Could make sense to have the other MigController (not the root one) stored and filled by a extrinsic from this origin too (but it means changing the substrate crate :().

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -979,3 +996,47 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
+	//type SignedFilter = frame_system::EnsureSigned<Self::AccountId>;
+
+	// Replace this with weight based on your runtime.
+	type WeightInfo = pallet_state_trie_migration::weights::SubstrateWeight<Runtime>;
+
+	type MaxKeyLen = MigrationMaxKeyLen;
+}
+
+frame_support::ord_parameter_types! {
+	pub const MigController: AccountId = AccountId::from(hex_literal::hex!("52bc71c1eca5353749542dfdf0af97bf764f9c2f44e860cd485f1cd86400f649"));
```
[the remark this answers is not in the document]
cheme: Ok this replies my previous question. I think it would be a good direction, just it needs to change the crate a bit, can be worth it I think.

parachains/runtimes/assets/statemine/src/lib.rs
cheme: I did switch to 70 as in the suggested change, let me know if it should be `5` (the numbering scheme is something I am not familiar with).

parachains/runtimes/assets/statemine/src/lib.rs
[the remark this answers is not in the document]
bkchr: > @bckchr So the idea is to remove `RootMigController` and use an xcm message instead to start, stop migration?

Yes. I mean this is the way that we will need to use on main net.

> Could make sense to have the other MigController (not the root one) stored and filled by a extrinsic from this origin too (but it means changing the substrate crate :().

Which other controller?

parachains/runtimes/assets/statemine/src/lib.rs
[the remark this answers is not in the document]
bkchr: After thinking about it again, it probably doesn't make that much sense. The upgrade needs to be voted on any way. This means, adding the controller again in some extra council/democracy approved call is a little bit of wasted work. Aka, we can just include it directly in the upgrade.

parachains/runtimes/assets/statemine/src/lib.rs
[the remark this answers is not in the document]
bkchr: Just keep the signed account in code.

parachains/runtimes/assets/statemine/Cargo.toml
```diff
@@ -72,16 +73,22 @@ pallet-xcm-benchmarks = { git = "https://github.com/paritytech/polkadot", branch
...
 [features]
 default = [ "std" ]
+# When enabled the `state_version` is set to `1`.
+# This means that the chain will start using the new state format. The migration is lazy, so
+# it requires to write a storage value to use the new state format. To migrate all the other
+# storage values that aren't touched the state migration pallet is added as well.
+# This pallet will migrate the entire state, controlled through some account.
+#
+# This feature should be removed when the main-net will be migrated.
+state-trie-migration = ["pallet-state-trie-migration"]
```
kianenigma, line 90: Why are we adding this feature? the next release/runttime should have this pallet, and once we are done with it, we can remove. 

I don't see why this requires a new feature. This might increase the chance of making a mistake and compiling different/wrong runtimes.

kianenigma (changes requested):
Not sure if/why we need a new feature flag.

parachains/runtimes/assets/statemine/Cargo.toml
bkchr: I proposed to add the feature. Especially for changing the `state_version`. I don't want this in master until this in tested on Rococo/Westend. If that is done, we can remove the feature flag. Aka when we want to do this on main net. For the test nets I'm okay with the risk that someone forgets this.

kianenigma: Okay, but let's make sure we have an issue to clear it later.

kianenigma (approved):
please merge https://github.com/paritytech/cumulus/pull/1424#pullrequestreview-1165031263 as well before starting the devops process.

parachains/runtimes/assets/statemine/Cargo.toml
kianenigma: What if we compile this runtime for a chain with this feature and later on make a mistake and not do it? What happens if we revert to state-version 0 again? @cheme 

Also, the feature should be named `state-version-1` or something similar.

cheme: I am pretty sure we were running some test from v1 to v0 at some point, so I think we would migrate to 0.
Outside the migration question, all changes will be written in the old format. So then we probably would run another migration round to restore them to v1.
So sure will not be a good thing (can move a migrated chain back to hybrid state).

cheme:
bot merge

cheme:
bot merge

PR #1748, opened 2022-10-09
contracts: Reduce session length and enable unstable interfaces

athei:
This changes two things:

- We reduce the session length so we can change collators more quickly. 
- Enable unstable interfaces for contracts. This is a test net. We want to experiment with those new unstable interfaces here before stabilizing them. Even if that may brick already deployed contracts.
- Set `MinCandidates` to 0 to allow only invulnerables as collators. Otherwise no candidate can ever leave if there are not `5`others.


ascjones (approved):

cmichi (approved):

athei:
bot merge

PR #1749, opened 2022-10-10
Companion for paritytech/polkadot#6117

altonen:


dmitry-markin (approved):

client/relay-chain-minimal-node/src/network.rs
```diff
@@ -116,6 +134,54 @@ pub(crate) fn build_collator_network(
 /// we provide a noop implementation.
 struct DummyChainSync;
 
+impl DummyChainSync {
```
skunert, line 137: This does not need to be on the `DummyChainSync` no?

altonen: It doesn't have to be but it's part of `ChainSync` in Substrate so I thought making it part of `DummyChainSync` would closest to what we have in Substrate

altonen:
> Since the substrate PR already merged you probably need to update substrate to make CI happy

Could you elaborate? This is a companion for a [polkadot companion](https://github.com/paritytech/polkadot/pull/6117) (which is showing green for this) so unless I've misunderstood something about the process, CI should be happy once the Polkadot PR is merged and this PR's references are updated accordingly.

skunert:
> > Since the substrate PR already merged you probably need to update substrate to make CI happy
> 
> Could you elaborate? This is a companion for a [polkadot companion](https://github.com/paritytech/polkadot/pull/6117) (which is showing green for this) so unless I've misunderstood something about the process, CI should be happy once the Polkadot PR is merged and this PR's references are updated accordingly.

Ah yes, should be fine then :+1:

client/relay-chain-minimal-node/src/network.rs
skunert: On second thought, it might make more sense to have a separate method in substrate that exposes this. In case something changes we want the cumulus implementation to follow.

altonen: I believe the vision is to make it more transparent so that you could instantiate some syncing engine implementation that would then install the necessary notification and request protocols to, e.g., `sc_network::Params` so that there would be no need for calling, e.g. `light_client_requests::generate_protocol_config()` (except for in the syncing implementation of course) but I'm not opposed to exposing this through Substrate as a temporary measure.

The fire I'm trying to put out right now though is that Polkadot CI is blocked by this PR. How do you want to proceed with this?

skunert: Okay, I think its fine for now then, we could have a follow up.

skunert (approved):

PR #1755, opened 2022-10-12
Fix a few migration issues with 2D weights

KiChjang:


pallets/parachain-system/src/migration.rs
```diff
@@ -14,39 +14,53 @@
...
 			.saturating_add(T::DbWeight::get().writes(1));
 		StorageVersion::new(1).put::<Pallet<T>>();
 	}
 
 	weight
 }
 
-/// V1: `LastUpgrade` block number is removed from the storage since the upgrade
-/// mechanism now uses signals instead of block offsets.
-mod v1 {
```
KiChjang, line 42: I'm unsure if V1 has already enacted on chain, perhaps I shouldn't have replaced the code here?

pallets/xcmp-queue/src/migration.rs
```diff
@@ -81,7 +82,10 @@ pub fn migrate_to_v2<T: Config>() -> Weight {
 			resume_threshold: pre.resume_threshold,
 			threshold_weight: Weight::from_ref_time(pre.threshold_weight),
 			weight_restrict_decay: Weight::from_ref_time(pre.weight_restrict_decay),
-			xcmp_max_individual_weight: Weight::from_ref_time(pre.xcmp_max_individual_weight),
+			xcmp_max_individual_weight: Weight::from_parts(
+				pre.xcmp_max_individual_weight,
+				MAX_POV_SIZE as u64,
```
bkchr, line 87: Not sure it is safe to use `MAX_POV_SIZE` everywhere?

pallets/parachain-system/src/migration.rs
```diff
@@ -14,18 +14,19 @@
...
 /// The current storage version.
-pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(1);
+pub const STORAGE_VERSION: StorageVersion = StorageVersion::new(2);
 
 /// Call this during the next runtime upgrade for this module.
 pub fn on_runtime_upgrade<T: Config>() -> Weight {
-	let mut weight: Weight = Weight::zero();
+	let mut weight: Weight = T::DbWeight::get().reads(2);
 
 	if StorageVersion::get::<Pallet<T>>() == 0 {
```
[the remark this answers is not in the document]
bkchr: Yeah, why not?

pallets/parachain-system/src/migration.rs
```diff
@@ -34,9 +35,45 @@ pub fn on_runtime_upgrade<T: Config>() -> Weight {
...
+
 	weight
 }
 
+/// V2: Migrate to 2D weights for ReservedXcmpWeightOverride and ReservedDmpWeightOverride.
+mod v2 {
+	use super::*;
+
+	pub fn migrate<T: Config>() -> Weight {
+		let translate = |pre: u64| -> Weight { Weight::from_parts(pre, MAX_POV_SIZE as u64) };
```
gavofyork, line 53: do we really want to use `MAX_POV_SIZE` here? is it a limit?

KiChjang: It appears so:

https://github.com/paritytech/cumulus/blob/882a892b60e25449903052b946d25bd3a00f30eb/pallets/parachain-system/src/lib.rs#L804-L830

Used as the `max_weight` to `handle_dmp_messages`. It does not seem to be getting any of these numbers from the remaining block weights, so this is set arbitrarily by the developer.

bkchr: But we clearly don't want to use `MAX_POV_SIZE` as the max weight here. When we reach this point of executing the messages, we already don't have the `MAX_POV_SIZE` available anymore.

KiChjang: Okay, so let's give it... 64 KB?

bkchr: Sounds reasonable to me.

bkchr (approved):

shawntabrizi (approved):

ggwpez (approved):
Do messages already have a PoV attached? No, or?

PR #1777, opened 2022-10-18
Reduce parachain template cognitive complexity

davxy:
Striving for simplicity, this PR removes some generics that are not really required in the template and thus reduces (at least a bit) some of the cognitive complexity required to understand the codebase by a new contributor

---

In particular here:
- some convenience type alias are defined
- remove generic `Runtime`. The parachain template is using only one runtime type

### DIFF
--- a/parachain-template/node/src/service.rs
+++ b/parachain-template/node/src/service.rs
@@ -58,10 +58,8 @@ type ParachainBackend = TFullBackend<Block>;
 ///
 /// Use this macro if you don't actually need the full service, but just the builder in order to
 /// be able to perform chain operations.
-#[allow(clippy::type_complexity)]
-pub fn new_partial<BIQ>(
+pub fn new_partial(
 	config: &Configuration,
-	build_import_queue: BIQ,
 ) -> Result<
 	PartialComponents<
 		ParachainClient,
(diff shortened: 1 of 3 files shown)


davxy:
Added a type alias for `TFullClient` and `TFullBackend` in the `polkadot-parachain` crate as well.

This is opinionated, but I personally think that in general we should - to improve code readability - use more type aliases when we can.

parachain-template/node/src/service.rs
```diff
@@ -52,56 +48,39 @@ impl sc_executor::NativeExecutionDispatch for TemplateRuntimeExecutor {
...
+
 /// Starts a `ServiceBuilder` for a full service.
 ///
 /// Use this macro if you don't actually need the full service, but just the builder in order to
 /// be able to perform chain operations.
 #[allow(clippy::type_complexity)]
-pub fn new_partial<RuntimeApi, Executor, BIQ>(
+pub fn new_partial<BIQ>(
 	config: &Configuration,
 	build_import_queue: BIQ,
```
bkchr, line 64: This doesn't need to be configurable, whoever copied the code didn't know this. So, if you remove this, the code will get even less complicated :D

bkchr (approved):
Thank you! :) 

You can make it even simpler, otherwise it looks good.

bkchr:
> Added a type alias for `TFullClient` and `TFullBackend` in the `polkadot-parachain` crate as well.
> 
> This is opinionated, but I **personally** think that in general we should - in order to improve code readability - use more type aliases when we can. Especially where there are a _bazilion_ of trait bounds for generics :-)

Yeah, we also have in other places similar types aliases. Not sure why we never used them here.

parachain-template/node/src/service.rs
davxy: Indeed. I'll do it before merge. Ty

altonen (approved):

PR #1782, opened 2022-10-19
Single ParachainBlockImport instance

davxy:
This PR allows to share a single `ParachainBlockImport` instance between the import queue and authoring engine, thus to have a shared state between the two components.

The implementation roughly follows the same design already used by Substrate node components.

---

The feature will be immediately used by https://github.com/paritytech/cumulus/pull/1559

### DIFF
--- a/test/service/src/lib.rs
+++ b/test/service/src/lib.rs
@@ -30,7 +30,9 @@ use url::Url;
 
 use crate::runtime::Weight;
 use cumulus_client_cli::CollatorOptions;
-use cumulus_client_consensus_common::{ParachainCandidate, ParachainConsensus};
+use cumulus_client_consensus_common::{
+	ParachainBlockImport, ParachainCandidate, ParachainConsensus,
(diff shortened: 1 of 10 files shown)


test/service/src/lib.rs
```diff
@@ -157,7 +161,7 @@ pub fn new_partial(
 
 	let import_queue = cumulus_client_consensus_relay_chain::import_queue(
 		client.clone(),
-		client.clone(),
+		block_import,
```
bkchr, line 164: You probably also want to return this from `new_partial`

bkchr (approved):

test/service/src/lib.rs
davxy: Done in https://github.com/paritytech/cumulus/pull/1782/commits/90329ffd657df12ce560881e0fa719f13985927b

michalkucharczyk (approved):

PR #1803, opened 2022-10-25
Companion for aura CompatibilityMode

bkchr:
As no parachain was launched with the broken authority set change handling we don't need to expose the compatibility mode.

### DIFF
--- a/client/consensus/aura/src/lib.rs
+++ b/client/consensus/aura/src/lib.rs
@@ -143,6 +143,7 @@ where
 				telemetry,
 				block_proposal_slot_portion,
 				max_block_proposal_slot_portion,
+				compatibility_mode: sc_consensus_aura::CompatibilityMode::None,
 			},
 		);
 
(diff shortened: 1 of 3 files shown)


client/consensus/aura/src/lib.rs
```diff
@@ -143,6 +143,7 @@ where
 				telemetry,
 				block_proposal_slot_portion,
 				max_block_proposal_slot_portion,
+				compatibility_mode: sc_consensus_aura::CompatibilityMode::None,
```
davxy, line 146:
```suggestion
				compatibility_mode: CompatibilityMode::None,
```
Silly nit :-) (feel free to ignore). Just to be inline with the way the symbol is imported in the import queue source.
Obviously this requires to adjust the `sc_consensus_aura` import list at the begin of the file

davxy (approved):

koute (approved):

skunert (approved):

michalkucharczyk (approved):

andresilva (approved):

PR #1807, opened 2022-10-30
Add RPC nodes to pov-recovery test

### LINKED ISSUE #1744: Fix pov-recovery zombienet test
The test as it is currently on master is missing `--disable-block-announcements` on the charlie node. However, adding this flag makes the test flaky. 

Todos: 
- Disable block announcements on charlie
- Investigate issues with pov-recovery, why is it sometimes failing?

skunert:
Add some nodes that use external relay chain node to the pov-recovery test

closes #1744

### DIFF
--- a/zombienet_tests/0002-pov_recovery.feature
+++ b/zombienet_tests/0002-pov_recovery.feature
@@ -10,6 +10,8 @@ validator-3: is up
 alice: is up within 60 seconds
 bob: is up within 60 seconds
 charlie: is up within 60 seconds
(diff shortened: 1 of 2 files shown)


koute (approved):
Don't know how feasible it is, but it'd be nice if we could also add some extra check (e.g. by grepping the logs?) that explicitly checks that the PoV recovery took place. Might not be strictly necessary, but it'd be nice to have just in case, and also as a form of documentation.

bkchr (approved):

Issue #585, opened 2021-08-24
Provide polkadot launch configs


bkchr:
We should provide some polkadot-launch configs to make it easier to play around with all of the stuff.

bkchr:
Polkadot launch is dead!

PR #1812, opened 2022-11-03
add statemine parachain-launch config

samelamin:
Added some parachain launch configs to help spin up the chains easily

https://github.com/paritytech/cumulus/issues/585

For reference polkadot-launch has been deprecated in favour of parachain launch


samelamin:
@bkchr sorry I must have misunderstood the ticket.  

I can push documentation on how to run it, I just assumed people who use parachain launch would be comfortable with spinning things up

Would you rather I add documentation how to use this or close this ticket

I would say that Zombienet is significantly more complex than parachain launch which essentially builds you a docker compose file so you can spin everything up with docker-compose

Zombienet has a few more requirements such as Kubernetes which might be a bit too much for someone new 

Let me know what you would like me to do 😄

bkchr:
After reading your comment in the zombienet roadmap pull request. I was maybe wrong. I was coming from the pov of a developer that changes something in the parachain and want to test this. Your example here was maybe more around adding an example for people who want to run the parachain locally to test some stuff on top of the parachain? If yes, I'm okay with reopening. Then please add some docs on how to use this.

samelamin:
@bkchr docs have been added, let me know if they are unclear or if there is anything else you would like me to add

scripts/parachain-launch/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+```
+docker-compose -f output/docker-compose.yml up -d
+```
+
+To regenerate the compose file, first ensure `parachain-launch` is installed and run the following command
+
+```
+parachain-launch generate statemine.yml
+```
```
bkchr, line 21: I think this should be reverted. We also should not commit the docker compose file. You should describe the way on how to convert the statemine.yml file to the docker compose file and then how to run this. Then we can remove the entire `output` folder.

samelamin: I would have thought having the docker-compose file added might have been simpler since its one less thing to do, but I understand it might be simpler to only have the config file in the repo

scripts/parachain-launch/README.md
```diff
@@ -0,0 +1,24 @@
+# parachain-launch
```
bkchr, line 1:
```suggestion
# Parachain-launch
```

scripts/parachain-launch/statemine.yml
```diff
@@ -0,0 +1,40 @@
...
+  flags:
+    - --rpc-methods=unsafe
+    - --force-authoring
+    - --wasm-execution=compiled
+    - --execution=wasm
+  relaychainFlags:
+    - --wasm-execution=compiled
+    - --execution=wasm
+  env:
+    RUST_LOG: sc_basic_authorship=trace,cumulus-consensus=trace,cumulus-collator=trace,collator_protocol=trace,collation_generation=trace,aura=debug
```
bkchr, line 37: Do we need them?

scripts/parachain-launch/statemine.yml
```diff
@@ -0,0 +1,40 @@
+relaychain:
+  image: parity/polkadot:v0.9.29
+  chain: rococo-local
+  runtimeGenesisConfig:
+    configuration:
+      config:
+        validation_upgrade_cooldown: 10
+        validation_upgrade_delay: 10
+  env:
+    RUST_LOG: parachain::candidate-backing=trace,parachain::candidate-selection=trace,parachain::pvf=debug,parachain::provisioner=trace
```
bkchr, line 10: And these?

scripts/parachain-launch/statemine.yml
```diff
@@ -0,0 +1,40 @@
...
+  flags:
+    - --rpc-methods=unsafe
+    - --wasm-execution=compiled
+    - --execution=wasm
```
bkchr, line 14:
```suggestion
```

These are the defaults.

scripts/parachain-launch/statemine.yml
```diff
@@ -0,0 +1,40 @@
...
+  id: 2000
+  parachain: true
+  flags:
+    - --rpc-methods=unsafe
+    - --force-authoring
+    - --wasm-execution=compiled
+    - --execution=wasm
+  relaychainFlags:
+    - --wasm-execution=compiled
+    - --execution=wasm
```
bkchr, line 35:
```suggestion
```

bkchr (approved):

scripts/parachain-launch/statemine.yml
samelamin: We dont need them, but it could help if you are trying to debug. 

The parachain-launch template examples include them. 

Would you like them removed?

bkchr: Yes, I would like to have them removed.

samelamin: Removed

skunert:
A bit late to the party, but I am a bit confused about the comments about zombienet.
As of today, you can totally spin up zombienet networks without kubernetes knowledge by using the `--provider native` flag. It will use the specified binaries just like polkadot launch. Also you can just spin up networks that run forever by just using the `spawn` instead of `test` command.

samelamin:
> A bit late to the party, but I am a bit confused about the comments about zombienet. As of today, you can totally spin up zombienet networks without kubernetes knowledge by using the `--provider native` flag. It will use the specified binaries just like polkadot launch. Also you can just spin up networks that run forever by just using the `spawn` instead of `test` command.

Hey @skunert the issue is with zombienet and specifically, the native flag is you have to have the native libraries installed.
But what if I am a DAPP developer who justs wants to spin things up locally because I need to test something, or I am new or exploring building on Substrate. 

Why make them wait to compile the binaries - which do take some time depending on your hardware - when we already have them precompiled on docker.

It becomes a much more pleasant developer user experience if you can spin up the entire testnet using one or 2 lines without requiring building the binaries or even preparing your environment for them to build


I hope that helps clarify things

skunert:
I see were you are coming from and like the goal of starting a network with minimum hassle!

My main problem is that I do not see added benefit.  There is also the `podman` provider in zombienet for those who want to use docker images but do not like kubernetes, all your arguments apply there. With `kubernetes`, `podman` and `native` provider, there should be something for everyone. So instead of additional tooling I would have liked more to see a network config file for zombienet for this use case.

However this is not a super strong opinion. If more people think that we should have an additional entry point, I am fine with going forward with this :+1:

bkchr:
> "Starting from scratch. can I spin everything up using 1 line"

I mean this isn't correct. You also need to install parachain launch, the same is true for zombienet&podman. So in overall, there isn't such a big difference. You will also need to build the docker compose file and then run it. I would propose @skunert creates a pr showing how it would be done in zombienet with an example readme on how to use it.

samelamin:
> > "Starting from scratch. can I spin everything up using 1 line"
> 
> I mean this isn't correct. You also need to install parachain launch, the same is true for zombienet&podman. So in overall, there isn't such a big difference. You will also need to build the docker compose file and then run it. I would propose @skunert creates a pr showing how it would be done in zombienet with an example readme on how to use it.

I agree that it isn't exactly the same. That's why I said it's closer to a yes rather than a yes 

That's also why I had the output folder committed because then you don't need to install anything.

If we can achieve the same with zombienet and use the docker images rather than build the binaries then yes that would be the same thing 

I'm happy helping out in any way but - based on my understanding and I could very well be wrong - that's not possible on zombienet today because the docker provider isn't available 

If we need to build the chain binaries then I would argue this is a simpler step to get a testnet running

bkchr:
@samelamin see https://github.com/paritytech/cumulus/pull/1839 by @skunert. Podman is just a different env for running docker containers.

PR #1835, opened 2022-11-08
Fixes for gav-xcm-v3

bkontur:


parachains/common/src/impls.rs
```diff
@@ -87,13 +88,16 @@ where
...
 pub struct NonZeroIssuance<AccountId, Assets>(PhantomData<(AccountId, Assets)>);
-impl<AccountId, Assets> Contains<<Assets as fungibles::Inspect<AccountId>>::AssetId>
-	for NonZeroIssuance<AccountId, Assets>
+impl<AccountId, Assets> AssetChecking<Assets::AssetId> for NonZeroIssuance<AccountId, Assets>
 where
 	Assets: fungibles::Inspect<AccountId>,
 {
-	fn contains(id: &<Assets as fungibles::Inspect<AccountId>>::AssetId) -> bool {
-		!Assets::total_issuance(*id).is_zero()
+	fn asset_checking(id: &Assets::AssetId) -> Option<MintLocation> {
```
bkontur, line 95: @joepetrowski 
I am not sure about this change, could you please check it?

there is a change for `FungiblesAdapter`: Contains -> AssetChecking:

actual master - `CheckAsset: Contains<Assets::AssetId>,`
https://github.com/paritytech/polkadot/blob/master/xcm/xcm-builder/src/fungibles_adapter.rs#L265

new xcm v3 - `CheckAsset: AssetChecking<Assets::AssetId>,`
https://github.com/paritytech/polkadot/blob/gav-xcm-v3/xcm/xcm-builder/src/fungibles_adapter.rs#L298

joepetrowski: I'm not sure of the motivation to change this interface, but from reviewing [its introduction](https://github.com/paritytech/polkadot/pull/5237/files#diff-aa492552e19e60a60e15bbdadb3de42f0b9ab33659e04ef6c23a4225f8f9cbe0), especially the docs on `MintLocation`, I don't think the logic is correct, that if an asset has some issuance then its `MintLocation` should be `Local`.

We could say that the `MintLocation` of bridged/para assets is `NonLocal`, however other locations within the network still trust system paras to the extent that they trust Polkadot governance, so would see the asset on Statemint as "equally local".

KiChjang: I don't think `NonZeroIssuance` should be changed at all. What looks to me is that it's meant to be used in conjunction with `LocalMint`, `NonLocalMint` and `DualMint`, rather than have `AssetChecking` directly implemented on it.

bkontur: @joepetrowski 
ok, next iteration :), I put everywhere `NonLocalMint`,

Lets say:
- we are configuring Statemint here, where we have pallet_assets
- we have Moonbeam parachain with GLMR

What is the basic use case for Moonbeam to use Statemint?
"Somebody" from Moonbeam should create asset for GLMR at first?
Then MoonbeamUser can transfer GLMR to some StatemintUserAccount? Then some StatemintUserAccount can transfer to other accounts or move something back to the Moonbeam (but not more then he receives), so this means `NonLocalMint`?

-----------
What would be example for `LocalMint` from the Statemint perspective?

-----------
You wrote here before: 
> We could say that the `MintLocation` of bridged/para assets is `NonLocal`

e.g.: How could I resolve/check that in the code? I am checking structs: `AssetDetails` / `AssetAccount` / `AssetMetadata`

or similar: implementation for `DualMint`, how could we differ/tell that some Asset should be `LocalMint` or `NonLocalMint`?

joepetrowski: > "Somebody" from Moonbeam should create asset for GLMR at first?

The `CreateOrigin` should ensure that only Moonbeam itself, i.e. via their Root origin (however it is accessed) could create GLMR. So yeah, they'd need some referendum to send a message to Statemint registering GLMR.

> Then MoonbeamUser can transfer GLMR to some StatemintUserAccount?

Right, so they should have (probably already do) some dispatchable function that will construct an XCM to send to Statemint. Something like `fn reserve_transfer_asset(GLMR, Statemint, my_address) { ... }`, and inside it would put together the `[BuyExecution, ReserveAssetDeposited, ...]` etc. instructions. For UX they'd probably want to `WithdrawAsset` from the Moonbeam sovereign account and charge a slightly higher tx fee on the Moonbeam side to cover it. Then their treasury can top up the sovereign account DOT every now and then.

> Then some StatemintUserAccount can transfer to other accounts or move something back to the Moonbeam (but not more then he receives), so this means NonLocalMint?

We are not dealing with teleports, so I'm not sure how much this `MintLocation` applies to e.g. GLMR. Since the `para` origin corresponding to Moonbeam conforms to the `CreateOrigin = ForeignCreators`, "this chain" (being Moonbeam) can mint the asset on Statemint. But, as I read these docs (full enum below), the key part for me is: "We ensure that...

- ...no more of the asset returns back to the chain than has been sent out." => This is nonsense, because in order for assets to be "sent out" they will have had to be created first.
- ...no more of the asset is sent out from the chain than has been previously received." => This makes sense, because the GLMR started on Moonbeam, and Statemint has received it, and an issuance of 0 would indicate that it's all been sent out, and it should be further unable to withdraw any more.

Based on that, I would call all parachain token `NonLocalMint`. An example of `LocalMint` would be DOT and the assets created under the first (existing) instance of the Assets pallet, e.g. USDT. So, as to how to resolve it in code, I would say:

- Any assets under `pallet_balances || pallet_assets::Instance1` are `Local` and any assets under `pallet_assets::Instance2` are `NonLocal`. @KiChjang wdyt?

```rust
/// The location which is allowed to mint a particular asset.
#[derive(Copy, Clone, Eq, PartialEq)]
pub enum MintLocation {
	/// This chain is allowed to mint the asset. When we track teleports of the asset we ensure that
	/// no more of the asset returns back to the chain than has been sent out.
	Local,
	/// This chain is not allowed to mint the asset. When we track teleports of the asset we ensure
	/// that no more of the asset is sent out from the chain than has been previously received.
	NonLocal,
}
```

bkontur: @joepetrowski 
ok, Joe, thank you, so I will change it to something like this:
```
LocalMint<parachains_common::impls::NonZeroIssuance<AccountId, AssetsPlusBalanceContainsImpl>>
```
and later for westmint/bridge-hub stuff to:
```
DualMint<
       parachains_common::impls::NonZeroIssuance<AccountId, TrustBackedAssetsPlusBalanceContainsImpl>,
       parachains_common::impls::NonZeroIssuance<AccountId, ForeignAssets>
>
```

KiChjang: @joepetrowski If instance 1 of the assets pallet refers to the one created using integer IDs and instance 2 of the assets pallet represents those created with `MultiLocation` IDs, then this makes sense. Although I do have to echo your sentiment about how this may not be _very_ relevant, as it is only used for teleports. I don't really see Statemint/e allowing for teleports between community parachains anytime soon; the most obvious use cases are however between system parachains teleporting DOT/KSM to each other.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2074497

parachains/common/src/impls.rs
bkontur: @joepetrowski 
so, I pushed there: `LocalMint<NonZeroIssuance<AccountId, Assets>>`,

or do we want to fix this just for DOT/KSM (which means allow relay chain currency)? in this case, how is identified "DOT/KSM" as AssetId?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2075599

parachains/common/src/impls.rs
joepetrowski: It would just be `{ parents: 1, interior: Here }`, which should correspond to the Balances pallet, not an AssetId.

parachains/runtimes/assets/statemine/src/xcm_config.rs
```diff
@@ -96,7 +96,7 @@ pub type FungiblesTransactor = FungiblesAdapter<
 	AccountId,
 	// We only want to allow teleports of known assets. We use non-zero issuance as an indication
 	// that this asset is known.
-	parachains_common::impls::NonZeroIssuance<AccountId, Assets>,
+	LocalMint<parachains_common::impls::NonZeroIssuance<AccountId, Assets>>,
```
KiChjang, line 99: I actually don't understand why this is `LocalMint`? This is Assets 1 as we discussed, yeah? Then it should be receiving native parachain tokens, and hence should be `NonLocalMint`.

KiChjang: DOT/KSM is not ever going to be represented in the Assets pallet; they're instead going to be represented in the Balances pallet, and hence won't even use `FungiblesAdapter` but instead `CurrencyAdapter`.

KiChjang: Oh wait nevermind, this is correct: instance 1 assets are integer-based, and so can be created locally by any user.

PR #1851, opened 2022-11-10
bump txn versions

EgorPopelyaev:
According to the [Extrinsic ordering check](https://github.com/paritytech/cumulus/issues/1844#issuecomment-1309050901) transaction_version needs to bumped

### DIFF
--- a/parachains/runtimes/starters/seedling/src/lib.rs
+++ b/parachains/runtimes/starters/seedling/src/lib.rs
@@ -68,7 +68,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 9320,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 2,
 	state_version: 0,
 };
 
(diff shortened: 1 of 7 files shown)


chevdor (approved):
LGTM

coderobe (approved):

parachains/runtimes/starters/seedling/src/lib.rs
```diff
@@ -68,7 +68,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 9320,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 2,
```
muharem, line 71: version bump for shell runtime is missing. 
also noticed that starters runtime is missing in the release template.

EgorPopelyaev: Shell does not seem to change mach https://github.com/paritytech/cumulus/issues/1844#issuecomment-1309034509
and should be fine without bump

muharem (approved):

joepetrowski (approved):

parachains/runtimes/starters/seedling/src/lib.rs
chevdor: > version bump for shell runtime is missing.

Not required

> also noticed that starters runtime is missing in the release template.

They used to be, they are now excluded on purpose to avoid bloating the release notes with starter runtimes that barely change. They are however rebuilt and provided as artifacts in each release.

PR #1891, opened 2022-11-17
Explicit recovery for announced candidate blocks

davxy:
This PR augments https://github.com/paritytech/cumulus/pull/1559 with block recovery via PoV.

(target branch has been set to https://github.com/paritytech/cumulus/pull/1559 branch in order to better highlight the diff and have a more  focused conversation)

---

A recovery channel end is passed to the parachain consensus worker and it is used to trigger/stimulate block recovery whenever the best parachain block is not locally present yet.

The rx side of the channel can be whatever and not strictly necessarily the PoV recovery mechanism.

This code is very unlikely from being triggered in practice. But better safe than sorry.

---

The tests were not aligned yet to some  of the new interfaces, so CI failure is expected.  Feedback first :-)

### DIFF
--- a/client/consensus/common/src/parachain_consensus.rs
+++ b/client/consensus/common/src/parachain_consensus.rs
@@ -145,7 +145,7 @@ pub async fn run_parachain_consensus<P, R, Block, B>(
 	parachain: Arc<P>,
 	relay_chain: R,
 	announce_block: Arc<dyn Fn(Block::Hash, Option<Vec<u8>>) + Send + Sync>,
-	recovery_chan_tx: Sender<RecoveryRequest<Block>>,
+	recovery_chan_tx: Option<Sender<RecoveryRequest<Block>>>,
 ) where
 	Block: BlockT,
 	P: Finalizer<Block, B>
@@ -178,7 +178,7 @@ async fn follow_new_best<P, R, Block, B>(
 	parachain: Arc<P>,
 	relay_chain: R,
 	announce_block: Arc<dyn Fn(Block::Hash, Option<Vec<u8>>) + Send + Sync>,
-	recovery_chan_tx: Sender<RecoveryRequest<Block>>,
+	recovery_chan_tx: Option<Sender<RecoveryRequest<Block>>>,
 ) where
 	Block: BlockT,
 	P: Finalizer<Block, B>
@@ -307,7 +307,7 @@ async fn handle_new_best_parachain_head<Block, P>(
 	head: Vec<u8>,
 	parachain: &P,
 	unset_best_header: &mut Option<Block::Header>,
-	mut recovery_chan_tx: Sender<RecoveryRequest<Block>>,
+	mut recovery_chan_tx: Option<Sender<RecoveryRequest<Block>>>,
 ) where
 	Block: BlockT,
 	P: UsageProvider<Block> + Send + Sync + BlockBackend<Block>,
@@ -357,17 +357,20 @@ async fn handle_new_best_parachain_head<Block, P>(
 					"Parachain block not yet imported, waiting for import to enact as best block.",
 				);
 
-				// Best effort to trigger a recovery.
-				// Error is not fatal. The relay chain will re-announce the best block anyway,
-				// thus we will have other opportunities to retry.
-				let req = RecoveryRequest { hash, delay: RECOVERY_DELAY, kind: RecoveryKind::Full };
-				if let Err(err) = recovery_chan_tx.try_send(req) {
-					tracing::warn!(
-						target: LOG_TARGET,
-						block_hash = ?hash,
-						error = ?err,
-						"Unable to notify block recovery subsystem"
-					)
+				if let Some(ref mut recovery_chan_tx) = recovery_chan_tx {
+					// Best effort channel to actively encourage block recovery.
+					// An error here is not fatal; the relay chain continuously re-announces
+					// the best block, thus we will have other opportunities to retry.
+					let req =
+						RecoveryRequest { hash, delay: RECOVERY_DELAY, kind: RecoveryKind::Full };
+					if let Err(err) = recovery_chan_tx.try_send(req) {
+						tracing::warn!(
+							target: LOG_TARGET,
+							block_hash = ?hash,
+							error = ?err,
+							"Unable to notify block recovery subsystem"
+						)
+					}
 				}
 			},
 			Err(e) => {
--- a/client/consensus/common/src/parachain_consensus.rs
+++ b/client/consensus/common/src/parachain_consensus.rs
@@ -15,7 +15,6 @@
 // along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
 
 use async_trait::async_trait;
-use cumulus_relay_chain_interface::{RelayChainInterface, RelayChainResult};
 use sc_client_api::{
 	Backend, BlockBackend, BlockImportNotification, BlockchainEvents, Finalizer, UsageProvider,
 };
@@ -27,6 +26,9 @@ use sp_runtime::{
 	traits::{Block as BlockT, Header as HeaderT},
 };
 
+use cumulus_primitives_core::{RecoveryDelay, RecoveryKind, RecoveryRequest};
+use cumulus_relay_chain_interface::{RelayChainInterface, RelayChainResult};
+
 use polkadot_primitives::v2::{Hash as PHash, Id as ParaId, OccupiedCoreAssumption};
 
 use codec::Decode;
@@ -35,6 +37,8 @@ use futures::{channel::mpsc::Sender, select, FutureExt, Stream, StreamExt};
 use std::{pin::Pin, sync::Arc};
 
 const LOG_TARGET: &str = "cumulus-consensus";
+
+const RECOVERY_MAX_DELAY: std::time::Duration = std::time::Duration::from_secs(10);
 
 /// Helper for the relay chain client. This is expected to be a lightweight handle like an `Arc`.
 #[async_trait]
@@ -136,7 +140,7 @@ pub async fn run_parachain_consensus<P, R, Block, B>(
 	parachain: Arc<P>,
 	relay_chain: R,
 	announce_block: Arc<dyn Fn(Block::Hash, Option<Vec<u8>>) + Send + Sync>,
-	recovery_chan_tx: Sender<Block::Hash>,
+	recovery_chan_tx: Sender<RecoveryRequest<Block>>,
 ) where
 	Block: BlockT,
 	P: Finalizer<Block, B>
@@ -169,7 +173,7 @@ async fn follow_new_best<P, R, Block, B>(
 	parachain: Arc<P>,
 	relay_chain: R,
 	announce_block: Arc<dyn Fn(Block::Hash, Option<Vec<u8>>) + Send + Sync>,
-	recovery_chan_tx: Sender<Block::Hash>,
+	recovery_chan_tx: Sender<RecoveryRequest<Block>>,
 ) where
 	Block: BlockT,
 	P: Finalizer<Block, B>
@@ -298,7 +302,7 @@ async fn handle_new_best_parachain_head<Block, P>(
 	head: Vec<u8>,
 	parachain: &P,
 	unset_best_header: &mut Option<Block::Header>,
-	mut recovery_chan_tx: Sender<Block::Hash>,
+	mut recovery_chan_tx: Sender<RecoveryRequest<Block>>,
 ) where
 	Block: BlockT,
 	P: UsageProvider<Block> + Send + Sync + BlockBackend<Block>,
@@ -351,7 +355,15 @@ async fn handle_new_best_parachain_head<Block, P>(
 				// Best effort to trigger a recovery.
 				// Error is not fatal. The relay chain will re-announce the best block anyway,
 				// thus we will have other opportunities to retry.
-				if let Err(err) = recovery_chan_tx.try_send(hash) {
+				let req = RecoveryRequest {
+					hash,
+					delay: RecoveryDelay {
+						min: core::time::Duration::ZERO,
+						max: RECOVERY_MAX_DELAY,
+					},
+					kind: RecoveryKind::Full,
+				};
+				if let Err(err) = recovery_chan_tx.try_send(req) {
 					tracing::warn!(
 						target: LOG_TARGET,
 						block_hash = ?hash,
--- a/client/consensus/common/src/parachain_consensus.rs
+++ b/client/consensus/common/src/parachain_consensus.rs
@@ -34,11 +34,16 @@ use polkadot_primitives::v2::{Hash as PHash, Id as ParaId, OccupiedCoreAssumptio
 use codec::Decode;
 use futures::{channel::mpsc::Sender, select, FutureExt, Stream, StreamExt};
 
-use std::{pin::Pin, sync::Arc};
+use std::{pin::Pin, sync::Arc, time::Duration};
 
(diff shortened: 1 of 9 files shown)


client/pov-recovery/src/lib.rs
```diff
@@ -343,13 +344,55 @@ where
...
+						target: LOG_TARGET,
+						error = ?e,
+						block_hash = ?hash,
+						"Failed to get block status",
+					);
+					break false
+				},
+			}
+
+			hash = match self.pending_candidates.get(&hash) {
```
skunert, line 364: Maybe I am missing something, but is it really an error if the hash is not in `pending_candidates`?

I see that you removed the removal of imported blocks, but it could still have been previously recovered and removed here no? https://github.com/paritytech/cumulus/blob/55520786c37f57600102ecfa6a008fb92add9a0e/client/pov-recovery/src/lib.rs#L233

client/pov-recovery/src/lib.rs
```diff
@@ -408,15 +450,9 @@ where
 						return;
 					}
 				},
-				imported = imported_blocks.next() => {
```
skunert, line 411: I dont think this is save to remove. Before we recover a block we check if the block has been imported in the meantime here:
https://github.com/paritytech/cumulus/blob/55520786c37f57600102ecfa6a008fb92add9a0e/client/pov-recovery/src/lib.rs#L233

Without removing the imported blocks from `pending_candidates`, we would always trigger a recovery even if the block is known by then.

client/pov-recovery/src/lib.rs
```diff
@@ -343,13 +344,55 @@ where
...
+		};
+
+		if do_recover {
+			for hash in to_recover.iter().rev() {
+				tracing::debug!(
+					target: LOG_TARGET,
+					block_hash = ?hash,
+					"Starting eager recovery",
+				);
+				self.recover_candidate(*hash).await;
```
skunert, line 384: Why are we not needing the delay here? This could potentially lead to many recoveries at the same time.

client/pov-recovery/src/lib.rs
davxy: Good point. Indeed we may trigger the recovery of something that in the meantime has been already imported.

I'm open to suggestion, but since an imported block may unexpectly "disappear" under our foot because of the "limit monitor" (or some other future reason) maybe the safer way is to re-check the state after the delay? Or just set a bool in the `PendingCandidate` to prevent auto recovery after the delay. E.g. a `pending_recovery` bool.

The entry is not removed on "block_import" anymore because we may need it to reconstruct the chain of missing entries in the backend (I added a `parent` entry in the `PendingCandidate` struct)

client/pov-recovery/src/lib.rs
davxy: The delay has not been added because we here we want to trigger the recovery immediately for all the missing blocks. Do you think that is better to add a random delay anyway? E.g. to spread the recovery in a bigger timespan?

client/pov-recovery/src/lib.rs
davxy: When we trigger an explicit recovery of block X we also want to silently trigger the recovery of all its ancestry.

To do this we need the block X ancestry hashes. This hashes are recovered from the `PendingCandidates` entries we have already saw.

If a parent of the block that we want to explicitly recover has never been announced by the RelayChain (I.e. is not in the pending candidates yet) then we are not able to perform the recovery at the moment

client/pov-recovery/src/lib.rs
skunert: Per this comment, the delay prevents spamming of the relay chain. 
https://github.com/skunert/cumulus/blob/master/client/pov-recovery/src/lib.rs#L209

As I understand, your level monitor behaves deterministically on all collators. So if there were a problem with the relay chain and all collators would hit the level limit, the monitor would delete the same branches on every collator, right? This also means that when a new best block finally comes in, all collators would request recovery of the same blocks simultaneously (given that it was deleted by the level monitor). 

Mabe @bkchr can clarify what the spamming scenario here was.

client/pov-recovery/src/lib.rs
skunert: Maybe we need to rethink the semantics of `pending_candidates`. Before this PR it was only adding unknown blocks. 

After your change, it is adding every candidate we see, and only removes on finalization or recovery. So blocks that are already known on notification stay until finalization. 

One idea could be to change `pending_candidates` to contain all pending candidates and remove only on finalization. We would need a separate cache that has all the imported hashes of the last $RecoveryDelay seconds to do the import check. I think this would also solve the problem from the other comment https://github.com/paritytech/cumulus/pull/1891#discussion_r1028517177

client/pov-recovery/src/lib.rs
bkchr: Having a random delay between/before requests is some "good" behavior to not dos your own architecture.

client/pov-recovery/src/lib.rs
```diff
@@ -175,51 +182,45 @@ where
 		}
 
 		let hash = header.hash();
+
+		if self.pending_candidates.contains_key(&hash) {
```
bkchr, line 186: Pending candidates should be renamed, as these are now not anymore "pending candidates"

client/pov-recovery/src/lib.rs
```diff
@@ -343,13 +344,55 @@ where
...
+				},
+			}
+
+			hash = match self.pending_candidates.get(&hash) {
+				Some(p) => p.parent_hash,
+				None => {
+					tracing::error!(
+						target: LOG_TARGET,
+						block_hash = ?hash,
+						"Cound not recover, block has never been announced as a candidate"
```
bkchr, line 370:
```suggestion
						"Cound not recover. Block was never announced as candidate"
```

client/pov-recovery/src/lib.rs
```diff
@@ -408,15 +450,9 @@ where
...
-					} else {
-						tracing::debug!(
-							target: LOG_TARGET,
-							"Imported blocks stream ended",
-						);
-						return;
+				recovery_req = self.recovery_chan_rx.next() => {
+					if let Some(hash) = recovery_req {
+						self.chain_recovery(hash).await;
 					}
```
bkchr, line 456: else { return }

client/pov-recovery/src/lib.rs
```diff
@@ -343,13 +344,55 @@ where
 		self.import_block(block).await;
 	}
 
+	async fn chain_recovery(&mut self, mut hash: Block::Hash) {
```
bkchr, line 347: Some docs on the logic would be nice.

bkchr (commented):
`recover_candidate` still calls `pending_candidate.remove`. This shouldn't be done.

client/pov-recovery/src/lib.rs
davxy: Addressed by https://github.com/paritytech/cumulus/pull/1891/commits/144445d36f86ecc7e227f2d7161daf4e94ed0028

client/pov-recovery/src/lib.rs
davxy: Addressed by https://github.com/paritytech/cumulus/pull/1891/commits/144445d36f86ecc7e227f2d7161daf4e94ed0028

client/pov-recovery/src/lib.rs
davxy: Addressed by https://github.com/paritytech/cumulus/pull/1891/commits/e7660ea44908a183964f69e19019f1ccb4d250ec

client/consensus/common/src/parachain_consensus.rs
```diff
@@ -27,15 +26,25 @@ use sp_runtime::{
...
+use std::{pin::Pin, sync::Arc, time::Duration};
 
 const LOG_TARGET: &str = "cumulus-consensus";
 
+// Delay range to trigger explicit requests.
+// The chosen value doesn't have any special meaning, a random delay within the order of
+// seconds in practice should be a good enough to allow a quick recovery without DOSing
+// the relay chain.
+const RECOVERY_DELAY: RecoveryDelay =
+	RecoveryDelay { min: Duration::ZERO, max: Duration::from_secs(30) };
```
davxy, line 46: The "automatic" recovery requests triggered when the relay chain announces a candidate are submitted according to the values specified in the `service` crate.

https://github.com/paritytech/cumulus/blob/e7660ea44908a183964f69e19019f1ccb4d250ec/client/service/src/lib.rs#L128

https://github.com/paritytech/cumulus/blob/e7660ea44908a183964f69e19019f1ccb4d250ec/client/service/src/lib.rs#L219

As can be seen the values to be used are different depending if the node is a full-node or a collator.

---

This type of recovery here instead I think that should be independent of the type of node as we require the value immediately and the delay is passed only to prevent DOS-ing the relay chain with the same request from all the nodes

The value used here is pretty arbitrary but good enough. I.e. a random value that goes from 0 to 30 seconds...
Maybe we can lower it?

davxy (commented):
All the review suggestions should have been addressed.

Regarding the renaming of `pending_candidates` I went for just `candidates`. 

The semantic here is that these are blocks announced as backed by the relay chain and that are candidates of becoming part of the (finalized) canonical blocks for the relay chain.

Candidates are removed from the list when are finalized

client/consensus/common/src/tests.rs
```diff
@@ -187,6 +214,68 @@ fn follow_new_best_works() {
 	});
 }
 
+#[test]
+fn follow_new_best_with_dummy_recovery_works() {
```
davxy, line 218: Test to excercise explicit best block recovery.

Even though currently we are using as a recovery mechanism the `PoVRecovery` subsystem, the motivation of this test is to also show how the recovery can be generic.

client/pov-recovery/src/lib.rs
```diff
@@ -423,8 +353,75 @@ where
...
+				None => {
+					tracing::error!(
+						target: LOG_TARGET,
+						block_hash = ?hash,
+						"Cound not recover. Block was never announced as candidate"
+					);
+					break false
+				},
+				// Recovery already in progress
+				_ => break false,
```
skunert, line 373: This branch makes the assumption that for `RecoveryKind::Full` none of the ancestors are in recovery right? Are we sure this assumption is always correct? Lets say we want to recover `C` and its ancestors `A->B->C`. If `A` has `waiting_recovery = true`, we will also not recover `B` and `C`.

davxy: Good point. Fix in progress

client/pov-recovery/src/lib.rs
```diff
@@ -423,8 +353,75 @@ where
...
+			if kind == RecoveryKind::Simple {
+				break true
+			}
+
+			hash = candidate.parent_hash;
+		};
+
+		if do_recover {
+			for hash in to_recover.into_iter().rev() {
+				let delay =
```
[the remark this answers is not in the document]
davxy: So, at this point the question mostly is: How the response is sent back to the parachain? Only to the node that is requesting it or to all the nodes listening to some topic?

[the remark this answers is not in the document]
skunert: I think your comment is correct, after recovery the parachain nodes import the block and announce it on the network, so the other nodes can get it too. So yeah makes sense then to keep the delay as it is :+1:

primitives/core/src/lib.rs
```diff
@@ -194,6 +195,37 @@ impl<B: BlockT> ParachainBlockData<B> {
...
+
+/// The delay between observing an unknown block and triggering the recovery of a block.
+#[derive(Clone, Copy)]
+pub struct RecoveryDelay {
+	/// Start recovering after `min` delay.
+	pub min: Duration,
+	/// Start recovering before `max` delay.
+	pub max: Duration,
+}
+
```
bkchr, line 228: Do we really need to have them defined here? I think they should stay in `pov-recovery` and then we import this crate in `consensus/common`

client/pov-recovery/src/lib.rs
```diff
@@ -380,6 +353,70 @@ where
...
+					break false
+				},
+			};
+
+			match self.parachain_client.block_status(&BlockId::Hash(hash)) {
+				Ok(BlockStatus::Unknown) if !candidate.waiting_recovery => {
+					candidate.waiting_recovery = true;
+					to_recover.push(hash);
+				},
+				Ok(_) => break true,
```
bkchr, line 379:
```suggestion
				Ok(BlockStatus::InChain) => break true,
```

client/pov-recovery/src/lib.rs
```diff
@@ -380,6 +353,70 @@ where
...
+	/// Attempts an explicit recovery of one or more blocks.
+	pub fn recover(&mut self, req: RecoveryRequest<Block>) {
+		let RecoveryRequest { mut hash, delay, kind } = req;
+		let mut to_recover = Vec::new();
+
+		let do_recover = loop {
+			let candidate = match self.candidates.get_mut(&hash) {
+				Some(candidate) => candidate,
+				None => {
+					tracing::error!(
```
bkchr, line 365:
```suggestion
					tracing::debug!(
```

I mean when you just have restarted your node, this is expected behavior. The operator also can not really do anything against this.

client/pov-recovery/src/lib.rs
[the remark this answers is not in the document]
bkchr: The chunks are requested using a request response protocol. This is not done using any kind of pubsub or similar (which would be horrible slow). 

But, nevertheless we can keep the randomness between blocks. I mean in the end this should be only very rarely being triggered and maybe another node in the network recovers before us and we receive a block announcement from this node and download the block in this way.

bkchr (approved):

primitives/core/src/lib.rs
davxy: At the moment not really, as you correctly suggested we can just import `pov-recovery`in `consensus/common`.

The rationale behind moving the basic types of block recovery out of PoCRecovery was to abstract the recovery mechanism.

When the consensus code finds that there is a new **best** it will trigger a recovery by sending a message (`RecoveryRequest`) via an mpsc channel. The recovery action can then **potentially** happen via any mean  (dummy examples: torrent, ipfs, a distributed filesystem, local filesystem, ...).

 I know, this is not the case **now**. But in this way we don't bound cumulus user of our consensus  to one possibility.

sandreim:
Seeing this error: `[Parachain] XXX Unable getting route to any leaf from 0xeec619ba0097d7aa2c1278574ac6d986cd56d1418429bd2b8bd8528e1ac4d6bf (this is a bug)
`

sandreim:
> Seeing this error: `[Parachain] XXX Unable getting route to any leaf from 0xeec619ba0097d7aa2c1278574ac6d986cd56d1418429bd2b8bd8528e1ac4d6bf (this is a bug) `

Tested with the base branch, not this PR. Logs here: https://grafana.parity-mgmt.parity.io/goto/2evUx3FVk?orgId=1

bkchr:
@sandreim you fucked up the master merge! :see_no_evil:

sandreim:
> @sandreim you fucked up the master merge! 🙈

Damn, I'll revert this, and force push ...

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2141586

primitives/core/src/lib.rs
davxy: Ok. I will put these structs it in the `pov-recovery` crate in the other PR

PR #1900, opened 2022-11-21
Relax Copy to Clone on asset_id (companion)

gilescope:
Companion PR to https://github.com/paritytech/substrate/pull/12731

### DIFF
--- a/parachains/common/src/impls.rs
+++ b/parachains/common/src/impls.rs
@@ -94,7 +94,7 @@ where
 	Assets: fungibles::Inspect<AccountId>,
 {
 	fn contains(id: &<Assets as fungibles::Inspect<AccountId>>::AssetId) -> bool {
-		!Assets::total_issuance(*id).is_zero()
+		!Assets::total_issuance(id.clone()).is_zero()
 	}
 }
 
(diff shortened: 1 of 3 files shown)


parachains/common/src/impls.rs
```diff
@@ -94,7 +94,7 @@ where
 	Assets: fungibles::Inspect<AccountId>,
 {
 	fn contains(id: &<Assets as fungibles::Inspect<AccountId>>::AssetId) -> bool {
-		!Assets::total_issuance(*id).is_zero()
+		!Assets::total_issuance(id.clone()).is_zero()
```
bkontur, line 97: @gilescope 
is there any reason why not to change trait?
from:
`fn total_issuance(asset: Self::AssetId)`
to:
`fn total_issuance(asset: &Self::AssetId)`

like `InspectMetadata`

gilescope: It's a good question - for copy types the clone is as fast as before. I'm in two minds of whether it would be less or more disruptive if we did that, though I agree if they're not Copy types then that would minimise clones.

gilescope:
I think we can close this for now. xcmv3 should be coming soon.

PR #1901, opened 2022-11-22
Template to not allow any Teleport and allowing Reserve transfers as defaults

NukeManDan:
https://github.com/substrate-developer-hub/substrate-parachain-template/issues/99 long time coming

### DIFF
--- a/parachain-template/runtime/src/xcm_config.rs
+++ b/parachain-template/runtime/src/xcm_config.rs
@@ -203,8 +203,8 @@ impl pallet_xcm::Config for Runtime {
 	// ^ Disable dispatchable execute on the XCM pallet.
 	// Needs to be `Everything` for local testing.
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Nothing;
+	type XcmTeleportFilter = Nothing;
+	type XcmReserveTransferFilter = Everything;
 	type Weigher = FixedWeightBounds<UnitWeightCost, RuntimeCall, MaxInstructions>;
 	type LocationInverter = LocationInverter<Ancestry>;
 	type RuntimeOrigin = RuntimeOrigin;


al3mart (approved):

gilescope (approved):

parachain-template/runtime/src/xcm_config.rs
```diff
@@ -203,8 +203,8 @@ impl pallet_xcm::Config for Runtime {
 	// ^ Disable dispatchable execute on the XCM pallet.
 	// Needs to be `Everything` for local testing.
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type XcmTeleportFilter = Everything;
-	type XcmReserveTransferFilter = Nothing;
+	type XcmTeleportFilter = Nothing;
+	type XcmReserveTransferFilter = Everything;
```
bkchr, line 207: Would be better with some comments on what these settings are doing?

bkchr (approved):

NukeManDan:
Looking to update the comments as suggested, I am not sure how this is enforcing what the comment says it does:

```rust
/// No local origins on this chain are allowed to dispatch XCM sends/executions.
pub type LocalOriginToLocation = SignedToAccountId32<RuntimeOrigin, AccountId, RelayNetwork>;
```

As we also have:  
```rust
pub const RelayNetwork: NetworkId = NetworkId::Any;
```

IIUC NetworkId could be anything including "self" (the local/intrenal NetworkId)... is this a bug or am I missing something?

parachain-template/runtime/src/xcm_config.rs
```diff
@@ -111,7 +109,7 @@ where
 	}
 }
 
-// See issue #5233
+// See issue https://github.com/paritytech/polkadot/issues/5233
```
NukeManDan, line 112: this issue is marked as resolved. Is this still relevant?

parachain-template/runtime/src/xcm_config.rs
NukeManDan: Dumb q: why should you not do this? And if you do, what would the effect be? 

Assuming it's something that should (almost) never happen, is there any check possible to ensure that there are no collisions in these settings, something like `Contains` impl for these so they explicitly exclude each other could help you from screwing this up.

joepetrowski: Because an attacker can drain the `CheckedAccount` on the reserve location by doing a reserve transfer to the parachain and then teleporting the asset back. This will then prevent others from teleporting back to the reserve location.

stiiifff:
@NukeManDan @joepetrowski gentle nudge on this one. Without it the parachain template does not provide a sane default XCM config.

PR #1903, opened 2022-11-23
Remove av-store, av-distribution, chain-api subsystems from minimal node

skunert:
This PR removes the following subsystems from the minimal node:
- Availability-Store (collator node should not store chunks)
- Availability-Distribution (since we have nothing to store, we do not distribute)
- Chain-API-Subsystem (This was used by the others)

Originally I was under the impression that we need these for PoV-Recovery, but that is not the case.
We still need to register the chunk fetching protocol, since we send requests for this. Answering was done by
the availability-distribution subsystem, which is now removed. If we receive such a request we always send back `NoSuchChunk` because we are guaranteed to not have it.


client/relay-chain-minimal-node/src/lib.rs
```diff
@@ -200,24 +202,47 @@ async fn new_minimal_relay_chain(
...
+fn drain_unwanted_request_channels(
+	task_manager: &TaskManager,
+	mut chunk_req_receiver: IncomingRequestReceiver<v1::ChunkFetchingRequest>,
+) {
+	task_manager.spawn_handle().spawn("chunk-request-drainer", None, async move {
+		loop {
+			if let Ok(request) = chunk_req_receiver.recv(|| vec![COST_INVALID_REQUEST]).await {
+				tracing::debug!(
+					target: LOG_TARGET,
+					"Received chunk fetching request. Collators should not receive this."
```
altonen, line 217: Is it a protocol violation to ask chunks/PoVs from collator nodes?

altonen (approved):

client/relay-chain-minimal-node/src/lib.rs
```diff
@@ -152,8 +158,10 @@ async fn new_minimal_relay_chain(
 		.extend(peer_sets_info(is_authority, &peer_set_protocol_names));
 
 	let request_protocol_names = ReqProtocolNames::new(genesis_hash, config.chain_spec.fork_id());
-	let (collation_req_receiver, available_data_req_receiver, pov_req_receiver, chunk_req_receiver) =
+	let (collation_req_receiver, available_data_req_receiver, chunk_req_receiver) =
 		build_request_response_protocol_receivers(&request_protocol_names, &mut config);
+	drain_unwanted_request_channels(&task_manager, chunk_req_receiver);
```
bkchr, line 163: Okay, this is unneeded. We can not do two things. Either we drop this channel or we do not even register the channel.

client/relay-chain-minimal-node/src/lib.rs
```diff
@@ -200,24 +202,47 @@ async fn new_minimal_relay_chain(
...
 	let (collation_req_receiver, cfg) =
 		IncomingRequest::get_config_receiver(request_protocol_names);
 	config.network.request_response_protocols.push(cfg);
 	let (available_data_req_receiver, cfg) =
 		IncomingRequest::get_config_receiver(request_protocol_names);
 	config.network.request_response_protocols.push(cfg);
-	let (pov_req_receiver, cfg) = IncomingRequest::get_config_receiver(request_protocol_names);
-	config.network.request_response_protocols.push(cfg);
 	let (chunk_req_receiver, cfg) = IncomingRequest::get_config_receiver(request_protocol_names);
 	config.network.request_response_protocols.push(cfg);
```
bkchr, line 246:
```suggestion
	let (_, cfg) = IncomingRequest::get_config_receiver(request_protocol_names);
	
	// We don't support inbound requests for this protocol.
	let cfg = RequestResponseConfig { inbound_queue: None, ..cfg };
	
	config.network.request_response_protocols.push(cfg);
```

bkchr: We should probably support this in Polkadot directly. When it is a collator/full node we should not register an incoming receiver

client/relay-chain-minimal-node/src/lib.rs
skunert: The [spec](https://spec.polkadot.network/#sect-candidate-recovery) says you should request this from validators. Everything else also does not make sense. But you can still send the request to everyone else running the availability-distribution subsystem, and they will [respond with `NoSuchChunk`](https://github.com/paritytech/polkadot/blob/e1cd7cec7a7bf4e441ff500fa4137c1cb526dd1e/node/network/availability-distribution/src/responder.rs).

skunert:
Waiting for https://github.com/paritytech/polkadot/pull/6343, then we can remove the inbound channel.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2084661

bkchr (approved):

bkchr:
bot merge

Issue #1856, opened 2022-11-10
Add Additional Instance of Assets Pallet to Statemint


joepetrowski:
Add a new instance of the Assets pallet to Statemint. This should be configured such that:

```rust
pub type ForeignAssets = pallet_assets::Instance2;
impl pallet_assets::Config<ForeignAssets> for Runtime {
	// snip...
	type AssetId = MultiLocation;
	type CreateOrigin = ForeignCreators;
}

pub struct ForeignCreators;
impl EnsureOriginWithArg<RuntimeOrigin, MultiLocation> for ForeignCreators {
	type Success = AccountId;

	fn try_origin(
		o: RuntimeOrigin,
		a: &MultiLocation,
	) -> sp_std::result::Result<Self::Success, RuntimeOrigin> {
		// impl...
	}
}
```

An example can be found in the [gav-xcm-v3 branch](https://github.com/paritytech/polkadot/blob/gav-xcm-v3/xcm/xcm-simulator/example/src/parachain.rs#L142-L163) of the XCM simulator.

---

This pallet will support assets from other parachains, and in XCM v3, from other consensus systems. For example:
```rust
// GLMR
asset_id = MultiLocation { parents: 1, interior: X1(Parachain(2004)) }
// xc20 token on Moonbeam
asset_id = MultiLocation { parents: 1, interior: X2(Parachain(2004), AccountId20(None, contract_address)) }
// ETH
asset_id = MultiLocation { parents: 2, interior: X1(GlobalConsensus(Ethereum)) }
```

joepetrowski:
Done in https://github.com/paritytech/cumulus/pull/2540

PR #1908, opened 2022-11-23
Instantiate All Assets Pallets

joepetrowski:
One step toward https://github.com/paritytech/cumulus/issues/1856

This should be a no-op in the runtime ~except for the migration of storage keys~. It's done as a preparation for an upgrade that will add another instance to the chains.

Open to another name than "Trust Backed Assets", although they do tend to have privileged roles corresponding to accessible accounts.


parachains/common/src/xcm_config.rs
```diff
@@ -82,24 +82,30 @@ pub struct AssetFeeAsExistentialDepositMultiplier<Runtime, WeightToFee, BalanceC
...
 where
-	Runtime: pallet_assets::Config,
+	Runtime: pallet_assets::Config<pallet_assets::Instance1>,
 	WeightToFee: WeightToFeePolynomial<Balance = CurrencyBalance>,
 	BalanceConverter: BalanceConversion<
 		CurrencyBalance,
-		<Runtime as pallet_assets::Config>::AssetId,
-		<Runtime as pallet_assets::Config>::Balance,
+		<Runtime as pallet_assets::Config<pallet_assets::Instance1>>::AssetId,
+		<Runtime as pallet_assets::Config<pallet_assets::Instance1>>::Balance,
```
joepetrowski, line 93: How can I make this generic over any instance of `pallet_assets`?

ggwpez: Can you not do it similar to the `AssetsToBlockAuthor` by adding a `AssetInstance: 'static` generic?

parachains/common/src/impls.rs
```diff
@@ -71,17 +71,18 @@ where
 
 /// A `HandleCredit` implementation that naively transfers the fees to the block author.
 /// Will drop and burn the assets in case the transfer fails.
-pub struct AssetsToBlockAuthor<R>(PhantomData<R>);
-impl<R> HandleCredit<AccountIdOf<R>, pallet_assets::Pallet<R>> for AssetsToBlockAuthor<R>
+pub struct AssetsToBlockAuthor<R, I = ()>(PhantomData<(R, I)>);
```
ggwpez, line 74:
```suggestion
pub struct AssetsToBlockAuthor<R, I>(PhantomData<(R, I)>);
```
I am not sure about the default here. Could accidentally reference the wrong instance AFAIK.

shawntabrizi:
AFAIK you do not need to add `Instance1` for the default instance. You can just add instances when you want to add new ones. The default one can stay "instanceless" which will still make it unique among instantiated pallets.

joepetrowski:
> AFAIK you do not need to add `Instance1` for the default/first instance. You can just add instances when you want to add new ones. The default one can stay "instanceless" which will still make it unique among instantiated pallets.

Interesting, OK I'll try a demo with that. We still need the underlying changes to some of the other traits to deal with multiple instances of Assets.

joepetrowski:
> > AFAIK you do not need to add `Instance1` for the default/first instance. You can just add instances when you want to add new ones. The default one can stay "instanceless" which will still make it unique among instantiated pallets.
> 
> Interesting, OK I'll try a demo with that. We still need the underlying changes to some of the other traits to deal with multiple instances of Assets.

I tried doing this but ran into a lot of errors when generics in other structs asked for a specific instance of the Assets pallet. Is there any downside to just leaving an explicit `Instance`?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2250510

parachains/runtimes/assets/westmint/src/lib.rs
```diff
@@ -669,9 +678,20 @@ pub type Executive = frame_executive::Executive<
...
+	// Do not release until verified!
+	MigrateAssetsPallet,
 >;
 
+pub struct MigrateAssetsPallet;
+impl frame_support::traits::OnRuntimeUpgrade for MigrateAssetsPallet {
+	fn on_runtime_upgrade() -> Weight {
+		use frame_support::storage::migration;
+		migration::move_pallet(b"Assets", b"TrustBackedAssets");
+		<Runtime as frame_system::Config>::DbWeight::get().writes(1)
```
bkchr, line 691: The weight is not really correct :P

bkchr (approved):
Mainly renamings and changing of the prefix. Looks good, but the weight needs to be corrected for each migration.

parachains/runtimes/assets/westmint/src/lib.rs
joepetrowski: Yeah, I was thinking of removing the rename because it will break most wallets/interfaces that read from Statemint.

bkchr: Yeah also sounds reasonable. I mean we can just name the new ones appropriate.

gilescope:
@joepetrowski if you want you could define `pub type TrustBackedAssetsInstance = ();` and  construct_runtime macro can stay as `Assets: pallet_assets::{Pallet, Call, Storage, Event<T>} = 50,`). - that seems to build ok on my machine. That would keep it using the default instance.

joepetrowski:
OK, is there any advantage to not using the instance? IMO `Instance1` is more expressive than `()`.

bkontur:
> OK, is there any advantage to not using the instance? IMO `Instance1` is more expressive than `()`.

I agree and I am curious also,
for some bridge pallets I needed to add `Instance2` and without `Instance1` it looked strange: why just Instance2 and where is Instance1? so, I ended up with explicit Instance1/Instance2

bkontur:
> ... That would keep it using the default instance.

@gilescope @joepetrowski 
TrustBackedAssetsInstance vs ForeignCreatorAssetsInstance, in theory, can we even say that one of them is **default**? I guess, they are two "equivalent" instances and none is default-like, right?

joepetrowski:
> @gilescope @joepetrowski TrustBackedAssetsInstance vs ForeignCreatorAssetsInstance, in theory, can we even say that one of them is **default**? I guess, they are two "equivalent" instances and none is default-like, right?

I agree in general, but I think Default vs. Instance1 vs. Instance2 etc. just has to do with the macro expansion in the bin. The pallets' storage and API get exposed by their name in `construct_runtime`, so `Assets`, `ForeignAssets`.

For example, in Polkadot, Council and Tech Comm are `Council` and `TechnicalCollective`, not `pallet_collectives::Instance{1, 2}`.

gilescope:
I'm thinking that `()` would equate to PalletInstance `0` in XCM and `Instance1` would equate to PalletInstance `1` so `()` would not change the xcm multilocation of the assets.

joepetrowski:
I thought `PalletInstance()` is the index of the pallet in the runtime, e.g. https://github.com/paritytech/cumulus/blob/9e211b8/parachains/runtimes/assets/westmint/tests/tests.rs#L61

Likewise, the [docs](https://github.com/paritytech/polkadot/blob/6b48ca4/xcm/src/v1/junction.rs#L53) say, "An instanced, indexed pallet that forms a constituent part of the context." I think the context is the runtime, not "all instances of pallet_assets".

@KiChjang can confirm?

gilescope:
Ah that's a good point. Ok happy that the assets will not change their addressable location.

gilescope (approved):

KiChjang:
@joepetrowski is correct.

stefan-sopic (approved):

Issue #1884, opened 2022-11-16
Allow Sufficient Assets for XCM Fee Payment on Statemint


joepetrowski:
Port https://github.com/paritytech/cumulus/pull/1278 to Statemint runtime.

PR #1910, opened 2022-11-24
Allow Sufficient Assets for XCM Fee Payment on Statemint #1884

bkontur:
fixes: https://github.com/paritytech/cumulus/issues/1884

### DIFF
--- a/parachains/common/src/xcm_config.rs
+++ b/parachains/common/src/xcm_config.rs
@@ -64,7 +64,7 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
 			message.0.iter().any(|inst| matches!(inst, ReserveAssetDeposited { .. }))
 		{
 			log::warn!(
-				target: "xcm::barrier",
+				target: "xcm::barriers",
 				"Unexpected ReserveAssetDeposited from the Relay Chain",
 			);
 		}
--- a/parachains/runtimes/assets/statemine/tests/tests.rs
+++ b/parachains/runtimes/assets/statemine/tests/tests.rs
@@ -302,3 +302,70 @@ fn test_that_buying_ed_refund_does_not_refund() {
 			assert_eq!(Assets::total_supply(1), ExistentialDeposit::get());
 		});
 }
+
+#[test]
+fn test_asset_xcm_trader_not_possible_for_non_sufficient_assets() {
+	ExtBuilder::<Runtime>::default()
+		.with_collators(vec![AccountId::from(ALICE)])
+		.with_session_keys(vec![(
+			AccountId::from(ALICE),
+			AccountId::from(ALICE),
+			SessionKeys { aura: AuraId::from(sp_core::sr25519::Public::from_raw(ALICE)) },
+		)])
+		.build()
+		.execute_with(|| {
+			// Create a non-sufficient asset
+			// We set existential deposit to be identical to the one for Balances first
+			assert_ok!(Assets::force_create(
+				RuntimeHelper::<Runtime>::root_origin(),
+				1,
+				AccountId::from(ALICE).into(),
+				false,
+				ExistentialDeposit::get()
+			));
+
+			// We first mint enough asset for the account to exist for assets
+			assert_ok!(Assets::mint(
+				RuntimeHelper::<Runtime>::origin_of(AccountId::from(ALICE)),
+				1,
+				AccountId::from(ALICE).into(),
+				ExistentialDeposit::get()
+			));
+
+			let mut trader = <XcmConfig as xcm_executor::Config>::Trader::new();
+
+			// Set Alice as block author, who will receive fees
+			RuntimeHelper::<Runtime>::run_to_block(2, Some(AccountId::from(ALICE)));
+
+			// We are going to buy 4e9 weight
+			let bought = 4_000_000_000u64;
+
+			// lets calculate amount needed
+			let amount_needed = WeightToFee::weight_to_fee(&Weight::from_ref_time(bought));
+
+			let asset_multilocation = MultiLocation::new(
+				0,
+				X2(
+					PalletInstance(
+						<Runtime as frame_system::Config>::PalletInfo::index::<Assets>().unwrap()
+							as u8,
+					),
+					GeneralIndex(1),
+				),
+			);
+
+			let asset: MultiAsset = (asset_multilocation, amount_needed).into();
+
+			// Make sure again buy_weight does return an error
+			assert_noop!(trader.buy_weight(bought, asset.into()), XcmError::TooExpensive);
+
+			// Drop trader
+			drop(trader);
+
+			// Make sure author(Alice) has NOT received the amount
+			assert_eq!(Assets::balance(1, AccountId::from(ALICE)), ExistentialDeposit::get());
+
+			// We also need to ensure the total supply NOT increased
+			assert_eq!(Assets::total_supply(1), ExistentialDeposit::get());
+		});
+}
--- a/parachains/runtimes/assets/statemine/tests/tests.rs
+++ b/parachains/runtimes/assets/statemine/tests/tests.rs
@@ -314,14 +314,14 @@ fn test_asset_xcm_trader_not_possible_for_non_sufficient_assets() {
 		)])
 		.build()
 		.execute_with(|| {
-			// Create a non-sufficient asset
-			// We set existential deposit to be identical to the one for Balances first
+			// Create a non-sufficient asset with specific existential deposit
+			let minimum_asset_balance = 1_000_000_u128;
 			assert_ok!(Assets::force_create(
 				RuntimeHelper::<Runtime>::root_origin(),
 				1,
 				AccountId::from(ALICE).into(),
 				false,
-				ExistentialDeposit::get()
+				minimum_asset_balance
 			));
 
 			// We first mint enough asset for the account to exist for assets
@@ -329,7 +329,7 @@ fn test_asset_xcm_trader_not_possible_for_non_sufficient_assets() {
 				RuntimeHelper::<Runtime>::origin_of(AccountId::from(ALICE)),
 				1,
 				AccountId::from(ALICE).into(),
-				ExistentialDeposit::get()
+				minimum_asset_balance
 			));
 
 			let mut trader = <XcmConfig as xcm_executor::Config>::Trader::new();
@@ -363,9 +363,9 @@ fn test_asset_xcm_trader_not_possible_for_non_sufficient_assets() {
 			drop(trader);
 
 			// Make sure author(Alice) has NOT received the amount
-			assert_eq!(Assets::balance(1, AccountId::from(ALICE)), ExistentialDeposit::get());
+			assert_eq!(Assets::balance(1, AccountId::from(ALICE)), minimum_asset_balance);
 
 			// We also need to ensure the total supply NOT increased
-			assert_eq!(Assets::total_supply(1), ExistentialDeposit::get());
+			assert_eq!(Assets::total_supply(1), minimum_asset_balance);
 		});
 }
--- a/parachains/runtimes/assets/statemine/tests/tests.rs
+++ b/parachains/runtimes/assets/statemine/tests/tests.rs
@@ -82,7 +82,7 @@ fn test_asset_xcm_trader() {
 			// Lets buy_weight and make sure buy_weight does not return an error
 			match trader.buy_weight(bought, asset.into()) {
 				Ok(unused_assets) => {
-					// Check if return correct unused assets
+					// Check whether a correct amount of unused assets is returned
 					assert_ok!(unused_assets
 						.ensure_contains(&(asset_multilocation, asset_amount_extra).into()));
 				},
--- a/parachains/runtimes/assets/statemine/tests/tests.rs
+++ b/parachains/runtimes/assets/statemine/tests/tests.rs
@@ -1,16 +1,19 @@
 use asset_test_utils::{ExtBuilder, RuntimeHelper};
+use cumulus_primitives_utility::ChargeWeightInFungibles;
 use frame_support::{
 	assert_noop, assert_ok,
 	traits::PalletInfo,
 	weights::{Weight, WeightToFee as WeightToFeeT},
 };
 use parachains_common::{AccountId, AuraId};
+use statemine_runtime::xcm_config::AssetFeeAsExistentialDepositMultiplierFeeCharger;
 pub use statemine_runtime::{
 	constants::fee::WeightToFee, xcm_config::XcmConfig, Assets, Balances, ExistentialDeposit,
 	Runtime, SessionKeys, System,
 };
 use xcm::latest::prelude::*;
 use xcm_executor::traits::WeightTrader;
+
 pub const ALICE: [u8; 32] = [1u8; 32];
 
 #[test]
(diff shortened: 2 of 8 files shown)


primitives/utility/src/lib.rs
```diff
@@ -120,11 +120,8 @@ impl<
 
 		let weight = Weight::from_ref_time(weight);
 
-		// We take the very first multiasset from payment
-		let multiassets: MultiAssets = payment.clone().into();
-
-		// Take the first multiasset from the selected MultiAssets
-		let first = multiassets.get(0).ok_or(XcmError::AssetNotFound)?;
+		// Take the very first multiasset from from payment
+		let first = payment.assets_iter().next().ok_or(XcmError::AssetNotFound)?;
```
bkontur, line 124: @joepetrowski 
I am just thinking, if it would be better here iterate and find "first.is_sufficient" maybe

bkontur: hmm, also , to remove `MultiAssets = payment.clone().into()` was not a good idea, there is sorting done in `impl From<Vec<MultiAsset>> for MultiAssets {`

bkontur: reverting `payment.clone().into();` it does that sorting `impl From<Vec<MultiAsset>> for MultiAssets {`

bkontur: now it takes "first" from sorted (probably by highest amount),
but it does not consider `is_sufficient` and/or `minimum_balance`,

so, the first asset could have higher amount then second,
but first asset could have also higher minimum_balance than second asset, so it couldnt be charged from first asset, but the second could be charged

bkontur: @joepetrowski 
do we want to change this "first selection" or let it be or just create following issue? wdyt?

at least, I am going to add here unit-test with multiple assets in payment

joepetrowski:
@girazoki

parachains/runtimes/assets/statemint/tests/tests.rs
```diff
@@ -61,7 +69,166 @@ fn test_asset_xcm_trader_does_not_work_in_statemine() {
...
+			drop(trader);
+
+			// Make sure author(Alice) has received the amount
+			assert_eq!(
+				Assets::balance(1, AccountId::from(ALICE)),
+				ExistentialDeposit::get() + amount_needed
+			);
+
+			// We also need to ensure the total supply increased
+			assert_eq!(Assets::total_supply(1), ExistentialDeposit::get() + amount_needed);
```
joepetrowski, line 85: Why did this happen? I don't see a mint anywhere.

parachains/runtimes/assets/statemint/tests/tests.rs
```diff
@@ -61,7 +69,166 @@ fn test_asset_xcm_trader_does_not_work_in_statemine() {
...
+			let mut trader = <XcmConfig as xcm_executor::Config>::Trader::new();
+
+			// Set Alice as block author, who will receive fees
+			RuntimeHelper::<Runtime>::run_to_block(2, Some(AccountId::from(ALICE)));
+
+			// We are going to buy 500e9 weight
+			// Because of the ED being higher in statemine
+			// and not to complicate things, we use a little
+			// bit more of weight
+			let bought = 50_000_000_000u64;
```
joepetrowski, line 209: 50 or 500?

joepetrowski (approved):
Just a few questions but looks good.

parachains/runtimes/assets/statemint/tests/tests.rs
bkontur: few lines above there is a mint :),
```
			assert_ok!(Assets::mint(
				RuntimeHelper::<Runtime>::origin_of(AccountId::from(ALICE)),
				1,
				AccountId::from(ALICE).into(),
				ExistentialDeposit::get()
			));
```

https://github.com/paritytech/cumulus/pull/1910/files#diff-4bcab62dac08701a3e6b73eda0face53c0d8b3b921c802752fdb165ad6a994e8R38-R43

parachains/runtimes/assets/statemine/tests/tests.rs
```diff
@@ -302,3 +302,70 @@ fn test_that_buying_ed_refund_does_not_refund() {
...
+		.build()
+		.execute_with(|| {
+			// Create a non-sufficient asset
+			// We set existential deposit to be identical to the one for Balances first
+			assert_ok!(Assets::force_create(
+				RuntimeHelper::<Runtime>::root_origin(),
+				1,
+				AccountId::from(ALICE).into(),
+				false,
+				ExistentialDeposit::get()
```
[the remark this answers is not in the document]
bkontur: @joepetrowski 
ok, thank you, I guess, I understand now,

for these new tests `test_asset_xcm_trader_not_possible_for_non_sufficient_assets` I changed it to: `let minimum_asset_balance = 1_000_000_u128;` and remove usage of `ExistentialDeposit::get()` and makes much more sense

I guess, everywhere in those Trader tests, `ExistentialDeposit::get()` is used as a some "random" constant/value,
because `ExistentialDeposit` is just used for `pallet_balances::Config` and this is not related to the `pallet_assets::Asset::min_balance`,
using `ExistentialDeposit::get()` just brings confusion

I am going to remove all usage of `ExistentialDeposit::get()` from those tests, ok?

parachains/runtimes/assets/statemine/tests/tests.rs
```diff
@@ -302,3 +302,70 @@ fn test_that_buying_ed_refund_does_not_refund() {
...
+			let mut trader = <XcmConfig as xcm_executor::Config>::Trader::new();
+
+			// Set Alice as block author, who will receive fees
+			RuntimeHelper::<Runtime>::run_to_block(2, Some(AccountId::from(ALICE)));
+
+			// We are going to buy 4e9 weight
+			let bought = 4_000_000_000u64;
+
+			// lets calculate amount needed
+			let amount_needed = WeightToFee::weight_to_fee(&Weight::from_ref_time(bought));
```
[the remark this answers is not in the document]
bkontur: @joepetrowski 
good point, I guess, looks like an another confusion in all of those tests,

`WeightToFee` works with runtime's `Balance`, right?

so, the `amount_needed` is meant to be `DOT or KSM`,
```
let amount_needed = WeightToFee::weight_to_fee(&Weight::from_ref_time(bought));
```

but the value of `amount_needed` is used as a amount of asset:
```
let asset: MultiAsset = (asset_multilocation, amount_needed).into();
```

maybe correct/exact would be something like this:
```
type Dex = AssetFeeAsExistentialDepositMultiplier< 
        Runtime,
        WeightToFee,
        pallet_assets::BalanceToAssetBalance<Balances, Runtime, ConvertInto>,
>;

let local_asset_id = 1,;
let amount_needed = Dex::charge_weight_in_fungibles(local_asset_id, Weight::from_ref_time(bought));
```

do we want to change it here for all tests?

parachains/runtimes/assets/statemine/tests/tests.rs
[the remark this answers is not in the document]
joepetrowski: yeah, that sounds good :+1:

parachains/runtimes/assets/statemint/tests/tests.rs
joepetrowski: doh! thanks

parachains/runtimes/assets/statemine/tests/tests.rs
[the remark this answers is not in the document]
joepetrowski: Yeah, that makes more sense. I wouldn't call it `type Dex` though, as it implies an asset exchange could actually take place. It's really declaring, "how should I convert an amount in DOT to an amount in the asset".

The complicated part is that, if we had a synchronous DEX, we actually could convert the asset using `swap_asset_for_exact(asset: AssetId, fee_amount: Balance)`, but we're not there yet. Of course, actually making the swap would itself consume more weight, so there could be an option to fetch the spot rate on a DEX but still just convert the fee amount into some amount of fungible asset, and take that as a fee.

We should consider `ChargeWeightInFungibles` more for another PR, but I think for where we are right now (in that we are converting DOT fee to asset balance, but not actually providing the ability to swap it), we should look at this as a `Convert` type of item. `AssetFeeAsExistentialDepositMultiplier` specifies the means (just take the ratio of min balance / existential deposit).

Perhaps, `type FeeItemConversion` ("item" implying that ownership of a non-fungible could even be a "ticket" so some execution rights)? But before renaming some of the existing Config interface I'd want to dive a bit deeper on the issue and where we see it going.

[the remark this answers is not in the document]
girazoki: If I recall correctly, the tests did set the same ED for the asset and DOT/KSM, hence the reason we can use `
` as is.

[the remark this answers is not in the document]
bkontur: yes, I see now, all the tests works ok, because if we create asset with `ExistentialDeposit::get()`,
then the `BalanceToAssetBalance` return the same value, because of multiplier is `1` `// balance * asset.min_balance / min_balance`

so these two return the same value
```
let amount_needed = WeightToFee::weight_to_fee(&Weight::from_ref_time(bought));
let amount_needed2 = AssetFeeAsExistentialDepositMultiplierFeeCharger::charge_weight_in_fungibles(local_asset_id, Weight::from_ref_time(bought)).expect("aaa");
```

primitives/utility/src/lib.rs
bkontur: @girazoki 
and please, wdyt about this "first selection" algo?

joepetrowski: Yeah, I like the iter approach. It would make it more user-friendly and make every effort to find an acceptable fee payment solution.

girazoki: multiassets object are always sorted I think (their creation involves sorting assets AFAIR). So we simply take the first.

There is no sufficiency indicator on a MultiAsset, so you probably will have to do db reads to know whether a multiasset is sufficient.

Besides, `BuyExecution` will always contain just one multiasset https://github.com/paritytech/polkadot/blob/53253de77f1649ecbb650df2ec3e29d0d1110a2d/xcm/src/v2/mod.rs#L564. The reason for `buy_weight` to receive `MUltiAssets` instead of `MultiAsset` as fee is because `try_take` from the holding register is defined in `MultiAsset`: https://github.com/paritytech/polkadot/blob/795b20c7150b04759ebddadd419e2f26fa382519/xcm/xcm-executor/src/lib.rs#L455

parachains/runtimes/assets/statemine/tests/tests.rs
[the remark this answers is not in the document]
bkontur: @joepetrowski 
I changed `WeightToFee` with `AssetFeeAsExistentialDepositMultiplierFeeCharger` + some cleaning in all `test_asset_xcm_trader` tests, 

so, I am done here with this PR (if there are no more comments), not to block release or anything, 
and at least, now I understand all this stuff and I will later look at other tests in following issue: https://github.com/paritytech/cumulus/issues/1915 , ok?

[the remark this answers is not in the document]
joepetrowski: Yes, sounds good

parachains/common/src/xcm_config.rs
```diff
@@ -64,7 +64,7 @@ impl ShouldExecute for DenyReserveTransferToRelayChain {
 			message.0.iter().any(|inst| matches!(inst, ReserveAssetDeposited { .. }))
 		{
 			log::warn!(
-				target: "xcm::barrier",
+				target: "xcm::barriers",
```
KiChjang, line 67: `barrier` is fine, we don't need to change this.

parachains/runtimes/assets/statemine/tests/tests.rs
```diff
@@ -25,61 +28,78 @@ fn test_asset_xcm_trader() {
...
+			let asset_amount_extra = 100_u128;
+			let asset: MultiAsset =
+				(asset_multilocation.clone(), asset_amount_needed + asset_amount_extra).into();
+
+			let mut trader = <XcmConfig as xcm_executor::Config>::Trader::new();
+
+			// Lets buy_weight and make sure buy_weight does not return an error
+			match trader.buy_weight(bought, asset.into()) {
+				Ok(unused_assets) => {
+					// Check if return correct unused assets
```
KiChjang, line 85:
```suggestion
					// Check whether a correct amount of unused assets is returned
```

parachains/common/src/xcm_config.rs
bkontur: I thought that this is a typo, because everywhere (polkadot, cumulus repos) is used just `target: "xcm::barriers"`,
e.g. other `impl ShouldExecute for DenyReserveTransferToRelayChain` in Cumulus have: `target: "xcm::barriers"`:

https://github.com/paritytech/cumulus/blob/master/parachain-template/runtime/src/xcm_config.rs#L146
https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/testing/penpal/src/xcm_config.rs#L206

@KiChjang 
so, I guess, we should better keep it "in the same way" at least for Cumulus,
so, there are several possibilities, wdyt?
1. just revert this change, and keep different `xcm::barrier` and `xcm::barriers` in cumulus
2. just for Cumulus, use everywhere ``xcm::barrier`
3. also for Cumulus, use everywhere ``xcm::barriers`

parachains/runtimes/assets/statemine/tests/tests.rs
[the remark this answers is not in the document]
bkontur: continue in following issue: https://github.com/paritytech/cumulus/issues/1915

primitives/utility/src/lib.rs
bkontur: @girazoki 
ok, thank you for the information,

yes, there have to be database hit for checking MultiAsset (is_suffiecient,min_balance), but this is already done in `AssetFeeAsExistentialDepositMultiplier -> BalanceConverter::to_asset_balance`.

and according to what you say, then maybe `let multiassets: MultiAssets = payment.clone().into();` the whole **clone/join_iterators/resorting** stuff is really not needed and `payment.fungible_assets_iter().next()` will do the same job with also less cloning of other stuff here, maybe without need to use Matcher here, so then maybe renaming `TakeFirstAssetTrader` to `TakeFirstFungibleAssetTrader`.

I understand, that TakeFirstAssetTrader is supposed to be as-much-as-possible abstract implementation, but I think, there is a little "space/tension" between API `payment: xcm_executor::Assets (MutliAssets)` vs. `BuyExecution { fees: MultiAsset`, and this "space/tension" could be possibly fixed by the improved "first selection" algo iterating multiassets. And based on the fact, that we are already doing database hit (is_suffiecient,min_balance) and "BuyExecution has only one asset", so there wont be any performance degradation, but opposite, without **clone/join_iterators/resorting** it could be just improved and on the other hand the algo will be more complete, maybe for future multiassets support for `BuyExecution(MutliAssets)`

we will come to this maybe later in the following issue: https://github.com/paritytech/cumulus/issues/1915

parachains/common/src/xcm_config.rs
KiChjang: Okay, if cumulus has been using `barriers`, then let's keep it that way.

parachains/runtimes/assets/statemine/tests/tests.rs
```diff
@@ -302,3 +322,70 @@ fn test_that_buying_ed_refund_does_not_refund() {
...
+			AccountId::from(ALICE),
+			SessionKeys { aura: AuraId::from(sp_core::sr25519::Public::from_raw(ALICE)) },
+		)])
+		.build()
+		.execute_with(|| {
+			// Create a non-sufficient asset with specific existential deposit
+			let minimum_asset_balance = 1_000_000_u128;
+			assert_ok!(Assets::force_create(
+				RuntimeHelper::<Runtime>::root_origin(),
+				1,
```
KiChjang, line 341: Let's also extract this into a `local_asset_id` binding.

parachains/runtimes/assets/statemint/tests/tests.rs
```diff
@@ -71,3 +258,146 @@ fn test_asset_xcm_trader_does_not_work_in_statemine() {
...
+			AccountId::from(ALICE),
+			SessionKeys { aura: AuraId::from(sp_core::ed25519::Public::from_raw(ALICE)) },
+		)])
+		.build()
+		.execute_with(|| {
+			// We need root origin to create a sufficient asset
+			// We set existential deposit to be identical to the one for Balances first
+			assert_ok!(Assets::force_create(
+				RuntimeHelper::<Runtime>::root_origin(),
+				1,
```
KiChjang, line 277: Same as above.

parachains/runtimes/assets/statemine/tests/tests.rs
bkontur: @KiChjang 
there is a following issue: https://github.com/paritytech/cumulus/issues/1915,
so I will fix your comment there plus check all other tests

parachains/runtimes/assets/statemint/tests/tests.rs
bkontur: Same as above :)

KiChjang (approved):

dvdplm (approved):
lgtm modulo some concerns about the naming (getting very long!).

PR #1913, opened 2022-11-25
Use the `relay-headers-and-messages` command

serban300:
Resolves https://github.com/paritytech/parity-bridges-common/issues/1662

Updating the README in order to use the `relay-headers-and-messages` command.


bkontur (approved):
cooool, thank you :),
squash/merge please

PR #1976, opened 2022-12-09
Collectives runtime: weight and fee info rpc for Call 

muharem:
Adding the rpc calls to query weight and fee details of a Call.


bkchr (approved):

muharem:
bot merge

PR #1982, opened 2022-12-12
Move BridgeHubRococo and BridgeHubWococo primitives to bridges-ext

serban300:
Related to https://github.com/paritytech/parity-bridges-common/issues/1709

We need to use the same weights (and other constants) in the bridge hubs definitions in cumulus and in the relayer definition.

I'm proposing to move these primitives in cumulus and then move the implementation for the Rococo <> Wococo relayer also in cumulus based on these primitives. 

The other approach would be to defined the primitives in parity-bridges-common, but it seems unnatural to me since:
- the chains are not in the ownership of the parity-bridges-common repo
- parity-bridges-common will have dependecies on cumulus and cumulus will have dependencies on parity-bridges-common
- there might also be duplication between parity-bridges-common and cumulus

@acatangiu @bkontur @svyatonik what do you think ?


svyatonik:
We're only using cumulus to build our `rialto-parachain` runtime and node. We don't need it outside of those projects. Weights you mention are used by `substrate-relay` (to compute message delivery transaction limits) and we need to be able to build `substrate-relay` in `parity-bridges-common` repo, without updating cumulus references (otherwise we are getting exactly what you're talking about - "parity-bridges-common will have dependencies on cumulus and cumulus will have dependencies on parity-bridges-common"). So we need to have those weights in our chain primitives crates.

Yes - that's a duplication, but that's imo better than dancing with cross-repo updates (e.g.). Integrity tests (which may need some improvements, though - e.g. we may have some subcommand that connects to nodes and gets actual information, comparing it to hardcoded version) are actually for that.

Feel free to argue, though :)

serban300:
Sorry, I thought that we could move the Rococo-Wococo relayers implementation in cumulus as well, and use parity-bridges-common only as a library. I thought that everything we need is here and that it would make sense. But I missed that only the bridge hubs are defined here. Rococo and Wococo are defined in the polkadot project and we need info about them as well.

Yes, having this definitions in parity-bridges-common is the only way

PR #1986, opened 2022-12-12
[ci] Add variable to disable selected jobs

alvicsam:
cc https://github.com/paritytech/ci_cd/issues/694


joao-paulo-parity (approved):

rcny (approved):

sergejparity (approved):

.gitlab-ci.yml
```diff
@@ -98,6 +98,7 @@ variables:
 .docker-env:
   image:                           "${CI_IMAGE}"
   before_script:
+    - if echo "$CI_DISABLED_JOBS" | grep -E "^$CI_JOB_NAME$"; then echo "The job has been cancelled in CI settings"; exit 0; fi
```
altaua, line 101: This feels brittle given $CI_JOB_NAME is allowed to contain characters such as [] that'd have a special meaning for grep.

altaua: grep -xF "$CI_JOB_NAME" would avoid that.

PR #1988, opened 2022-12-13
Change CGP Topic Labels

joepetrowski:
Changes based on:

- Removed a "Statemint" label, because this will not scale as we add more parachains like Collectives, Bridge Hub, Staking, etc.
- Changed "CGP" to "System Parachains", as we are dropping the CGP name (worth a discussion on its own).
- Added a "Bridges" label. Why not drop this like Statemint? I see this as similar to "smart contracts", in that the issue will probably be routed to another team (e.g. Parity Bridges, Snowfork) and the changes will probably need to happen in another repo (contracts: Substrate or ink, bridges: parity-bridges, snowbridge,...).

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -74,15 +74,15 @@ It contains 65 labels:
 ### T labels: Topics - to be used in combination with other labels
 - `T0-node`: *This PR/Issue is related to the topic “node”.*
 - `T1-runtime`: *This PR/Issue is related to the topic “runtime”.*
-- `T10-release`: *This PR/Issue is related to topics touching the release notes.*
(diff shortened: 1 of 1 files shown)


alvicsam:
fyi, the file was moved [here](https://github.com/paritytech/labels/blob/main/docs/doc_cumulus.md)

the-right-joyce (changes requested):
Please, add the changes to this document:
https://github.com/paritytech/labels/blob/main/ruled_labels/specs_cumulus.yaml

joepetrowski:
@the-right-joyce will the Labels repo change apply here automatically? I.e., can we close this?

the-right-joyce:
yes we can close it :)

PR #2054, opened 2023-01-03
Add benchmarked weights to pallet parachain-system

nbaztec:
Adds benchmarked weights to pallet `parachain-system`


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2217423

pallets/parachain-system/src/benchmarking.rs
```diff
@@ -0,0 +1,124 @@
...
+use frame_system::{EnsureNone, EnsureRoot, EnsureSigned, EventRecord, RawOrigin};
+use sp_core::H256;
+use sp_std::{collections::btree_set::BTreeSet, prelude::*};
+use sp_trie::StorageProof;
+
+const SEED: u32 = 0;
+
+benchmarks! {
+	set_validation_data_no_messages {
+		// root and proof are obtained via `RelayStateSproofBuilder::default().into_state_root_and_proof()`
```
ggwpez, line 39: This is the same problem I had here https://github.com/paritytech/cumulus/pull/1709  
Do you plan on making the mocked proof builder no-std compatible?

nbaztec: I looked into it, but unfortunately seemed somewhat of an effort with my limited understanding and some dependent crates themselves being no-std :(

nbaztec: Or would you maybe have an estimate how long it would take? Maybe I can try anyway and ask for help if I'm stuck somewhere.

ggwpez: The base version already works fine. But updating all to no-std was annoying. I can take another look at it though.  
Currently there is no timeline. If it does not work we may merge the solution with hard-coded proof blobs.

crystalin:
Any update on this ?

girazoki:
We used hard-coded proofs also in Tanssi to benchmark state-proof related pallets:

https://github.com/moondance-labs/tanssi/blob/master/pallets/author-noting/src/mock_proof.rs

It's not ideal and probably does not cover some proof-depth related stuff, but for now it was good enough (cc @tmpolaczyk).

PR #2068, opened 2023-01-07
Remove the `RelaychainClient` trait

bkchr:
It was just some historical trait that isn't really required anymore. Besides that this pr re-exports types that are being used by the relay chain interface to make its usage easier.

### DIFF
--- a/client/consensus/common/src/parachain_consensus.rs
+++ b/client/consensus/common/src/parachain_consensus.rs
@@ -14,7 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
 
-use async_trait::async_trait;
 use sc_client_api::{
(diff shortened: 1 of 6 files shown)


client/consensus/common/src/parachain_consensus.rs
```diff
@@ -210,7 +190,7 @@ async fn follow_new_best<P, R, Block, B>(
 						h,
 						&*parachain,
 						&mut unset_best_header,
-						recovery_chan_tx.clone(),
+						recovery_chan_tx.as_mut(),
```
bkchr, line 193: As we clonded the sender every time before, the `try_send` would have always succeeded as each sender instance gets at least one slot in the channel.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2235554

melekes (approved):
👍

koute (approved):

PR #2069, opened 2023-01-09
Refactor `validate_block`

bkchr:
This pull request changes the `validate_block` implementation. One of the key changes are that we free data structures as early as possible. The memory while validating the block is scarce and we need to give as much as possible to the actual execution of the block. Besides that the pr moves the validation of the `validation_data` into the `validate_block` implementation completely instead of using this machinery with putting the data into some global variable that would then be read while executing the block. There are also some new docs to explain the internals of `validate_block`.

### DIFF
--- a/pallets/parachain-system/src/validate_block/implementation.rs
+++ b/pallets/parachain-system/src/validate_block/implementation.rs
@@ -19,15 +19,21 @@
 use frame_support::traits::{ExecuteBlock, ExtrinsicCall, Get, IsSubType};
 use sp_runtime::traits::{Block as BlockT, Extrinsic, HashFor, Header as HeaderT};
 
-use sp_io::KillStorageResult;
-use sp_std::prelude::*;
+use cumulus_primitives_core::{
+	relay_chain::v2::Hash as RHash, ParachainBlockData, PersistedValidationData,
+};
+use cumulus_primitives_parachain_inherent::ParachainInherentData;
 
-use polkadot_parachain::primitives::{HeadData, ValidationParams, ValidationResult};
+use polkadot_parachain::primitives::{
+	HeadData, RelayChainBlockNumber, ValidationParams, ValidationResult,
+};
 
 use codec::{Decode, Encode};
 
 use sp_core::storage::{ChildInfo, StateVersion};
 use sp_externalities::{set_and_run_with_externalities, Externalities};
+use sp_io::KillStorageResult;
+use sp_std::{mem, prelude::*};
 use sp_trie::MemoryDB;
 
 type TrieBackend<B> = sp_state_machine::TrieBackend<MemoryDB<HashFor<B>>, HashFor<B>>;
@@ -38,7 +44,32 @@ fn with_externalities<F: FnOnce(&mut dyn Externalities) -> R, R>(f: F) -> R {
 	sp_externalities::with_externalities(f).expect("Environmental externalities not set.")
 }
 
-/// Validate a given parachain block on a validator.
+/// Validate the given parachain block.
+///
+/// This function is doing roughly the following:
+///
+/// 1. We decode the [`ParachainBlockData`] from the `block_data` in `params`.
+///
+/// 2. We are doing some security checks like checking that the `parent_head` in `params`
+/// is the parent of the block we are going to check. We also ensure that the `set_validation_data`
+/// inherent is present in the block and that the validation data matches the values in `params`.
+///
+/// 3. We construct the sparse in-memory database from the storage proof inside the block data and
+/// then ensure that the storage root matches the storage root in the `parent_head`.
+///
+/// 4. We replace all the storage related host functions with functions inside the wasm blob.
+/// This means instead of calling into the host, we will stay inside the wasm execution. This is
+/// very important as the relay chain validator hasn't the state required to verify the block. But
+/// we have the in-memory database that contains all the values from the state of the parachain
+/// that we require to verify the block.
+///
+/// 5. We are going to run `check_inherents`. This is important to check stuff like the timestamp
+/// matching the real world time.
+///
+/// 6. The last step is to execute the entire block in the machinery we just have setup. Executing
+/// the blocks include running all transactions in the block against our in-memory database and
+/// ensuring that the final storage root matches the storage root in the header of the block. In the
+/// end we return back the [`ValidationResult`] with all the required information for the validator.
 #[doc(hidden)]
 pub fn validate_block<
 	B: BlockT,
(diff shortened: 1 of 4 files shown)


pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -153,34 +181,75 @@ where
...
+		.iter()
+		// Inherents are at the front of the block and are unsigned.
+		//
+		// If `is_signed` is returning `None`, we keep it safe and assume that it is "signed".
+		// We are searching for unsigned transactions anyway.
+		.take_while(|e| !e.is_signed().unwrap_or(true))
+		.filter_map(|e| e.call().is_sub_type())
+		.find_map(|c| match c {
+			crate::Call::set_validation_data { data: validation_data } =>
+				Some(validation_data.clone()),
```
koute, line 229: Is this `clone` necessary?

koute (approved):

michalkucharczyk (approved):

pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -52,19 +83,30 @@ where
...
+
 	let (header, extrinsics, storage_proof) = block_data.deconstruct();
 
 	let head_data = HeadData(header.encode());
 
 	let block = B::new(header, extrinsics);
-	assert!(parent_head.hash() == *block.header().parent_hash(), "Invalid parent hash",);
+	assert!(parent_head.hash() == *block.header().parent_hash(), "Invalid parent hash");
+
+	let inherent_data = extract_parachain_inherent_data(&block);
```
melekes, line 102: dummy question: should we drop `block.entrinsics` after this?

melekes (approved):
👍 (code-wise; don't know enough to reason about business logic)

pallets/parachain-system/src/validate_block/implementation.rs
bkchr: No, we still need to execute them

pallets/parachain-system/src/validate_block/mod.rs
```diff
@@ -27,7 +27,32 @@ mod tests;
...
+/// `block_data` and `head_data` are represented as [`bytes::Bytes`] to make them reuse
+/// the memory of the input parameter of the exported `validate_blocks` function.
+///
+/// The layout of this type must match exactly the layout of
+/// [`ValidationParams`](polkadot_parachain::primitives::ValidationParams) to have the same
+/// SCALE encoding.
+#[derive(codec::Decode)]
+#[cfg_attr(feature = "std", derive(codec::Encode))]
+#[doc(hidden)]
+pub struct MemoryOptimizedValidationParams {
```
koute, line 53: Any reason why not just directly change the `ValidationParams` to hold `Bytes` instead of `Vec<u8>`s?

(This is fine by me, I'd just like to know why because I haven't checked myself how annoying it would be to change `ValidationParams`.)

pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -77,39 +77,41 @@ pub fn validate_block<
...
 	let inherent_data = extract_parachain_inherent_data(&block);
 
 	validate_validation_data(
 		&inherent_data.validation_data,
-		params.relay_parent_number,
-		params.relay_parent_storage_root,
-		params.parent_head,
+		relay_parent_number,
+		relay_parent_storage_root,
+		parent_head,
```
koute, line 110: As I've said in my previous comment, this could be dropped earlier if we'd compare the hashes instead of doing a bytewise comparison of `parent_head` to `validation_data.parent_head` (which, assuming the hash is a cryptographic hash, would be safe).

I'm not sure if it makes a difference (it only would if there could have conceivably be an allocation before this point which could be put in the same bucket as the full deallocated blob with the parameters, which I'd guess there isn't, so it most likely won't make a difference) and I'm fine with it as-is, but just like to make sure that this is an explicit decision and you didn't accidentally skip over my comment. (:

pallets/parachain-system/src/validate_block/implementation.rs
```diff
@@ -77,39 +77,41 @@ pub fn validate_block<
...
-		B::Header::decode(&mut &params.parent_head.0[..]).expect("Invalid parent head");
-
-	// We decoded the block data and don't need it anymore. So, we release it to free the memory.
-	mem::drop(params.block_data);
+	let parent_header =
+		codec::decode_from_bytes::<B::Header>(parent_head.clone()).expect("Invalid parent head");
 
 	let (header, extrinsics, storage_proof) = block_data.deconstruct();
 
 	let head_data = HeadData(header.encode());
```
koute, line 99: I think this allocation could be delayed? (e.g. by doing `let head_data = HeadData(block.header().encode())` since `block` is alive until the very end).

koute (approved):
Looks good to me.

It's a little tricky to review as it's somewhat speculative whether a given change will actually help anything (beyond getting rid of the obvious `clone`s), or whether it's just cargo culting. (E.g. delaying some allocations later may or may not prevent an extra unnecessary bucket to be created by the allocator, but without actually knowing the exact sizes of the allocations and tracing the execution it's essentially impossible to tell.)

We could revisit this once I get Bytehound hooked into the WASM allocator so that we will be able to gather some concrete data and see exactly what would help and what wouldn't.

pallets/parachain-system/src/validate_block/mod.rs
bkchr: I would have needed to change Polkadot for some implementation/optimization detail of the validation function in Cumulus. Then there would need to be some kind of comment that we don't change this etc. I mean it is possible, but then I decided to keep the changes local and observable. Also prevents that this accidentally is reverted or whatever.

pallets/parachain-system/src/validate_block/implementation.rs
koute: > The problem being that hash could be implemented in any way (will probably not happen, but possible as it is a trait method).

Well, we could always just calculate a concrete hash that we know is secure. :P

But yeah, it's fine as-is, and might not be worth it. This is just my OCD pointing out the opportunity. :P

bkchr:
> It's a little tricky to review as it's somewhat speculative whether a given change will actually help anything (beyond getting rid of the obvious `clone`s), or whether it's just cargo culting. (E.g. delaying some allocations later may or may not prevent an extra unnecessary bucket to be created by the allocator, but without actually knowing the exact sizes of the allocations and tracing the execution it's essentially impossible to tell.)

Good points ;) I also don't really know, but lets be nice and clean up our stuff as early as possible ;) 

But I also have done some other refactorings, so this is not just about random changes :D

pallets/parachain-system/src/validate_block/mod.rs
koute: Since now we're using `decode_from_bytes` on those fields I think we'd be safe from an accidental revert, but yeah, as long as there's a test that the two don't go out of sync it's fine.

bkchr:
bot merge

pallets/parachain-system/src/validate_block/mod.rs
bkchr: > Since now we're using `decode_from_bytes` on those fields I think we'd be safe from an accidental revert, but yeah, as long as there's a test that the two don't go out of sync it's fine.

What I meant by accidental revert is that we "loose" the memory optimizations, nothing really breaking.

PR #2091, opened 2023-01-15
Remove assumption that there is a local bootnode in RPC collator

skunert:
Initially, when developing the RPC node, we assumed that the relay chain node should run locally and added the local listen address of the specified relay chain node automatically as bootnode.

In our own CI and many other setups, I do not think this assumption holds and only leads to confusion, so let's remove it. Users should manually add the correct bootnodes for the relay chain.


bkchr (approved):
Does this solves the CI issues?

skunert:
No, I think there are other issues. I will take a look at the pov-recovery test again.

davxy (approved):

sergejparity (approved):

melekes (approved):
👍

PR #2101, opened 2023-01-17
update from 'X-labels' to 'T-labels'

the-right-joyce:
X-labels are no longer existing


chevdor:
Didn't we switch already to `ruled_labels` in cumulus ?
The script you are fixing seems to be the old one.

alvicsam:
This script doesn't work anymore. You need to make changes in https://github.com/paritytech/labels/blob/main/ruled_labels/specs_cumulus.yaml

Issue #1669, opened 2022-09-22
[Fix] Hardcoded weight for initiate_teleport


ruseinov:
Once the XCM v3 is out we need to replace that by proper logic.

Here's one example [occurrence](https://github.com/paritytech/cumulus/blob/3f11a7a6cfe357955050a6e3b8d4201cd1fff2e5/parachains/runtimes/assets/statemint/src/weights/xcm/mod.rs#L149).

Most likely it will mean that we will fall back to `MultiAssetFilter.weigh_multi_assets` logic and would have to modify it to match other invariants of `Wild`. Currently there are only `Definite` and `Wild`, Wild only supports either `All` or `AllOf`, none of which help us with weight calculation, because we want to know exactly the amount of assets being sent to calculate the weight. 

This also means we have to make sure that `do_teleport_assets` method and other methods that are using `Wild(All)` are revisited to include the actual amount of assets (max_assets, if you will) that were present at the beginning of the operation.

cc @KiChjang @gilescope @joepetrowski

PR #2102, opened 2023-01-17
[Enhancement] Use XCM V3 for initiate_teleport weight calc

ruseinov:
Fixes https://github.com/paritytech/cumulus/issues/1669


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2277927

gavofyork (approved):

gilescope (approved):

parachains/runtimes/assets/statemine/src/weights/xcm/mod.rs
```diff
@@ -35,7 +35,12 @@ impl WeighMultiAssets for MultiAssetFilter {
 		match self {
 			Self::Definite(assets) =>
 				weight.saturating_mul(assets.inner().into_iter().count() as u64),
-			Self::Wild(_) => weight.saturating_mul(MAX_ASSETS as u64),
+			Self::Wild(asset) => match asset {
+				All => weight.saturating_mul(MAX_ASSETS as u64),
+				AllOf { .. } => weight,
```
KiChjang, line 40: I'm actually still uncertain about this -- comments on AllOfCounted says that each individual instance of an NFT is considered a different class of asset, and so wouldn't it make more sense to use the number of NFT instances as an upper bound here?

ruseinov: Hmm, good question. Let me check it out and come back to you.

ruseinov: Yeah, I think this makes sense. I'm not sure though where the NFT instance number should be coming from in this case.

KiChjang: So this requires you to actually look inside the holding register and determine just how many `AssetInstance`s there are... We could match against `fun` in `AllOf` and see if it's a `NonFungible`, and if so, it would basically be `weight * MaxAssetsIntoHolding::get()`, otherwise it's just `weight`.

ruseinov: Thanks!

ruseinov: @KiChjang One thing that disturbs me a bit is the comment below, should we then do something like `weight * MaxAssetsIntoHolding::get() * 2` to be safe? 

```rust
	/// The maximum number of assets we target to have in the Holding Register at any one time.
	///
	/// NOTE: In the worse case, the Holding Register may contain up to twice as many assets as this
	/// and any benchmarks should take that into account.
	type MaxAssetsIntoHolding: Get<u32>;
```

ruseinov:
bot rebase

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2370797

ruseinov:
bot merge

Issue #2142, opened 2023-01-25
Add retry mechanism for pov-recovery


skunert:
While investigating flaky pov-recovery test behaviour, I found that it sometimes fails because data is still Unavailable on the relay chain.
[Example Test](https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2306919)
[Node Logs](https://grafana.parity-mgmt.parity.io/explore?orgId=1&left=%7B%22datasource%22:%22PCF9DACBDF30E12B3%22,%22queries%22:%5B%7B%22expr%22:%22%7Bpod%3D~%5C%22zombie-3b5f511c3fbcc09fc107f1063ac1b153%2Ftwo%5C%22%7D%20%7C~%20%6030311%60%22,%22refId%22:%22A%22,%22range%22:true,%22editorMode%22:%22builder%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221674548407166%22,%22to%22:%221674549777386%22%7D%7D)

We are starting recovery as soon as we see a candidate pending availability. Since availability is not guaranteed, it can happen that the candidate is not yet available on the relay chain. We should have a backup mechanism that retries failed recovery requests.

rphmeier:
It should maybe instead be that availability requests are queued until requisite bitfields are posted to the relay-chain. Although things might be available already, despite not being posted to the relay chain yet. Watching the offchain bitfield gossip would also be a way to learn about this faster.

rphmeier:
Also, one thing I am curious about is at what point we actually initiate PoV recovery.

In principle, PoV recovery should only be necessary on collators in the case of collator block withholding attacks (when the collator sends the PoV to validators but doesn't share it with others). What I am getting at is that we shouldn't need PoV recovery unless there is either a malicious collator or the sync code isn't fetching the block properly for some reason.

In the issues we've seen in test networks, malicious collators are almost certainly not the underlying issue.

skunert:
> In principle, PoV recovery should only be necessary on collators in the case of collator block withholding attacks (when the collator sends the PoV to validators but doesn't share it with others). What I am getting at is that we shouldn't need PoV recovery unless there is either a malicious collator or the sync code isn't fetching the block properly for some reason.

[...]

PR #2143, opened 2023-01-25
Increase recovery delay for pov-recovery

skunert:
Increase the time window before recovering a block via pov-recovery. Should alleviate the test flakyness caused by https://github.com/paritytech/cumulus/issues/2142 before implementing a proper retry.

This usage here is fine since it only recovers included blocks, these should be available.
https://github.com/paritytech/cumulus/blob/977da03c2d6cb833f9d1de2688b14efb5dadbdb9/client/consensus/common/src/parachain_consensus.rs#L41-L42


client/service/src/lib.rs
```diff
@@ -121,7 +121,10 @@ where
 		overseer_handle.clone(),
 		// We want that collators wait at maximum the relay chain slot duration before starting
 		// to recover blocks.
-		RecoveryDelay { min: core::time::Duration::ZERO, max: relay_chain_slot_duration },
+		RecoveryDelay {
+			min: relay_chain_slot_duration.checked_div(2).expect("2 is larger than 0; qed"),
```
bkchr, line 125:
```suggestion
			min: relay_chain_slot_duration / 2,
```

bkchr: And some comment here why we are doing this, something like "we want to wait for half to the next slot to have something available bla". Basically what we talked about.

bkchr (approved):

davxy (approved):

client/service/src/lib.rs
```diff
@@ -120,8 +120,9 @@ where
 	let pov_recovery = PoVRecovery::new(
 		overseer_handle.clone(),
 		// We want that collators wait at maximum the relay chain slot duration before starting
-		// to recover blocks.
-		RecoveryDelay { min: core::time::Duration::ZERO, max: relay_chain_slot_duration },
+		// to recover blocks. Additionally, we wait at least half the slot time to give the
+		// relay chain the chance to increase availability.
+		RecoveryDelay { min: relay_chain_slot_duration / 2, max: relay_chain_slot_duration },
```
michalkucharczyk, line 125: Why increasing from 0 to `duration/2` helps?

Because there is only one try?
And trying later has higher chance of succes?

test/service/src/lib.rs
```diff
@@ -385,10 +385,7 @@ where
 			para_id,
 			relay_chain_interface,
 			import_queue: import_queue_service,
-			// The slot duration is currently used internally only to configure
-			// the recovery delay of pov-recovery. We don't want to wait for too
-			// long on the full node to recover, so we reduce this time here.
-			relay_chain_slot_duration: Duration::from_millis(6),
+			relay_chain_slot_duration: Duration::from_secs(6),
```
michalkucharczyk, line 388: Why 6? Could we use a constant defined somewhere else?

michalkucharczyk (approved):

client/service/src/lib.rs
bkchr: This here is just a first step until we have it properly fixed: https://github.com/paritytech/cumulus/issues/2142

> Because there is only one try?

Currently yes

skunert: Yes, you can find some additional details in the issue I linked.

Basically the chance increases that the block is available on the relay chain. The relay chain just had more time to distribute the chunks.

I plan to implement a second try too, but not in this PR.

test/service/src/lib.rs
bkchr: Currently it is the de-facto standard slot duration. At some point we may need to revisit this.

skunert:
Please no merging on this yet, ironically pov test fails consistently here.

skunert:
Closing in favor of https://github.com/paritytech/cumulus/pull/2164

PR #2157, opened 2023-01-31
Use `Message Queue` pallet for DMP queue dispatch

ggwpez:
WIP


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2338037

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2338062

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2338100

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2350045

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2357296

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2357322

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2364313

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2372416

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2430747

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2431049

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2431361

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2503622

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2588525

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2595641

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2601118

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2601333

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2602860

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2603209

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2603409

ggwpez:
@paritytech/parachains-core could someone please take a look?  

Gav only commented
>it might be sensible to extend the message queue pallet (or have it provide a config item to allow such extensibility) to be able to handle the suspension/QoS logic and avoid the complexity on the right side of the schematic.

parachain-template/runtime/src/lib.rs
```diff
@@ -379,26 +383,60 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
 
 impl parachain_info::Config for Runtime {}
 
+parameter_types! {
+		pub MessageQueueServiceWeight: Weight = Perbill::from_percent(10) *
```
kianenigma, line 387:
```suggestion
	pub MessageQueueServiceWeight: Weight = Perbill::from_percent(10) *
```

parachain-template/runtime/src/lib.rs
```diff
@@ -489,7 +527,8 @@ construct_runtime!(
 		XcmpQueue: cumulus_pallet_xcmp_queue = 30,
 		PolkadotXcm: pallet_xcm = 31,
 		CumulusXcm: cumulus_pallet_xcm = 32,
-		DmpQueue: cumulus_pallet_dmp_queue = 33,
+		// RIP DmpQueue 33,
```
kianenigma, line 530: Probably no need to mention it here as it is just a "template".

parachains/common/src/process_xcm_message.rs
```diff
@@ -0,0 +1,118 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Implementation of `ProcessMessage` for an `ExecuteXcm` implementation.
+
+use codec::{Decode, FullCodec, MaxEncodedLen};
+use cumulus_primitives_core::{AggregateMessageOrigin, ParaId};
+use frame_support::{
```
kianenigma, line 21: Why is this called `AggregateMessageOrigin`, namely what is `Aggregate` supposed to mean?

kianenigma: probably because of https://github.com/paritytech/polkadot/blob/b749009549428344e6aec63d4a581a6a6e688af6/runtime/parachains/src/inclusion/mod.rs#L240?

ggwpez: The idea is the keep it extensible since we soon also want `Loopback` and `Bridged(…)` origins. So it aggregates all of them.

primitives/core/src/lib.rs
```diff
@@ -72,6 +71,42 @@ impl From<MessageSendError> for &'static str {
 	}
 }
 
+/// The origin of an inbound message.
+#[derive(Encode, Decode, MaxEncodedLen, Clone, Eq, PartialEq, TypeInfo, Debug)]
+pub enum AggregateMessageOrigin {
```
vgeddes, line 76: Just an FYI, In the BridgeHub system parachain, we are also wanting to use the message queue for buffering messages before bridging them to Ethereum. Busy implementing this in https://github.com/Snowfork/snowbridge/pull/851.

In our case, this our aggregate origin looks like this:
```
pub enum AggregateMessageOrigin {
 	#[codec(index = 0)]
 	Parachain(ParaId),
 }
```

So each parachain sending messages to ethereum has its own queue.

Hoping we can add an enum variant to cater for this use case.

vgeddes: Also just a note to myself (for future reference):

For the BridgeHub system parachain, we'd like for the message-queue pallet to be registered after our Ethereum-specific pallets, to ensure it's on_initialize hander is called after our on_initialize handlers.

This is to ensure our message processing queues are correctly initialised before message-queue starts doing its thing.

ggwpez: Nice. Yea LMK if you have feedback on the MQ pallet. 

Currently we use the MQ pallet for inbound dispatch queues (not delivery) like DMP/XCMP on paras and UMP dispatch on the relay. But I guess it can also be used for delivery, so we could add an `OutboundSnowbridged(ParaId)` or similar.  

How do you currently handle incoming snowbridged messages? Are they executed immediately or queued as well?

gavofyork:
> @paritytech/parachains-core could someone please take a look?
> 
> Gav only commented
> 
> > it might be sensible to extend the message queue pallet (or have it provide a config item to allow such extensibility) to be able to handle the suspension/QoS logic and avoid the complexity on the right side of the schematic.

I wrote that code, so it's probably going to need to be us who perform the refactoring.

ggwpez:
> I wrote that code, so it's probably going to need to be us who perform the refactoring.

Do you think we should do that now or as follow up? So far it seems to work, just a bit complicated...

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2947445

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2947589

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2952467

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2954606

primitives/core/src/lib.rs
vgeddes: Excuse the tardy reply, I was OOO for most of this week.

> Currently we use the MQ pallet for inbound dispatch queues (not delivery) like DMP/XCMP on paras and UMP on the relay. 

This is fine and I still think still fits the design envisioned for MessageQueue. Strictly speaking, in the Polkadot->Ethereum direction, when messages from origin parachains reach bridge hub, they certainly haven't been finally delivered yet, and still need to be processed before they are suitable for delivery to Ethereum. Buffering is needed, as our bridge can only process a finite number of messages per block, in order for the weights/benchmarking side to work out.

> How do you currently handle incoming snowbridged messages? Are they executed immediately or queued as well?

Inbound messages from ethereum are immediately routed to a sibling parachain via HRMP - no queuing or buffering or anything like that. For reference, see https://github.com/Snowfork/snowbridge/blob/bd1773d2e983e3c558e61db5b4348ecc5fe922c4/parachain/pallets/inbound-queue/src/lib.rs#L199.

So yeah, is standard XCMP.

> PS: We can also simplify the traits in the parachain-system pallet, as to not force every parachain to adopt the same AggregateMessageOrigin.

Yes please, our enum variant in `AggregateMessageOrigin` very specific and probably won't make much sense for parachains other than BridgeHub.

gavofyork:
> > I wrote that code, so it's probably going to need to be us who perform the refactoring.
> 
> Do you think we should do that now or as follow up? So far it seems to work, just a bit complicated... @gavofyork

I think probably do it with this PR.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3088087

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3088274

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3097050

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3104542

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3264994

primitives/core/src/lib.rs
```diff
@@ -78,6 +77,42 @@ impl From<MessageSendError> for &'static str {
...
+	/// This is used by the HRMP queue.
+	Sibling(ParaId),
+}
+
+impl From<AggregateMessageOrigin> for xcm::v3::MultiLocation {
+	fn from(origin: AggregateMessageOrigin) -> Self {
+		match origin {
+			AggregateMessageOrigin::Loopback => MultiLocation::here(),
+			AggregateMessageOrigin::Parent => MultiLocation::parent(),
+			AggregateMessageOrigin::Sibling(id) => Junction::Parachain(id.into()).into(),
```
svyatonik, line 100:
```suggestion
			AggregateMessageOrigin::Sibling(id) => MultiLocation::new(1, Junction::Parachain(id.into())),
```

Should probably changed as suggested?

ggwpez: Yep looks better. Thanks!

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3268943

pallets/xcmp-queue/src/lib.rs
```diff
@@ -447,21 +328,25 @@ pub struct QueueConfigData {
...
 }
 
 impl Default for QueueConfigData {
 	fn default() -> Self {
+		#![allow(deprecated)]
 		Self {
-			suspend_threshold: 2,
-			drop_threshold: 5,
-			resume_threshold: 1,
+			suspend_threshold: 200,
```
svyatonik, line 347: Just wanted to clarify to understand it better (also relevant for bridges).

You've said (in the PR description): "Bumped the default suspension, drop and resume thresholds". But is it really a bump? It is hard to compare, because previously they were measured in pages and now it is individual XCM messages. Page size is `102_400` bytes e.g. at asset hubs (unless I miss something). So previously pallet was fine with `2 * 102_400 = 204_800` bytes of concatenated XCM messages before suspension has been activated. Now it suspends when there are `200` XCM messages enqueued. So if that's really a bump, then the assumption is that XCM messages are larger than `204_800 / 200 = 1024` bytes, right? If they're smaller than `1Kb` (average), then it isn't really a bump :)

pallets/xcmp-queue/src/lib.rs
```diff
@@ -472,6 +357,22 @@ impl Default for QueueConfigData {
...
 }
 
+impl QueueConfigData {
+	/// Validate all assumptions about `Self`.
+	///
+	/// Should be called prior to accepting this as new config.
+	pub fn validate<T: crate::Config>(&self) -> sp_runtime::DispatchResult {
+		if self.resume_threshold < self.suspend_threshold &&
+			self.suspend_threshold <= self.drop_threshold &&
+			self.resume_threshold > 0
```
svyatonik, line 367: Should `resume_threshold` also be lesser than `suspend_threshold` (i.e. `1..self.suspend_threshold.contains(self.resume_threshold)`)?

pallets/xcmp-queue/src/lib.rs
```diff
@@ -941,84 +510,160 @@ impl<T: Config> Pallet<T> {
...
+				VersionedXcm::<T::RuntimeCall>::decode_with_depth_limit(MAX_XCM_DECODE_DEPTH, data)
+					.map_err(|_| ())?;
+			let bounded = xcm.encode().try_into().map_err(|_| ())?;
+			encoded_xcms.push(bounded);
+		}
+		Ok(encoded_xcms)
+	}
+}
+
+impl<T: Config> OnQueueChanged<ParaId> for Pallet<T> {
```
svyatonik, line 566: I guess it's probably coming in next commits, but just wanted to mention that it should probably be changed to:
```rust
impl<T: Config> OnQueueChanged<cumulus_primitives_core::AggregateMessageOrigin> for Pallet<T> {
```

and then used in `pallet_message_queue` configuration at all runtimes:
```rust
type QueueChangeHandler = XcmpQueue;
```

Now the `QueueChangeHandler` is set to `()` everywhere, which means that backpressure mechanism will never be activated.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -941,84 +510,160 @@ impl<T: Config> Pallet<T> {
...
+			} else if let Err(err) = suspended_channels.try_insert(para) {
+				log::error!("Too many channels suspended; cannot suspend sibling {:?}: {:?}; further messages may be dropped.", para, err);
+			} else {
+				<InboundXcmpSuspended<T>>::put(suspended_channels);
+			}
+		}
+	}
+}
+
+impl<T: Config> QueuePausedQuery<ParaId> for Pallet<T> {
```
svyatonik, line 597: Maybe also worth changing to:
```rust
impl<T: Config> QueuePausedQuery<cumulus_primitives_core::AggregateMessageOrigin> for Pallet<T> {
```

Then we may remove the `NarrowToSiblings` and use the `XcmpQueue` directly:
```rust
type QueuePausedQuery = XcmpQueue;
```

svyatonik (commented):
Left some questions + feedback based on my (successful) attempt to use this stuff in bridges. Thank you for working on that! *(will continue review a bit later)*

pallets/xcmp-queue/src/lib.rs
ggwpez: WDYM? I check it right here `self.resume_threshold < self.suspend_threshold`.  
And then further down we have the resume condition like `suspended && count <= resume_threshold as u64`.

pallets/xcmp-queue/src/lib.rs
ggwpez: Ouch 🤦‍♂️ good catch!  

The `XcmpQueue` pallet is only ever able to handle `ParaId`s , no the full `AggregateMessageOrigin` so I am re-using the narrowing approach and now set:  
`type QueueChangeHandler = NarrowOriginToSibling<XcmpQueue>;`

pallets/xcmp-queue/src/lib.rs
ggwpez: But then we have the problem that there is an "impossible" case in the code, where a non-sibling origin is passed to the function.  
To rule this out directly, i think its better to have an adapter á la `NarrowToSiblings`.

pallets/xcmp-queue/src/lib.rs
svyatonik: Thresholds are now checked from the  `pallet-message-queue`' `OnQueueChanged` implementation, where the count is the count of messages, not pages. So yep - it is not measured in messages, not pages. I was mostly interested in seeing actual configuration thresholds here, so all looks fine to me.

pallets/xcmp-queue/src/lib.rs
svyatonik: :facepalm: Of course, you're right - sorry for the noise

pallets/xcmp-queue/src/lib.rs
svyatonik: Ok - both options LGTM :) Thanks!

pallets/xcmp-queue/src/lib.rs
svyatonik: Ok, this works too :)

pallets/xcmp-queue/src/lib.rs
```diff
@@ -941,84 +510,173 @@ impl<T: Config> Pallet<T> {
...
 			}
 		});
 	}
+
+	fn enqueue_xcmp_messages(
+		sender: ParaId,
+		mut xcms: Vec<BoundedVec<u8, MaxXcmpMessageLenOf<T>>>,
+		meter: &mut WeightMeter,
+	) {
+		if meter
```
svyatonik, line 523: I've been looking through the code couple of times and those drops is what worries me most. In previous implementation, we've only been using the `meter` to process messages. And now it also accounts pushing to the queue for later processing. While this is technically correct, it (it theory) can lead to accidental message drops.

For example, in `enqueue_xcmp_messages` benchmark, we are using maximal possible message size. So the weight returned is the maximal possible. But what if instead of largest possible messages, the sending chain sends a bunch of smallest possible messages? Then, it the size effect is too large, then we can drop it here, because `T::WeightInfo::enqueue_xcmp_messages(<nuymbe-of-small-messages>)` is large enough to fit into the `meter`.

I know that it is unlikely to happen, because we have `1/4` of block to enqueue messages. But if there are many inbound HRMP channels (I could imaging bridge hub will have many) and a lot of inbound messages, it can happen.

I also think that we probably can ignore the overweight problem here and simply return actual used weight, because IIUC it is called from the inherent (`set_validation_data`).

That said, I'm not a pro here - it just caught my eye :)

pallets/parachain-system/src/lib.rs
```diff
@@ -175,24 +186,32 @@ pub mod pallet {
...
-		/// The message handler that will be invoked when messages are received via DMP.
-		type DmpMessageHandler: DmpMessageHandler;
+		/// Queues inbound downward messages for delayed processing.
+		///
+		/// All inbound DMP messages from the relay are pushed into this. The handler is expected to
+		/// eventually process all the messages that are pushed to it.
+		///
+		/// There is a slightly caveat here since the only variant of [`AggregateMessageOrigin`]
+		/// that is ever used is the `Parent` variant. It would therefore be better to use a more
+		/// constrained trait instead such that no incorrect origin can be used.
```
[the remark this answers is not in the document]
ggwpez: Ah yes, it should be possible since this is merged https://github.com/paritytech/substrate/pull/14644

pallets/parachain-system/src/lib.rs
```diff
@@ -829,26 +847,33 @@ impl<T: Config> Pallet<T> {
...
+				dmq_head.extend_downward(m);
+			}
+			// Note: we are not using `.defensive()` here since that prints the whole value to console. In case that the message is too long, this clogs up the log quite badly.
+			let bounded =
+				downward_messages
+					.iter()
+					.filter_map(|m| match BoundedSlice::try_from(&m.msg[..]) {
+						Ok(bounded) => Some(bounded),
+						Err(_) => {
+							defensive!("Inbound Downward message was too long; dropping");
```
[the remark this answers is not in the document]
ggwpez: We always have to call `extend_downward`, otherwise the block will get rejected. This just records that the message was "processed" by the runtime - whatever that means.  
Further there is nothing that we can do if the message is too long...

pallets/parachain-system/src/lib.rs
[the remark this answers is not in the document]
ggwpez: I changed it, but it makes the code more complicated and ugly. I guess its better like it is now.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -941,84 +510,173 @@ impl<T: Config> Pallet<T> {
...
+							.is_err()
 						{
-							status[i].state = InboundState::Suspended;
-							let r = Self::send_signal(sender, ChannelSignal::Suspend);
-							if r.is_err() {
-								log::warn!(
-									"Attempt to suspend channel failed. Messages may be dropped."
-								);
-							}
+							defensive!("Not enough weight to process signals - dropping.");
```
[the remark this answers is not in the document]
ggwpez: Hm its difficult... with the current logic it will try to re-send the suspend signal when the next message on a suspended channel comes in. This goes on until the `drop_threshold` is reached, so i think with proper configuration this should be fine.  
I am always very hesistant to ignore the weight limit, its a hard limit after all.

serban300 (approved):
Looks good to me.  Please just take into consideration the fact that I am not very familiar with this repo and it would be best if someone from cumulus would also take a look.

pallets/xcmp-queue/src/lib.rs
svyatonik: > We now use the weight only for enqueueing and signal processing - not message processing anymore.

True and that's what is worrying me :) Like - previously we have been ignoring this weight and now we don't (which is technically correct). And because of that we may drop some messages.

But ok - let's go with he current impl. I think there's some significant reserve (block quarter) for just enqueueing.

svyatonik (approved):
LGTM. We've been using this branch in our bridge tests (HRMP) and it worked like a charm. Backpressure worked fine too.

xlc:
where is the migration code for storage in the removed dmp queue pallet?

ggwpez:
> where is the migration code for storage in the removed dmp queue pallet?

There is a generic RemovePallet migration in FRAME somwhere. I will deploy it here (its a TODO).

xlc:
don’t we need to migrate the items from old pallet to new pallet? we can’t just delete them

ggwpez:
> don’t we need to migrate the items from old pallet to new pallet? we can’t just delete them

Yes... i added `cumulus_pallet_dmp_queue::UndeployDmp` and `cumulus_pallet_dmp_queue::DeleteDmp` as described in the description.  
I will check again, but something like this is probably also needed for the XCMP queue pallet.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3409352

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3409582

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3420281

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3420383

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3424752

ggwpez:
@xlc the migration that i added for the DMP messages has the problem that it migrates them all at once.  
This can cause overweight blocks and stall the chain.

Maybe we can do a multi-block migration in the DMP pallet without locking the runtime since the storage is inaccessible anyway.  
Then the DMP pallet would migrate them one by one into the MQ pallet. Order would be wrong but XCM cannot rely on that anyway.

PR #2164, opened 2023-02-06
Add retry mechanism for pov-recovery, fix full-node pov-recovery

### LINKED ISSUE #2142: Add retry mechanism for pov-recovery

skunert:
This PR introduces some changes to the PoV recovery mechanism.

### Retry in case of unavailability
PoV recovery is listening to a stream of pending candidates that is derived from the relay chain import notification stream. Since the candidates are pending, availability is not yet guaranteed. In that case we now just retry the operation.

Currently we retry only once.

### Ordered recovery queue
Until now for each recovery operation we added a randomized delay after which the recovery would start. This was not a problem for collators. For full nodes however, we have a configured recovery delay of 150 - 300 seconds. Since the recovery can start anywhere  in that range, it could happen that we attempted to recover a child before its parent was available. There is already a mechanism to wait for parents, but it was only waiting for parent currently in recovery, which is not true for queued ones. 

This is not changed into an ordered queue that will add randomized recovery slots on insertion, but start recoveries in order.

### Finalization
When full-nodes recover a block via PoV (which should be very rare), they are behind by several minutes. Since we are following relay chain final and best block, the node had no way of knowing the status of the recovered block. We are now maintaining a cache of seen finalized blocks. If a block is imported and its hash is in the cache, we immediately set it as finalized.

 
Closes #2142


client/pov-recovery/src/lib.rs
```diff
@@ -105,26 +132,84 @@ pub struct RecoveryDelay {
 	pub max: Duration,
 }
 
+impl RecoveryDelay {
+	/// Produce a randomized duration between `min` and `max`.
+	fn duration(&self) -> Duration {
+		self.min + self.max.saturating_sub(self.min).mul_f64(thread_rng().gen())
```
koute, line 138: Why not `Uniform::from(self.min..=self.max).sample(&mut thread_rng())`?

client/consensus/common/src/parachain_consensus.rs
```diff
@@ -41,64 +43,130 @@ const LOG_TARGET: &str = "cumulus-consensus";
...
+			);
+			return
+		},
+	};
+
+	let hash = header.hash();
+
+	last_seen_finalized_hashes.insert(hash, ());
+	// don't finalize the same block multiple times.
+	if parachain.usage_info().chain.finalized_hash != hash {
```
melekes, line 71: under which conditions we might attempt to refinalize the same block?

client/pov-recovery/src/lib.rs
```diff
@@ -105,26 +132,84 @@ pub struct RecoveryDelay {
...
+				} else {
+					tracing::warn!(
+						target: LOG_TARGET,
+						"Recovery was signaled, but no candidate hash available. This is a bug."
+					);
+				};
+			}
+			futures::pending!()
+		}
+	}
```
koute, line 200: Hmm... can't this in theory mix up the delays? e.g.:

```rust
queue.push_recovery(BLOCK_A, Duration::from_secs(10));
queue.push_recovery(BLOCK_B, Duration::from_secs(5));
```

The 5s duration from the BLOCK_B would fire first (since this uses `FuturesUnordered`), but would pop from the front of the `recovery_queue`, so will pop BLOCK_A.

Also, perhaps it'd be better to pass the `retry_delay: Duration` and `recovery_delay: RecoveryDelay` in `new()` (since those are effectively constant after they're calculated, if I'm looking at it right?) and then just sample the `recovery_delay` internally here in the queue when it's needed?

client/pov-recovery/src/lib.rs
```diff
@@ -105,26 +132,84 @@ pub struct RecoveryDelay {
 	pub max: Duration,
 }
 
+impl RecoveryDelay {
```
koute, line 135: I know you didn't add this type, but it might be nice to rename it as it's a little confusing. Reading `RecoveryDelay` I thought that it contains a delay for the recovery as the name suggests, but it actually contains the range the delay can take. So maybe `RecoveryDelayRange`?

client/pov-recovery/src/lib.rs
skunert: > Hmm... can't this in theory mix up the delays? e.g.:

Yes this is correct. Basically the delay adds only a new "signaling slot" to the signaling queue at which the next block of the queue is recovered. Insertion order should be respected, but it does not matter by which signal triggers the recovery. 

The other suggestion is nice, I think we can even get rid of retry_delay and just use the same delay one for retries and simple recoveries.

client/consensus/common/src/parachain_consensus.rs
skunert: Basically, this piece of code follows the relay chain finalization notifications. When we see one, we find the currently included parachain header for that block and finalize it. However, if the parachain did not make progress (for whatever reason) during two finalization notifications of the relay chain, the included parachain header would be the same, and we need to guard against this.

melekes (approved):
👍 (code-wise; don't know enough to reason about business logic)

client/pov-recovery/src/lib.rs
```diff
@@ -91,52 +118,114 @@ pub struct RecoveryRequest<Block: BlockT> {
...
 	parachain_client: Arc<PC>,
 	parachain_import_queue: Box<dyn ImportQueueService<Block>>,
 	relay_chain_interface: RC,
 	para_id: ParaId,
 	/// Explicit block recovery requests channel.
 	recovery_chan_rx: Receiver<RecoveryRequest<Block>>,
+	/// Blocks that we are retrying currently
+	candidates_in_retry: HashSet<Block::Hash>,
+	/// Waiting time before a retry
+	retry_delay: Duration,
```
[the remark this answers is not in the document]
skunert: Yes, that is the plan :+1:

client/service/src/lib.rs
```diff
@@ -113,25 +115,27 @@ where
...
-
 	let pov_recovery = PoVRecovery::new(
-		overseer_handle.clone(),
+		recovery_handle,
 		// We want that collators wait at maximum the relay chain slot duration before starting
-		// to recover blocks.
-		RecoveryDelay { min: core::time::Duration::ZERO, max: relay_chain_slot_duration },
+		// to recover blocks. Additionally, we wait at least half the slot time to give the
+		// relay chain the chance to increase availability.
+		RecoveryDelayRange { min: relay_chain_slot_duration / 2, max: relay_chain_slot_duration },
```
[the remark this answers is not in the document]
skunert: `relay_chain_slot_duration` is a `Duration` which has an implementation for division in place, so we should be fine here.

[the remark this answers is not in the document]
davxy: I've corrected my comment above. There was a typo... I was intending strictly greater than 0. That is, I was suspecting that maybe we want it > 0?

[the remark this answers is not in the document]
davxy: Nevermind. I was thinking using integer arithmetic 🤦‍♂️. Duration will scale down to subseconds. Sorry

client/pov-recovery/src/lib.rs
skunert: Delays have been merged into the queue :+1:

client/consensus/common/src/parachain_consensus.rs
```diff
@@ -48,57 +99,74 @@ const RECOVERY_DELAY: RecoveryDelay =
...
-						target: LOG_TARGET,
-						error = ?e,
-						block_hash = ?hash,
-						"Failed to finalize block",
-					),
+			imported = imported_blocks.next() => {
+				match imported {
+					Some(imported_block) => {
+						// When we see a block import that is already finalized, we immediately finalize it.
+						if last_seen_finalized_hashes.peek(&imported_block.hash).is_some() {
```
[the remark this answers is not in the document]
skunert: Yes, we are only receiving the finalized notifications every few relay chain blocks. But this code is super niche, only relevant for full-node recovery. Since we recover block by block, we will soon reach one that was explicitly marked as finalized.
Since the notifications we are reacting on here are import notifications, I am worried about receiving an import notification of a block below `finalized_num` but that should not be finalized. I think the current method of just finalizing what we have explicitly seen it more foolproof.

test/service/src/lib.rs
```diff
@@ -126,6 +131,41 @@ pub type ParachainBlockImport = TParachainBlockImport<Block, Arc<Client>, Backen
...
 
+/// Recovery handle that fails regularly to simulate unavailable povs.
+pub struct FailingRecoveryHandle {
+	overseer_handle: OverseerHandle,
+	counter: u32,
+}
+
+impl FailingRecoveryHandle {
+	/// Create a new FailingRecoveryHandle
+	pub fn new(overseer_handle: OverseerHandle) -> Self {
```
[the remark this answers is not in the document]
skunert: I thought about this, but would be in favor of adding this is we ever have a use case.

davxy (approved):
Looks good 🟢

client/pov-recovery/src/lib.rs
```diff
@@ -241,16 +328,17 @@ where
 		}
 	}
 
-	/// Clear `waiting_for_parent` from the given `hash` and do this recursively for all child
-	/// blocks.
-	fn clear_waiting_for_parent(&mut self, hash: Block::Hash) {
+	/// Clear `waiting_for_parent` and `waiting_recovery` for the candidate with `hash`.
```
bkchr, line 331:
```suggestion
	/// Clear `waiting_for_parent` and `waiting_recovery` for the candidate with `hash`.
	///
```

client/pov-recovery/src/lib.rs
```diff
@@ -302,12 +402,19 @@ where
 
 		match self.parachain_client.block_status(parent) {
 			Ok(BlockStatus::Unknown) => {
-				if self.active_candidate_recovery.is_being_recovered(&parent) {
+				// If the parent block is currently being recovered or is scheduled to be recovered,
+				// we want to wait for the parent.
+				let parent_scheduled_for_recovery =
+					self.candidates.get(&parent).map_or(false, |parent| parent.waiting_recovery);
+				if self.active_candidate_recovery.is_recovering(&parent) ||
+					parent_scheduled_for_recovery
```
bkchr, line 410:
```suggestion
				let parent_being_recovered =
					self.candidates.get(&parent).map_or(false, |parent| parent.waiting_recovery);
				if parent_being_recovered
```

You just updated the comment above `waiting_recovery` saying that this is only `true` when you either waiting for recovery or it is being recovered.

bkchr (approved):
Some last things.

client/pov-recovery/src/lib.rs
```diff
@@ -425,22 +533,7 @@ where
 
 		if do_recover {
```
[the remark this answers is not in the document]
skunert: Good catch!

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2369515

PR #2133, opened 2023-01-24
Add Support for Foreign Assets

joepetrowski:
Adds support to Statemine and Westmint for foreign fungible assets. This will allow other parachains to represent their assets, as well as bridges to represent assets from other networks, on Statemine. A follow up PR will introduce this to Statemint once it has been audited and _in the Kusama coal mines_ for some time.

PR #2167, opened 2023-02-07
Assets/ForeignAssets tests and fixes (still in-progress)

bkontur:
- [ ] how to setup Deposits (AssetDeposit, ExistentialDeposit), admin/owner + initial balance for ForeignCreator?
- [ ] cannot call `Transact([create, set_metadata])` with xcm origin, because `set_metadata` uses just `ensure_signed()`
- [ ] fix Weight for `fn universal_origin(_: &Junction) -> Weight {`
- [ ] UniversalAliases/TrustedBridgedNetworks - create some general pallet for on-chain cfg (something like bridge-asset-transfer has, maybe split)
- [ ] move all usefull origin_conversion/barriers stuff to the polkadot/xcm-builder/xcm-executor

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -340,6 +340,8 @@ dependencies = [
  "sp-runtime",
  "sp-std",
  "substrate-wasm-builder",
+ "xcm",
+ "xcm-executor",
 ]
 
 [[package]]
--- a/parachains/runtimes/assets/test-utils/Cargo.toml
+++ b/parachains/runtimes/assets/test-utils/Cargo.toml
@@ -23,6 +23,10 @@ sp-core = { git = "https://github.com/paritytech/substrate", default-features = 
 pallet-collator-selection = { path = "../../../../pallets/collator-selection", default-features = false }
 parachains-common = { path = "../../../common", default-features = false }
 
+# Polkadot
+xcm = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "master" }
+xcm-executor = { git = "https://github.com/paritytech/polkadot", default-features = false, branch = "master" }
+
 [dev-dependencies]
 hex-literal = "0.3.4"
 
@@ -42,4 +46,6 @@ std = [
     "sp-io/std",
     "sp-runtime/std",
     "sp-std/std",
+    "xcm/std",
+    "xcm-executor/std",
 ]
--- a/parachains/runtimes/assets/test-utils/src/lib.rs
+++ b/parachains/runtimes/assets/test-utils/src/lib.rs
@@ -56,6 +56,11 @@ impl<Runtime: frame_system::Config + pallet_balances::Config + pallet_session::C
 		self
 	}
 
+	pub fn with_tracing(self) -> Self {
+		frame_support::sp_tracing::try_init_simple();
+		self
+	}
+
 	pub fn build(self) -> sp_io::TestExternalities
 	where
 		Runtime:
--- a/parachains/runtimes/assets/test-utils/src/lib.rs
+++ b/parachains/runtimes/assets/test-utils/src/lib.rs
@@ -6,6 +6,11 @@ use parachains_common::AccountId;
 use sp_consensus_aura::AURA_ENGINE_ID;
 use sp_core::Encode;
 use sp_runtime::{Digest, DigestItem};
+use xcm::{
+	latest::{MultiAsset, MultiLocation, XcmContext},
+	prelude::{Concrete, Fungible, XcmError},
+};
+use xcm_executor::{traits::TransactAsset, Assets};
 
 pub type BalanceOf<Runtime> = <Runtime as pallet_balances::Config>::Balance;
 pub type AccountIdOf<Runtime> = <Runtime as frame_system::Config>::AccountId;
@@ -135,5 +140,22 @@ where
 		account_id: AccountIdOf<Runtime>,
 	) -> <Runtime as frame_system::Config>::RuntimeOrigin {
 		<Runtime as frame_system::Config>::RuntimeOrigin::signed(account_id.into())
+	}
+}
+
+impl<XcmConfig: xcm_executor::Config> RuntimeHelper<XcmConfig> {
+	pub fn do_transfer(
+		from: MultiLocation,
+		to: MultiLocation,
+		(asset, amount): (MultiLocation, u128),
+	) -> Result<Assets, XcmError> {
+		<XcmConfig::AssetTransactor as TransactAsset>::transfer_asset(
+			&MultiAsset { id: Concrete(asset), fun: Fungible(amount) },
+			&from,
+			&to,
+			// We aren't able to track the XCM that initiated the fee deposit, so we create a
+			// fake message hash here
+			&XcmContext::with_message_hash([0; 32]),
+		)
 	}
 }
--- a/parachains/runtimes/assets/westmint/src/foreign_conversions.rs
+++ b/parachains/runtimes/assets/westmint/src/foreign_conversions.rs
@@ -1,0 +1,128 @@
+// Copyright (C) 2023 Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// 	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! TODO: this will go away from Westmint for better reusability
+//!
+//! All conversions stuff to support bridging scenario (asset transfer, asset creation, ...)
+
+use codec::{Decode, Encode};
+use frame_support::{sp_runtime::RuntimeDebug, traits::Contains};
+use scale_info::TypeInfo;
+use sp_core::{Get, TypeId};
+use sp_runtime::traits::AccountIdConversion;
+use xcm::latest::{prelude::*, Junction, MultiLocation};
+use xcm_executor::traits::Convert;
+
+/// [`MultiLocation`] is supposed to be local bridge-hub and [`Junction]` is foreign [`GlobalConsensus`]
+pub type BridgedUniversalAliases = sp_std::vec::Vec<(MultiLocation, Junction)>;
+
+/// Checks if [`MultiLocation`] is from different global consensus and if it is allowed
+pub struct IsTrustedGlobalConsensus<UniversalAliases>(
+	sp_std::marker::PhantomData<UniversalAliases>,
+);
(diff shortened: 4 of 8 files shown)


bkontur:
bot rebase

bkontur:
bot rebase

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2437828

xlc:
Not sure this is the best place to ask but I have some questions:

1. Is this going to be permisionless and universal? i.e. support all assets from all parachains without governance on relaychain
2. How to prevent dust issue?
3. Can foreign asset be self-sufficient and eligible for tx payment?
4. How to handle asset metadata?
5. Any high level design docs that may already have answers to my questions?

cc @joepetrowski

joepetrowski:
1. Yes, you can see the [`try_origin`](https://github.com/paritytech/cumulus/pull/2133/files#diff-07287c62dd91150d975f7504bd14cc62d9889c44f7ba8ca44c30c03b05d5b071R401) implementation in the base PR. So e.g. Acala governance will have a successful origin for ACA as well as tokens registered under its governance system.
2. Same as any other asset.
3. Yes. First, via governance referendum making them sufficient. Once DOTswap is working with that I'd like to re-evaluate sufficiency in general. As in, perhaps it is safe to say that anything that can be swapped for DOT to pay fees with on-the-spot is eligible for tx payment.
4. On asset creation, the para origin (i.e., e.g., Acala governance) will be the owner of that asset. It could further control the other privileged accounts, or use `set_team` to designate a local account with those privileges.
5. Just the discussions on the forum.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2481057

bkontur:
bot rebase

parachains/runtimes/assets/common/src/lib.rs
```diff
@@ -15,39 +15,78 @@
...
+	MultiLocationConvertedConcreteId<
+		EverythingBut<(
+			// Excludes relay/parent chain currency
+			Equals<ParentLocation>,
+			// Here we rely on fact that something like this works:
+			// assert!(MultiLocation::new(1, X1(Parachain(100))).starts_with(&MultiLocation::parent()));
+			// assert!(X1(Parachain(100)).starts_with(&Here));
+			StartsWith<LocalMultiLocationPattern>,
+			// Here we can exclude more stuff or leave it as `()`
+			AdditinalMultiLocationExcludeFilter,
```
joepetrowski, line 74:
```suggestion
			AdditionalMultiLocationExclusionFilter,
```

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -361,6 +378,7 @@ impl Default for ProxyType {
 		Self::Any
 	}
 }
+// TODO:check-parameter - setup ProxyType for ForeignAssets?
```
joepetrowski, line 381: I was thinking no (based on [this comment](https://github.com/paritytech/cumulus/pull/2167/files#diff-67f8d30fd426048ab88d224694ba10c5f225eb66b5afdcdcb10ead450e6d2c02R277)), but am also not strongly opposed to it. It's more of a guess about how people might use this, but perhaps they would want proxies.

parachains/runtimes/assets/statemine/tests/tests.rs
```diff
@@ -371,9 +373,15 @@ fn test_assets_balances_api_works() {
 		.build()
 		.execute_with(|| {
 			let local_asset_id = 1;
+			let foreign_asset_id_multilocation =
+				MultiLocation { parents: 2, interior: X1(GlobalConsensus(Kusama)) };
```
joepetrowski, line 377: This might actually be something we want to block in `ForeignAssetsConvertedConcreteId` as in this context (Statemine) it is actually `Equals<ParentLocation>`.

joepetrowski: And should just be KSM.

bkontur:
bot rebase

parachains/runtimes/assets/statemine/src/lib.rs
bkontur: I removed that TODO, so maybe lets wait for demands from comunity

parachains/runtimes/assets/statemine/tests/tests.rs
bkontur: yes, we do that blocking for ForeignAssets:
- local relay chain token is not allowed (means KSM for statemine)
- also Statemine's `TrustedBackAssets` are not allowed here

https://github.com/paritytech/cumulus/pull/2167/files#diff-c92582ea200acffaa2858161fe6684de752bfd60d6828fc4d4e283116c036c94R59-R74
```
/// Excludes by default:
/// - parent as relay chain
/// - all local MultiLocations
///
/// `AdditionalMultiLocationExclusionFilter` can customize additional excluded MultiLocations
pub type ForeignAssetsConvertedConcreteId<AdditionalMultiLocationExclusionFilter, Balance> =
	MultiLocationConvertedConcreteId<
		EverythingBut<(
			// Excludes relay/parent chain currency
			Equals<ParentLocation>,
			// Here we rely on fact that something like this works:
			// assert!(MultiLocation::new(1, X1(Parachain(100))).starts_with(&MultiLocation::parent()));
			// assert!(X1(Parachain(100)).starts_with(&Here));
			StartsWith<LocalMultiLocationPattern>,
			// Here we can exclude more stuff or leave it as `()`
			AdditionalMultiLocationExclusionFilter,
```

bkontur: ah, I see, it was copy/paste from westmint, which adds here confusion, I changed that example here.

Anyway, from **Statemine** perspective `MultiLocation { parents: 2, interior: X1(GlobalConsensus(Kusama)) }`  is not a **Statemine/Kusama true KSM**, because there is a `parent: 2`, so it means more like **relay chain currency of some 
different Kusama global consensus**.

`GlobalConsensus(Kusama)` is `NetworkId` from `xcm:v3`.
So maybe, we could add additional exclude to `ForeignAssetsConvertedConcreteId` for these cases: **do not even allow any other foreign `GlobalConsensus(same_as_local_consensus)` where parent > 1**.
So, from the `Statemine` perspective it would mean we only accept **only one true GlobalConsensus(Kusama)** which is our and we ignore any other `GlobalConsensus(Kusama)`.

I am not sure if any other Kusama will exist, but if does and will be bridged to our Statemine, all assets will end in `ForeignAssets`.

Hypotetically, problem will come when there will be two more foreign `GlobalConsensus(Kusama)`, so in that case , from Statemine perspective, both will be `MultiLocation { parents: 2, interior: X1(GlobalConsensus(Kusama)) }` and there is no way to differ between them, that means their assets will be also mixed in `ForeignAssets`.

So, that makes me thing, that we should really exclude **`GlobalConsensus(same_as_local_consensus)` where parent > 1**.

So:
Statemine should exclude **`GlobalConsensus(Kusama)` where parent > 1**
Westmint should exclude **`GlobalConsensus(Westend)` where parent > 1**
Statemint should exclude **`GlobalConsensus(Polkadot)` where parent > 1**

joepetrowski: Yeah this seems like a good approach.

> so it means more like relay chain currency of some different Kusama global consensus.

`NetworkId` is an [enum in XCM](https://github.com/paritytech/polkadot/blob/c77d92ae46e000a35fe366c56d3e415e9eceb81c/xcm/src/v3/junction.rs#L40) though, so any variants have to be added explicitly there. So yes there could be some other network called "Kusama" in the future, its disambiguation would need to be handled in a change to that enum.

bkontur: here is a simple test that simulates more global consensuses Kusama and how should different consensus reanchor their assets:
```
                // local asset on different global consensus Kusama
		let my_asset_location = MultiLocation {parents: 0, interior: X2(PalletInstance(1), GeneralIndex(1234))};
		let my_universal_location: InteriorMultiLocation = X2(GlobalConsensus(Kusama), Parachain(2000)).into();

		// our statemine on our global consensus Kusama
		let statemine_as_target = MultiLocation {parents: 2, interior: X2(GlobalConsensus(Kusama), Parachain(1000))};

		// how should my asset look like from foreign statemine perspective
		let mut my_asset_location_for_statemine = my_asset_location.clone();
		my_asset_location_for_statemine.reanchor(&statemine_as_target, my_universal_location).expect("failed");
		println!("{:?}", my_asset_location_for_statemine);

		// our statemint on our global consensus Polkadot
		let statemint_as_target = MultiLocation {parents: 2, interior: X2(GlobalConsensus(Polkadot), Parachain(1000))};

		// how should my asset look like from foreign statemint perspective
		let mut my_asset_location_for_statemint = my_asset_location.clone();
		my_asset_location_for_statemint.reanchor(&statemint_as_target, my_universal_location).expect("failed");
		println!("{:?}", my_asset_location_for_statemint);
```

and the result is:
```
// this will come to Statemine
MultiLocation { parents: 1, interior: X3(Parachain(2000), PalletInstance(1), GeneralIndex(1234)) }

// this will come to Statemint
MultiLocation { parents: 2, interior: X4(GlobalConsensus(Kusama), Parachain(2000), PalletInstance(1), GeneralIndex(1234)) }
```



so in this case, xcm origin will be like:
```
UniversalOrigin(GlobalConsensus(Kusama)),
...

but on Statemine, it will be rejected by `XcmExecutor`
```
   UniversalOrigin(new_global) => {
				let universal_location = Config::UniversalLocation::get();
				ensure!(universal_location.first() != Some(&new_global), XcmError::InvalidLocation);
```


so that could mean, that XCM expects only one `GlobalConsensus(Kusama)` to exist

Comment (bkontur) [reply parachains/runtimes/assets/statemine/tests/tests.rs]: > `NetworkId` is an [enum in XCM](https://github.com/paritytech/polkadot/blob/c77d92ae46e000a35fe366c56d3e415e9eceb81c/xcm/src/v3/junction.rs#L40) though, so any variants have to be added explicitly there. So yes there could be some other network called "Kusama" in the future, but its disambiguation would need to be handled in a change to that enum.

aha, ok, thank you, cool, that explains also my test above

File: parachains/runtimes/assets/test-utils/src/test_cases.rs [hunk 4]
```diff
@@ -0,0 +1,1016 @@
...
+				),
+				pallet_assets::Error::<Runtime, ForeignAssetsPalletInstance>::NoPermission
+			);
+
+			// lets try create asset for different parachain(3333) (foreign_creator(2222) can create just his assets)
+			let foreign_asset_id_multilocation =
+				MultiLocation { parents: 1, interior: X2(Parachain(3333), GeneralIndex(1234567)) };
+			let asset_id = AssetIdConverter::convert(foreign_asset_id_multilocation).unwrap();
+
+			// prepapre data for xcm::Transact(create)
```
joepetrowski, line 931:
```suggestion
			// prepare data for xcm::Transact(create)
```

joepetrowski (approved):

Polkadot-Forum:

https://forum.polkadot.network/t/bridge-hub-questions/2693/2

PR #2216, opened 2023-02-20
remove every usage of `#[pallet::getter]`

muraca:
companion for [Substrate #13365](https://github.com/paritytech/substrate/pull/13365/)

@kianenigma @shawntabrizi


pallets/parachain-system/src/lib.rs
```diff
@@ -220,7 +220,7 @@ pub mod pallet {
 					return
 				},
 			};
-			let relevant_messaging_state = match Self::relevant_messaging_state() {
```
bkchr, line 223: You didn't removed any of the getters.

PR #2221, opened 2023-02-22
Collectives integration tests xcm v3

muharem:
Adopting existing integration tests of the Polkadot Collectives chain for xcm v3.

Requires:
https://github.com/paritytech/parachains-integration-tests/pull/89
https://github.com/paritytech/cumulus/pull/2222

PR #2111, opened 2023-01-19
BridgeHub Rococo/Wococo runtimes

bkontur:
This Rococo/Wococo bridge is PoC for final goal Kusama/Polkadot Bridge, which will be based on this PoC.

PR contains:
- BridgeHub runtime implementations + tests for Rococo/Wococo bridge.
- zombienets + bash script for local run
- bash script which is used by Kubernetes cron job to periodically transfer assets from live Rococo:Rockmine2 to Wococo:Wockmint

Implementation is based on bridging pallets which are added to the Cumulus as git subtree here https://github.com/paritytech/cumulus/pull/2458

---
Rococo/Wococo bridge is deployed live:
https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Frococo-bridge-hub-rpc.polkadot.io#/explorer
https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fwococo-bridge-hub-rpc.polkadot.io#/explorer

---
More about bridges:
https://github.com/paritytech/parity-bridges-common/blob/master/docs/high-level-overview.md
https://github.com/paritytech/parity-bridges-common/blob/master/docs/polkadot-kusama-bridge-overview.md
https://forum.parity.io/t/bridges-team-updates/1403/11
https://forum.polkadot.network/t/bridge-hub-questions/2693/2
https://forum.polkadot.network/t/bridge-hub-questions/2693/5

PR #2222, opened 2023-02-22
Collectives chain xcm filter

muharem:
Allow the executing of Alliance::init_member and XcmPallet::force_xcm_version via XCM Transact operation.

This is nice to have, both calls required for the integration tests https://github.com/paritytech/cumulus/pull/2221.

### DIFF
--- a/parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
+++ b/parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
@@ -155,6 +155,9 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 			RuntimeCall::DmpQueue(..) |
 			RuntimeCall::Utility(pallet_utility::Call::as_derivative { .. }) |
 			RuntimeCall::Alliance(
+				// `init_members` accepts unbounded vecs as arguments,
+				// but the call can be initiated only by root origin.
+				pallet_alliance::Call::init_members { .. } |
 				pallet_alliance::Call::vote { .. } |
 				pallet_alliance::Call::close_old_weight { .. } |
 				pallet_alliance::Call::disband { .. } |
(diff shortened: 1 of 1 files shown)


joepetrowski (approved):

parachains/runtimes/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -175,7 +178,8 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 				pallet_collective::Call::close_old_weight { .. } |
 				pallet_collective::Call::disapprove_proposal { .. } |
 				pallet_collective::Call::close { .. },
-			) => true,
+			) |
+			RuntimeCall::PolkadotXcm(pallet_xcm::Call::force_xcm_version { .. }) => true,
```
bkontur, line 182: @muharem @joepetrowski 
other cumulus runtimes do not need this `force_xcm_version` ?
what is the use-case with `force_xcm_version` for collectives?

bkontur: for bridge-hubs, I am testing if governance can call some dedicated calls with transact:
https://github.com/paritytech/cumulus/pull/2111/files#diff-70774adbfac7198ac09980fc1be181b5a22d99282ae82e454540dfd28e35febdR228-R267

muharem: we set the version manually here https://github.com/paritytech/cumulus/pull/2221/files#diff-883b2a05c895554621314a4511df8dde5546f1d2ce910ac6ba5899a56c95bae1R85
otherwise xcm does it by itself, after the first xcm call, and if you test using parachains-integration-tests, it might mess the events and you get flaky tests.

bkontur (approved):

gilescope (approved):

muharem:
bot merge

PR #2224, opened 2023-02-22
wip: upgrade statemine/t integration tests to xcm v3

gilescope:
(best to look commit by commit to reduce from initially updating the tests from the other repo)

### DIFF
--- a/parachains/integration-tests/statemine/xcm/0_init.yml
+++ b/parachains/integration-tests/statemine/xcm/0_init.yml
@@ -11,7 +11,7 @@ settings:
       paraId: &pp_id 2000
   variables:
     common:
-      xcm_verison: &xcm_version '2'
+      xcm_version: &xcm_version '2'
       require_weight_at_most: &weight_at_most 1000000000
     chains:
       relay_chain:
@@ -37,7 +37,7 @@ tests:
     its:
       - name: XCM supported versions between chains
         actions:
-          - extrinsics: # Relay Chain sets supported version for Assset Parachain
+          - extrinsics: # Relay Chain sets supported version for Asset Parachain
             - chain: *relay_chain
               sudo: true
               signer: *rc_signer
@@ -56,13 +56,13 @@ tests:
               ]
               events:
                 - name: sudo.Sudid
-                  attribute:
-                    type: Result<Null, SpRuntimeDispatchError>
-                    value: Ok
+                  attributes:
+                    - type: Result<Null, SpRuntimeDispatchError>
+                      value: Ok
                 - name: xcmPallet.SupportedVersionChanged
-                  attribute:
-                    type: u32
-                    value: *xcm_version
+                  attributes:
+                    - type: u32
+                      value: *xcm_version
           - extrinsics: # Relay Chain sets supported version for Penpal Parachain
             - chain: *relay_chain
               sudo: true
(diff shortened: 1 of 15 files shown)


parachains/integration-tests/statemine/xcm/0_init.yml
```diff
@@ -98,28 +98,46 @@ tests:
               args: [
                 *ap_dest, # destination
                 {
-                  v2: [ #message
+                  v3: [ #message
+                    {
+                      UnpaidExecution: {
+                        weightLimit: Unlimited,
```
muharem, line 104: Not sure if this will pass the Barrier, did it?

parachains/integration-tests/statemine/xcm/1_dmp.yml
```diff
@@ -69,16 +69,16 @@ tests:
...
+                          xcmOutcome: Complete
                     - name: dmpQueue.ExecutedDownward
                       chain: *assets_parachain
-                      attribute:
-                        type: XcmV2TraitsOutcome
-                        isComplete: true
-                        threshold: [10, 10]
-                        value: 1,021,973,000
+                      attributes:
+                        - type: XcmV2TraitsOutcome
```
muharem, line 78: v3 is set on init, but here v2.
check in the output if the assertions did really happen

parachains/integration-tests/statemine/xcm/0_init.yml
gilescope: yeah that got past the barrier (once the call was added to the safe call list)

parachains/integration-tests/statemine/xcm/1_dmp.yml
gilescope: just got 0 passing. Have changed this locally but not got 1_dmp passing yet. Will upload my changes.

parachains/integration-tests/statemine/xcm/3_hrmp-open-channels.yml
```diff
@@ -170,245 +77,76 @@ tests:
           - name: Penpal Parachain sends a request to the Relay Chain to open a channel with the Assets Parachain
```
ruseinov, line 77: Seems like the context has changed here. We are no longer sending the request from Penpal, but rather we call an extrinsic on the relay chain to open the channel. I guess we should then change the name of this test to something that reflects this scenario better.

parachains/integration-tests/statemine/xcm/3_hrmp-open-channels.yml
```diff
@@ -170,245 +77,76 @@ tests:
...
-                        {
-                          WithdrawAsset: [*rc_ksm_fungible]
-                        },
-                        {
-                          BuyExecution: {
-                              fees: *rc_ksm_fungible,
-                              weightLimit: Unlimited
-                          }
-                        },
+                    2000,
```
ruseinov, line 86: I guess this has to be the opposite for this test [1000, 2000...]
I have fixed it in integration-tests repo.

parachains/integration-tests/statemine/xcm/3_hrmp-open-channels.yml
ruseinov: Got it fixed in `parachains-integration-tests` repo.

parachains/integration-tests/statemine/xcm/3_hrmp-open-channels.yml
```diff
@@ -161,248 +74,79 @@ tests:
...
+                  requested_channels_a_to_p:
                     chain: *relay_chain
                     pallet: hrmp
                     call: hrmpOpenChannelRequestsList
                     args: []
               - asserts:
                   equal:
                     args: [
-                      $requested_channels,
+                      $requested_channels_a_to_p,
```
ruseinov, line 145: This assertion always fails for me. And it totally depends on the sequence of execution, if we change those tests around - it still fails. Because the list is empty. Since request processing happens on session start, this is going to be flaky even for the first test, depending on the machine and other variables. I suggest calling `force_process_hrmp_open` and then querying `HrmpIngressChannelsIndex` or `HrmpEgressChannelsIndex`. Otherwise we could just remove the assert completely as  the receipt of `HrmpChannelForceOpened` has to be enough.

xlc:
How will you test new kusama runtime works with both old statemine runtime and new statemine runtime?

parachains/integration-tests/statemine/xcm/3_hrmp-open-channels.yml
gilescope: Agreed - as long as we have the event I think we're good. The tests in the next file would fail if the channel was not open.

gilescope:
Closing this PR as most of it is put into https://github.com/paritytech/parachains-integration-tests/pull/92 - we will assume that is master for the tests from now on.

PR #2226, opened 2023-02-23
bump `zombienet` version to v1.3.35

pepoviola:
Bump zombienet version and removing the is up assertions since that is now checked by zombienet before running the tests.

Thanks!

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,7 +30,7 @@ variables:
   CI_IMAGE:                        "paritytech/ci-linux:production"
   DOCKER_OS:                       "debian:stretch"
   ARCH:                            "x86_64"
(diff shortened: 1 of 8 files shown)


skunert:
Nice! What is the exact logic after this change? When are the nodes considered "not up"?

skunert (approved):

bkchr (approved):
Nice :heart:

pepoviola:
> Nice! What is the exact logic after this change? When are the nodes considered "not up"?

We move the `is up` check to the spawning phase, so the test-runner will start only if all the nodes are `up` (this means that the prometheus server of the node is reachable).

Thanks!

sergejparity (approved):

alvicsam (approved):

PR #2261, opened 2023-03-01
Add Version Checks on Para Upgrade

joepetrowski:
Adds checks to ensure that the spec version is increasing and that the spec name is the same. This will prevent accidentally upgrading a parachain to a different runtime.

### Migration

Although this changes storage, I did not increase the storage version or provide a migration. The reason is that the upgrade itself will `kill` the affected storage item, and the next time an upgrade is authorized, the new type will be written. But this will break any front ends that show authorized upgrades, so let me know if I should bump the storage version for this.

### Transaction Version

Unfortunately breaks it, although it's a Root call. Could also add a new function `authorize_upgrade_without_checks` as an alternative.

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -138,7 +138,8 @@ impl CheckAssociatedRelayNumber for AnyRelayNumber {
 
 /// Information needed when a new runtime binary is submitted and needs to be authorized before
 /// replacing the current runtime.
-#[derive(Decode, Encode, Default, PartialEq, Eq, MaxEncodedLen)]
+#[derive(Decode, Encode, Default, PartialEq, Eq, MaxEncodedLen, TypeInfo)]
+#[scale_info(skip_type_params(T))]
 struct CodeUpgradeAuthorization<T>
 where
 	T: Config,
@@ -147,23 +148,6 @@ where
 	code_hash: T::Hash,
 	/// Whether or not to carry out version checks.
 	check_version: bool,
-}
-
-impl<T> TypeInfo for CodeUpgradeAuthorization<T>
-where
-	T: Config,
-{
-	type Identity = Self;
-
-	fn type_info() -> scale_info::Type {
-		scale_info::Type::builder()
-			.path(scale_info::Path::new("CodeUpgradeAuthorization", module_path!()))
-			.composite(scale_info::build::Fields::unnamed().field(|f| {
-				f.ty::<u8>().docs(&[
-					"Raw code upgrade authorization bytes, encodes code_hash + check_version",
-				])
-			}))
-	}
 }
 
 #[frame_support::pallet]
@@ -507,7 +491,7 @@ pub mod pallet {
 		/// Note that this function will not apply the new `code`, but only attempt to schedule the
 		/// upgrade with the Relay Chain.
 		///
-		/// The origin for this call should be unsigned.
+		/// All origins are allowed.
 		#[pallet::call_index(3)]
 		#[pallet::weight(1_000_000)]
 		pub fn enact_authorized_upgrade(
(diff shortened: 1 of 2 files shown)


pallets/parachain-system/src/lib.rs
```diff
@@ -134,6 +136,36 @@ impl CheckAssociatedRelayNumber for AnyRelayNumber {
 	fn check_associated_relay_number(_: RelayChainBlockNumber, _: RelayChainBlockNumber) {}
 }
 
+/// Information needed when a new runtime binary is submitted and needs to be authorized before
+/// replacing the current runtime.
+#[derive(Decode, Encode, Default, PartialEq, Eq, MaxEncodedLen)]
+struct CodeUpgradeAuthorization<T>
```
bkchr, line 142:
```suggestion
#[derive(Decode, Encode, Default, PartialEq, Eq, MaxEncodedLen, TypeInfo)]
#[scale_info(skip_type_params(T))]
struct CodeUpgradeAuthorization<T>
```

pallets/parachain-system/src/lib.rs
```diff
@@ -134,6 +136,36 @@ impl CheckAssociatedRelayNumber for AnyRelayNumber {
...
+	fn type_info() -> scale_info::Type {
+		scale_info::Type::builder()
+			.path(scale_info::Path::new("CodeUpgradeAuthorization", module_path!()))
+			.composite(scale_info::build::Fields::unnamed().field(|f| {
+				f.ty::<u8>().docs(&[
+					"Raw code upgrade authorization bytes, encodes code_hash + check_version",
+				])
+			}))
+	}
+}
```
bkchr, line 167:
```suggestion
```

pallets/parachain-system/src/lib.rs
```diff
@@ -442,17 +474,40 @@ pub mod pallet {
...
 
+		/// Provide the preimage (runtime binary) `code` for an upgrade that has been authorized.
+		///
+		/// If the authorization required a version check, this call will ensure the spec name
+		/// remains unchanged and that the spec version has increased.
+		///
+		/// Note that this function will not apply the new `code`, but only attempt to schedule the
+		/// upgrade with the Relay Chain.
+		///
+		/// The origin for this call should be unsigned.
```
bkchr, line 510:
```suggestion
		/// All origins are allowed.
```

We don't check the origin at all (on purpose).

pallets/parachain-system/src/lib.rs
```diff
@@ -712,9 +774,28 @@ pub mod pallet {
...
+				.and_then(|v| RuntimeVersion::decode(&mut &v[..]).ok())
+				.ok_or(Error::<T>::FailedToExtractRuntimeVersion)?;
+
+			if new_version.spec_name != current_version.spec_name {
+				return Err(Error::<T>::InvalidSpecName.into())
+			}
+
+			if new_version.spec_version <= current_version.spec_version {
+				return Err(Error::<T>::SpecVersionNeedsToIncrease.into())
+			}
```
bkchr, line 796:
```suggestion
			frame_system::Pallet::<T>::can_set_code(code)?;
```

bkchr (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2468624

gilescope (approved):

ruseinov (approved):

joepetrowski:
bot merge

xlc:
Can we have some scary labels to ensure this PR is getting attentions by parachain teams? It is a breaking change and will impact scripts related to runtime upgrade.
It will be even better if you can just make it a non-breaking change.

Issue #2195, opened 2023-02-16
Tracking issue for translating XCM tests from Zombienet to XCM simulator. 


ruseinov:
The goal is to be able to run as many as possible, if not all, XCM tests in the CI. This way we'll always know if any of the assumptions have changed and communicate that to the teams accordingly. Currently XCM tests reside in integration-tests directory and leverage Zombienet to run, but that has to be run manually.

A list of XCM tests that need to be translated as seen in [integration-tests](https://github.com/paritytech/cumulus/tree/0df51f43d4210b63dde235a2773056cd25105eb9/parachains/integration-tests/) directory:

WIP

bkchr:
Zombienet is the way to, but the mentioned integrration tests are using some layer put on top of Zombienet. The solution here should be that we write these XCM in "pure" Zombienet. Currently not that trivial, but you can already "branch out" to some js script where you can use polkadot-js. We should use this way for writing the tests and then we should use the correct Statemine/Statemint/etc runtimes to run these XCM tests in CI.

ruseinov:
> Zombienet is the way to, but the mentioned integrration tests are using some layer put on top of Zombienet. The solution here should be that we write these XCM in "pure" Zombienet. Currently not that trivial, but you can already "branch out" to some js script where you can use polkadot-js. We should use this way for writing the tests and then we should use the correct Statemine/Statemint/etc runtimes to run these XCM tests in CI.

The idea behind this issue is to replicate as many of those tests as possible via XCM-simulator in order to have something that documents XCM usage and configuration well for ecosystem teams as discussed with @joepetrowski. 
Perhaps implementing the Zombienet tests properly as you mentioned is something that needs to be mentioned in it's own issue and indeed used for the CI, while XCM-simulator tests would be more of a usage example doc for the community. 

@joepetrowski would be nice if you could chip in with our vision of the subject.

bkchr:
Okay! Yeah, I also just realized that `xcm-simulator` is using the runtimes directly. I had thought that this was mainly using some mocked stuff, but that way your issue makes more sense! Ty!

xlc:
FYI this is how we plan to write e2e xcm tests with Chopsticks https://github.com/AcalaNetwork/e2e-tests/blob/master/tests/karura-statemine/transfer-asset.test.ts

PR #2265, opened 2023-03-02
don't have integration tests in 2 places.

gilescope:
use https://github.com/paritytech/parachains-integration-tests/tree/master/examples as the definitive.


ruseinov (approved):

NachoPal:
The idea was to have them in Cumulus to be able to run them with a CI job: https://github.com/paritytech/ci_cd/issues/499

`parachains-integration-tests` was never meant to be the source of truth for the tests.

However, if we are moving away from these tests, to be replaced by this: https://github.com/paritytech/cumulus/issues/2195
then remove them from Cumulus.

ruseinov:
> The idea was to have them in Cumulus to be able to run them with a CI job: [paritytech/ci_cd#499](https://github.com/paritytech/ci_cd/issues/499)
> 
> `parachains-integration-tests` was never meant to be the source of truth for the tests. They are there just as an example of how to use the tool.
> 
> However, if we are moving away from these tests to be replaced by this: #2195, then it makes sense to remove them from Cumulus.

But the truth is that not only those tests ended up being the source of truth, they also come with a better runner!

joepetrowski:
Can close?

NachoPal:
> Can close?

Yes, will keep the tests in Cumulus. Closing it.

PR #2268, opened 2023-03-02
introduce RelaychainStateProvider to parachain-system

seunlanlege:
This will allow parachains to read the relay chain state via storage proofs.

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -1125,13 +1125,39 @@ pub trait OnSystemEvent {
 	/// Called when the validation code is being applied, aka from the next block on this is the new runtime.
 	fn on_validation_code_applied();
 }
+/// Holds the relay chain state root and block number.
+pub struct RelayChainState {
+	/// Current relay chain height.
+	pub number: relay_chain::BlockNumber,
+	/// State root for current relay chain height.
+	pub state_root: relay_chain::Hash,
+}
 
-/// Implements [`BlockNumberProvider`] that returns relay chain block number fetched from
+/// This exposes the [`RelayChainState`] to other runtime modules.
+///
+/// Enables parachains to read relay chain state via state proofs.
+pub trait RelaychainStateProvider {
+	/// Maybe called by any runtime module to obtain the current state of the relay chain.
+	fn current_relay_chain_state() -> RelayChainState;
+}
+
+impl<T: Config> RelaychainStateProvider for RelaychainDataProvider<T> {
+	fn current_relay_chain_state() -> RelayChainState {
+		Pallet::<T>::validation_data()
+			.map(|d| RelayChainState {
+				number: d.relay_parent_number,
+				state_root: d.relay_parent_storage_root,
+			})
+			.unwrap_or_default()
+	}
+}
+
+/// Implements [`BlockNumberProvider`] and [`RelaychainStateProvider`] that returns relevant relay data fetched from
 /// validation data.
-/// NTOE: When validation data is not available (e.g. within on_initialize), 0 will be returned.
-pub struct RelaychainBlockNumberProvider<T>(sp_std::marker::PhantomData<T>);
+/// NTOE: When validation data is not available (e.g. within on_initialize), default values will be returned.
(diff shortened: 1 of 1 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2476494

pallets/parachain-system/src/lib.rs
```diff
@@ -1125,13 +1125,39 @@ pub trait OnSystemEvent {
 	/// Called when the validation code is being applied, aka from the next block on this is the new runtime.
 	fn on_validation_code_applied();
 }
+/// Holds the relay chain state root and block number.
```
rphmeier, line 1128: docs: worth mentioning that this is only the metadata about the most recent relay-parent of this fork of the parachain

pallets/parachain-system/src/lib.rs
```diff
@@ -1125,13 +1125,39 @@ pub trait OnSystemEvent {
...
+	}
+}
 
-/// Implements [`BlockNumberProvider`] that returns relay chain block number fetched from
+/// Implements [`BlockNumberProvider`] and [`RelaychainStateProvider`] that returns relevant relay data fetched from
 /// validation data.
-/// NTOE: When validation data is not available (e.g. within on_initialize), 0 will be returned.
-pub struct RelaychainBlockNumberProvider<T>(sp_std::marker::PhantomData<T>);
+/// NTOE: When validation data is not available (e.g. within on_initialize), default values will be returned.
+pub struct RelaychainDataProvider<T>(sp_std::marker::PhantomData<T>);
```
rphmeier, line 1158: For backwards compatibility, maybe better to keep a `RelaychainBlockNumberProvider` struct.

seunlanlege: Yeah makes sense

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2476535

rphmeier (approved):

bkchr:
You can just use `validation_data()` in your pallets. I don't see the reason for this trait. It is not that you will have any other source in your parachain that can return these values.

seunlanlege:
> You can just use `validation_data()` in your pallets. I don't see the reason for this trait. It is not that you will have any other source in your parachain that can return these values.

I think we should prefer loose coupling over tight coupling here, so pallets are immune to the implementation details of parachain-system

pallets/parachain-system/src/lib.rs
```diff
@@ -1139,6 +1162,47 @@ impl<T: Config> BlockNumberProvider for RelaychainBlockNumberProvider<T> {
...
+				number: d.relay_parent_number,
+				state_root: d.relay_parent_storage_root,
+			})
+			.unwrap_or_default()
+	}
+}
+
+/// Implements [`BlockNumberProvider`] and [`RelaychainStateProvider`] that returns relevant relay data fetched from
+/// validation data.
+/// NTOE: When validation data is not available (e.g. within on_initialize), default values will be returned.
```
bkchr, line 1194:
```suggestion
/// NOTE: When validation data is not available (e.g. within on_initialize), default values will be returned.
```

bkchr (approved):

pallets/parachain-system/src/lib.rs
```diff
@@ -1126,11 +1127,33 @@ pub trait OnSystemEvent {
...
+	pub number: relay_chain::BlockNumber,
+	/// State root for current relay chain height.
+	pub state_root: relay_chain::Hash,
+}
+
+/// This exposes the [`RelayChainState`] to other runtime modules.
+///
+/// Enables parachains to read relay chain state via state proofs.
+pub trait RelaychainStateProvider {
+	/// Maybe called by any runtime module to obtain the current state of the relay chain.
```
joepetrowski, line 1143:
```suggestion
	/// May be called by any runtime module to obtain the current state of the relay chain.
```

Polkadot-Forum:

https://forum.polkadot.network/t/state-proof-based-parachain-parachain-messaging/2214/1

PR #2270, opened 2023-03-03
Allow integrated relay chain light client

skunert:
In this PR, I added a command line flag `--relay-chain-rpc-client` to enable experimental support for an embedded light client.

## This leaves us with three execution modes for the relay chain:
1. Embedded polkadot full node
2. External polkadot node (using embedded mini-node for collation and pov-recovery)
3. Embedded light-client (using embedded mini-node for collation and pov-recovery)

## Implementation Overview
- I reused the existing RPC infrastructure as much as possible. Only the worker is switched out
- The light client  is run in-process and gets the chain spec that is passed via relay chain args to the collator
- We communicate with the help of the `jsonrpsee` library and let it handle subscription and request management
- The example light-client implementation in the smoldot repo uses an implementation of the platform trait based on async-std. Since we use tokio here, I converted the implementation 1:1 using tokio structures.

## Remaining issues
This issue is impacting us, I will take a look at what needs to be done on the substrate side:
- https://github.com/smol-dot/smoldot/issues/239


dmitry-markin (approved):
LGTM as far as I understood the underlying logic from the code of PR.

client/relay-chain-rpc-interface/src/light_client_worker.rs
```diff
@@ -0,0 +1,290 @@
...
+pub async fn build_smoldot_client(
+	spawner: SpawnTaskHandle,
+	chain_spec: &str,
+) -> RelayChainResult<(SmoldotClient<TokioPlatform, ()>, ChainId, JsonRpcResponses)> {
+	let config = ClientConfig {
+		tasks_spawner: Box::new(move |_, task| {
+			spawner.spawn("cumulus-relay-chain-light-client-task", None, task);
+		}),
+		system_name: "cumulus-relay-chain-light-client".to_string(),
+		system_version: env!("CARGO_PKG_VERSION").to_string(),
```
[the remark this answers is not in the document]
skunert: This works for binaries as well. `env!(CARGO_PKG_VERSION)` is replaced at compile time with the crate version.

client/cli/src/lib.rs
```diff
@@ -321,6 +329,9 @@ impl RunCmd {
 pub struct CollatorOptions {
```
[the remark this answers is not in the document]
skunert: These CollatorOptions are a pool of cli options that are collator-specific. By chance, the current ones are conflicting. But any other collator-specific option that can be added will also go here.

skunert:
I disabled collation for now, since we need the `persisted_validation_data` runtime call. This runtime call uses the state root host function which is currently not supported in smoldot. We should try again once https://github.com/smol-dot/smoldot/issues/177 is resolved.

client/relay-chain-rpc-interface/src/light_client_worker.rs
```diff
@@ -0,0 +1,290 @@
...
+		.add_chain(smoldot_light::AddChainConfig {
+			specification: chain_spec,
+			disable_json_rpc: false,
+			potential_relay_chains: core::iter::empty(),
+			database_content: "",
+			user_data: (),
+		})
+		.map_err(|e| RelayChainError::GenericError(e.to_string()))?;
+
+	Ok((client, chain_id, json_rpc_responses.expect("JSON RPC is not disabled; qed")))
```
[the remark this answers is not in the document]
skunert: Looking at the other examples in the code base, I think the pattern is that we write the reason why it can not fail. And in this case, JSON RPC is not disabled, which is the reason why it will not fail.

client/relay-chain-rpc-interface/src/light_client_worker.rs
```diff
@@ -0,0 +1,290 @@
...
+
+impl LightClientRpcWorker {
+	/// Create new light-client worker.
+	/// Returns the worker itself and a channel to send messages.
+	pub fn new(
+		smoldot_client: smoldot_light::Client<TokioPlatform, ()>,
+		json_rpc_responses: JsonRpcResponses,
+		chain_id: ChainId,
+	) -> (LightClientRpcWorker, TokioSender<RpcDispatcherMessage>) {
+		let (tx, rx) = tokio_channel(100);
```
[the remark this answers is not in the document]
skunert: This is arbitrary. In the team, we aligned on using bounded channels wherever possible, so I set this boundary here. I expect the worker to consume the messages pretty much instantly, so the limit should not be hit.

melekes (approved):
👍

client/relay-chain-rpc-interface/src/tokio_platform.rs
```diff
@@ -0,0 +1,329 @@
+// Copyright 2023 Parity Technologies (UK) Ltd.
```
tomaka, line 1: I just wanted to mention that the code that you're adapting is a bit of a "shitshow"

Since it's an implementation detail within smoldot I didn't pay much attention to it, but it's definitely meant to be rewritten at some point

client/relay-chain-rpc-interface/src/tokio_platform.rs
```diff
@@ -0,0 +1,329 @@
...
+struct StreamSharedGuarded {
+	write_queue: VecDeque<u8>,
+}
+
+/// Platform implementation for tokio
+/// This implementation is a conversion of the implementation for async-std:
+/// https://github.com/smol-dot/smoldot/blob/54d88891b1da202b4bf612a150df7b4dbfa03a55/light-base/src/platform/async_std.rs#L40
+pub struct TokioPlatform;
+
+impl smoldot_light::platform::Platform for TokioPlatform {
```
tomaka, line 55: As a heads up I'm going to change the trait a bit definition: https://github.com/smol-dot/smoldot/pull/315
The implementation in the smoldot repo is also much cleaner now.

skunert: Fair 👍, I will adjust here accordingly. Thanks for notifying!

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2624225

client/relay-chain-rpc-interface/src/tokio_platform.rs
```diff
@@ -0,0 +1,327 @@
...
+}
+
+struct StreamSharedGuarded {
+	write_queue: VecDeque<u8>,
+}
+
+/// Platform implementation for tokio
+/// This implementation is a conversion of the implementation for async-std:
+/// https://github.com/smol-dot/smoldot/blob/54d88891b1da202b4bf612a150df7b4dbfa03a55/light-base/src/platform/async_std.rs#L40
+impl smoldot_light::platform::Platform for TokioPlatform {
```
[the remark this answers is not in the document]
koute: > I would return the question and ask why rewrite this code for tokio when smoldot already provides something that works and is actively maintained.

In general I think it makes sense for us to standardize on a single async executor, and like it or not, everything runs on tokio these days.

Is there any particular reason why you're only supporting async-std and not tokio in smoldot? Considering that tokio is 1) faster than async-std, 2) more actively maintained (last updated 9 days ago vs last updated 12 months ago), 3) effectively "won" as the defacto Rust async executor, and 4) even has an LTS branch?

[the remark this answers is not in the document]
tomaka: Yes I have a strong reason: using a tokio socket panics if you don't use a tokio executor. Using tokio sockets from within a library is fundamentally a hack, because it ties a hidden implementation detail of the library with an implementation detail of the code that uses it.

Given that a light client connects to 3 peers, performance **doesn't matter**. At all. I want to strongly emphasize that. If you had a list of potential performance improvements to make, ordered from most impactful to least impactful, using tokio rather than async-std is probably item number 5000.

As for maintenance, having no maintenance is a good thing. It works. In which world code that is modified every week is considered better than code that is modified every year? Should we stop using zlib because it's unmaintained? I really don't understand this argument.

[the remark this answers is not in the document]
tomaka: As a heads up I have also renamed `AsyncStdTcpWebSocket` to `DefaultPlatform`: https://github.com/smol-dot/smoldot/pull/675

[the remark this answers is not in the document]
koute: > Yes I have a strong reason: using a tokio socket panics if you don't use a tokio executor. Using tokio sockets from within a library is fundamentally a hack, because it ties a hidden implementation detail of the library with an implementation detail of the code that uses it.

Do you mean that ideally it should return an error instead of panicking, or do you mean that it should implicitly and silently spawn an executor behind your back?

Nevertheless, in smolldot you already accept the platform trait which implies which executor is used, so how is this a *hidden* implementation detail?

> Given that a light client connects to 3 peers, performance **doesn't matter**.

Yes, I know. I was mostly speaking *in general*.

> As for maintenance, having no maintenance is a good thing. It works. In which world code that is modified every week is considered better than code that is modified every year?

Well, sure, fundamentally I agree with you here, *if* the code in question can be considered finished/done/complete/bugfree.

> Should we stop using zlib because it's unmaintained?

Yes, because the API's horrible and it's still getting CVEs.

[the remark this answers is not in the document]
tomaka: > Do you mean that ideally it should return an error instead of panicking, or do you mean that it should implicitly and silently spawn an executor behind your back?

Ideally it would just work, which could in principle technically be done with the help of the `std`.

A solution that would make tokio less bad could be to explicitly pass as parameter some kind of `RuntimeContext` object when creating a socket. It would give a compile-time guarantee that creating the socket will reasonably succeed, which is my pet peeve with tokio.

> Nevertheless, in smoldot you already accept the platform trait which implies which executor is used, so how is this a hidden implementation detail?

What is hidden is the fact that tokio uses a thread-local storage internally. Calling `TcpSocket::new` reads from that internal thread-local storage. This is a hidden behavior.

Unless I name the executor `PlatformThatRequiresYouToExecuteTheClientFromWithinATokioRuntime`, I will consider this a footgun. I don't consider it obvious for people that the word "tokio" implies "tokio must be used in your main application as well".

Cargo.toml
```diff
@@ -73,3 +73,266 @@ opt-level = 3
 inherits = "release"
 lto = true
 codegen-units = 1
+
+[patch."https://github.com/paritytech/substrate"]
```
skunert, line 77: This will be removed once I have figured out a suitable path for the wasmtime version situation.

client/cli/src/lib.rs
```diff
@@ -321,6 +329,9 @@ impl RunCmd {
...
 	/// Location of relay chain full node
 	pub relay_chain_rpc_urls: Vec<Url>,
+
+	/// Use embedded light client for the relay chain
+	pub relay_chain_light_client: bool,
 }
```
michalkucharczyk, line 335: Maybe this could go as enum, as the options are exclusive?
```
enum {
  EmbeddedNode,
  RpcNode(Vec<Url>),
  LightClientNode,
}
```

zombienet/tests/0002-pov_recovery.toml
```diff
@@ -73,3 +73,11 @@ add_to_genesis = false
   image = "{{COL_IMAGE}}"
   command = "test-parachain"
   args = ["-lparachain::availability=trace,sync=debug,parachain=debug,cumulus-pov-recovery=debug,cumulus-consensus=debug", "--disable-block-announcements", "--bootnodes {{'bob'|zombie('multiAddress')}}", "--relay-chain-rpc-url {{'ferdie'|zombie('wsUri')}}", "--", "--reserved-only", "--reserved-nodes {{'ferdie'|zombie('multiAddress')}}"]
+
+  # run two as a RPC parachain full node
```
michalkucharczyk, line 77:
```suggestion
  # run two nodes with light client
```

zombienet/tests/0003-full_node_catching_up.toml
```diff
@@ -40,3 +40,11 @@ cumulus_based = true
   image = "{{COL_IMAGE}}"
   command = "test-parachain"
   args = ["--reserved-only", "--reserved-nodes {{'charlie'|zombie('multiAddress')}}", "--relay-chain-rpc-url {{'alice'|zombie('wsUri')}}"]
+
+  # run cumulus eve (a parachain full node) and wait for it to sync some blocks
```
michalkucharczyk, line 44:
```suggestion
  # run cumulus node ferdie (with embedded light client) and wait for it to sync some blocks
```

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -40,35 +40,161 @@ use cumulus_primitives_core::{
...
+	chain_spec: String,
+	task_manager: &mut TaskManager,
+) -> RelayChainResult<RelayChainRpcClient> {
+	let (client, chain_id, json_rpc_responses) =
+		build_smoldot_client(task_manager.spawn_handle(), &chain_spec).await?;
+	let (worker, sender) = LightClientRpcWorker::new(client, json_rpc_responses, chain_id);
+
+	task_manager
+		.spawn_essential_handle()
+		.spawn("relay-chain-rpc-worker", None, worker.run());
```
michalkucharczyk, line 98:
```suggestion
		.spawn("relay-light-client-worker", None, worker.run());
```

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -40,35 +40,161 @@ use cumulus_primitives_core::{
...
+
+	let rpc_frontend = RpcFrontend::new(sender);
 
-	let client = RelayChainRpcClient::new(ws_client).await?;
+	let client = RelayChainRpcClient::new(rpc_frontend);
 
 	Ok(client)
 }
 
+/// Messages for communication between [`ReconnectingWsClient`] and [`ReconnectingWebsocketWorker`].
```
michalkucharczyk, line 107: `ReconnectingWsClient` seems not to be defined:
```suggestion
/// Messages for communication between [`RelayChainRpcClient`] and [`ReconnectingWebsocketWorker`].
```

Cargo.lock
davxy, line 1: This pulls-in quite a lot of new dependencies :-/
But I don't think we can do anything at this level.
Mostly are deps introduced by smoldot using different backend libs than the one we're using (e.g.  blake2-rfc and bip39)

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -40,35 +40,161 @@ use cumulus_primitives_core::{
...
+#[derive(Debug)]
+pub enum RpcDispatcherMessage {
+	RegisterBestHeadListener(Sender<RelayHeader>),
+	RegisterImportListener(Sender<RelayHeader>),
+	RegisterFinalizationListener(Sender<RelayHeader>),
+	Request(String, ArrayParams, OneshotSender<Result<JsonValue, JsonRpseeError>>),
```
davxy, line 113: Variants docs. E.g. what `String` stands or in the `Request` variant etc

client/relay-chain-rpc-interface/src/tokio_platform.rs
```diff
@@ -0,0 +1,397 @@
...
+	net::{IpAddr, SocketAddr},
+	pin::Pin,
+};
+use tokio::net::TcpStream;
+
+use tokio_util::compat::{Compat, TokioAsyncReadCompatExt};
+type CompatTcpStream = Compat<TcpStream>;
+
+/// Platform implementation for tokio
+/// This implementation is a conversion of the implementation for async-std:
```
davxy, line 36:
```suggestion
/// This implementation is a port of the implementation for async-std:
```

davxy:
LGTM mostly nitpicks

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3423953

client/relay-chain-rpc-interface/src/rpc_client.rs
```diff
@@ -40,35 +40,161 @@ use cumulus_primitives_core::{
...
 
 const LOG_TARGET: &str = "relay-chain-rpc-client";
+const NOTIFICATION_CHANNEL_SIZE_LIMIT: usize = 20;
 
 /// Client that maps RPC methods and deserializes results
 #[derive(Clone)]
 pub struct RelayChainRpcClient {
 	/// Websocket client to make calls
-	ws_client: ReconnectingWsClient,
+	ws_client: RpcFrontend,
```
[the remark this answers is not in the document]
skunert: I reconsidered this and agree that the `RpcFrontend` did not bring much benefit here. Since we already have too many abstraction layers, I scrapped it and merged into the `RelayChainRpcClient` as suggested.

PR #2274, opened 2023-03-04
`polkadot-service` deps

bkontur:


bkchr:
https://github.com/paritytech/cumulus/pull/2287 this is the better solution for the problem.

PR #2276, opened 2023-03-06
[Fix] Allow force_xcm_version call for assets parachains

ruseinov:
Same as https://github.com/paritytech/cumulus/pull/2222 but for statemine/t.


gilescope (approved):

joepetrowski (approved):

bkontur:
and what bridge-hubs, do we need that for them also?

joepetrowski:
Yeah I merged this too early. This is also the release branch, so also Collectives.

PR #2287, opened 2023-03-07
relay-chain-interface: Do not depend on `polkadot-service`

bkchr:


skunert (approved):

client/relay-chain-interface/src/lib.rs
```diff
@@ -61,10 +60,8 @@ pub enum RelayChainError {
 	WorkerCommunicationError(String),
 	#[error("Scale codec deserialization error: {0}")]
 	DeserializationError(CodecError),
-	#[error("Polkadot service error: {0}")]
-	ServiceError(#[from] polkadot_service::Error),
-	#[error("Substrate service error: {0}")]
-	SubServiceError(#[from] SubstrateServiceError),
```
ggwpez, line 67: You can also get that inner error from `sc_service` without including Polkadot, but it dont have a strong opinion on this code.

ggwpez (approved):
Thanks!

PR #2293, opened 2023-03-08
Proof of concept for deferred xcm messages in xcmp queue

jak-pan:
## XCMP Deferred queue

We have been hacking around the rate limits for XCMP messages together with @dmoka and @apopiak as a security measure and came up with a possible solution that utilizes generic XCMP filter and a queue for deferred messages. This is a draft implementation with a some stuff missing but we would like to get feedback before we continue to work on this.

## Why

The main reasoning behind this is that we would like to protect Polkadot and Kusama users from toxic flow / XCM in the ecosystem without hindering the UX too much. By implementing a queue on XCM messages we can postpone processing of messages that come to our chain or are leaving the chain. 

This filter is optional and completely customizable. If you don't implement it it will just pass the message through. 

### Volume limits
We will introduce reasonable inflow and outflow volume limits per block and also per asset per block. These should not be triggered in normal circumstances but if some suspiciously large transactions start entering or leaving our chain, we will postpone these. In case of a hack this will allow us to react and either pause these messages until the problem is resolved or completely discard them by utilizing governance (extrinsics are pending impl).

### Unsupported message queue
Instead of dumping unsupported messages (i.e. XCMv4 not yet implemented by us) we could add them to this queue for later execution. We can add None option to deferred_by that would require manual triggering, thus not losing yet unsupported messages which could be executed manually.

### Other
We can postpone controversial, more dangerous XCMs or XCMs made by other parachains to be able to check for potential attacks and be able to react.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -500,7 +500,7 @@ dependencies = [
 [[package]]
 name = "binary-merkle-tree"
 version = "4.0.0-dev"
-source = "git+https://github.com/paritytech/substrate?branch=master#8a4746e37c39dfa62552af0a47886dcc4ef1575f"
+source = "git+https://github.com/paritytech/substrate?branch=master#77c675b15cd805e5a9d61345763655ad8941591e"
 dependencies = [
  "hash-db",
  "log",
@@ -918,9 +918,9 @@ checksum = "e3b5ca7a04898ad4bcd41c90c5285445ff5b791899bb1b0abdd2a2aa791211d7"
 
 [[package]]
 name = "bytemuck"
-version = "1.13.0"
+version = "1.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c041d3eab048880cb0b86b256447da3f18859a163c3b8d8893f4e6368abe6393"
+checksum = "17febce684fd15d89027105661fec94afb475cb995fbc59d2865198446ba2eea"
 
 [[package]]
 name = "byteorder"
@@ -3226,7 +3226,7 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 [[package]]
 name = "fork-tree"
 version = "3.0.0"
-source = "git+https://github.com/paritytech/substrate?branch=master#8a4746e37c39dfa62552af0a47886dcc4ef1575f"
+source = "git+https://github.com/paritytech/substrate?branch=master#77c675b15cd805e5a9d61345763655ad8941591e"
 dependencies = [
  "parity-scale-codec",
 ]
@@ -3249,7 +3249,7 @@ checksum = "6c2141d6d6c8512188a7891b4b01590a45f6dac67afb4f255c4124dbb86d4eaa"
 [[package]]
 name = "frame-benchmarking"
 version = "4.0.0-dev"
-source = "git+https://github.com/paritytech/substrate?branch=master#8a4746e37c39dfa62552af0a47886dcc4ef1575f"
+source = "git+https://github.com/paritytech/substrate?branch=master#77c675b15cd805e5a9d61345763655ad8941591e"
 dependencies = [
  "frame-support",
  "frame-support-procedural",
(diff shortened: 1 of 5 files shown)


ggwpez:
https://github.com/paritytech/cumulus/pull/2157 is going to use a proper queue for the delayed message dispatch. The processing then happens in `on_initialize`.

>By implementing a queue on XCM messages we can postpone processing of messages that come to our chain or are leaving the chain.

There already is some kind of overweight message handling. With the MQ pallet this will be easier but the basic concept is that a message can be temporarily or permanently overweight. Temporary overweight messages will be re-tried, but permanently will be marked as such. They can lateron be processed manually with am [execute_overweight](https://github.com/paritytech/substrate/blob/306382bce6b21a8ddeaf9a3574557e51cc8a48b6/frame/message-queue/src/lib.rs#L612).

>Instead of dumping unsupported messages (i.e. XCMv4 not yet implemented by us) we could add them to this queue for later execution

Yea the MQ pallet and the current DMP queue discard those… idiomatically you could do just that; de-qeueue them and enqueue them into a specific "unsupported" queue. The MQ pallet supports multiple queues, but it would not be ideal from a weight perspective.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2499584

jak-pan:
> #2157 is going to use a proper queue for the delayed message dispatch. The processing then happens in `on_initialize`.
> 
> > By implementing a queue on XCM messages we can postpone processing of messages that come to our chain or are leaving the chain.
> 
> There already is some kind of overweight message handling. With the MQ pallet this will be easier but the basic concept is that a message can be temporarily or permanently overweight. Temporary overweight messages will be re-tried, but permanently will be marked as such. They can lateron be processed manually with am [execute_overweight](https://github.com/paritytech/substrate/blob/306382bce6b21a8ddeaf9a3574557e51cc8a48b6/frame/message-queue/src/lib.rs#L612).
> 
> > Instead of dumping unsupported messages (i.e. XCMv4 not yet implemented by us) we could add them to this queue for later execution
> 
> Yea the MQ pallet and the current DMP queue discard those… idiomatically you could do just that; de-qeueue them and enqueue them into a specific "unsupported" queue. The MQ pallet supports multiple queues, but it would not be ideal from a weight perspective.

I have had a look into message_queue and the PR and it doesn't really look like it does what we want at the moment. What we want to do is to defer message by specific time (or amount of  blocks) or potentially indefinitely if some custom condition is met and we don't want to execute them otherwise as explained in the examples.

As with overweight messages we still want to use the overweight logic that is independent of this.

rphmeier:
cc @ggwpez , can you take a look at this?

ggwpez:
So you need:
- Deferring messages and automatic execution of deferred messages at a later point in time.
- Stashing messages and manual execution of stashed messages.

I think the idea is fine. But implementing this in a PoV economical way would basically partially re-implement the MQ pallet.  
But it looks alright as a make-shift solution. Otherwise we can probably extend the MQ pallet to do both of that, not sure if we should @gavofyork.

jak-pan:
> So you need:
> 
> * Deferring messages and automatic execution of deferred messages at a later point in time.
> * Stashing messages and manual execution of stashed messages.
> 
> I think the idea is fine. But implementing this in a PoV economical way would partially re-implement the MQ pallet. But it looks alright as a make-shift solution. Otherwise we can probably extend the MQ pallet to do both of that, not sure if we should @gavofyork.
> 
> This will need to be rebased onto the other MR or master once it merges.

Yes you are right. As MQ pallet is will be done this is indeed a makeshift solution and could be done in a nicer way. 

We will add this (continue working on this) to our runtime anyway since we want to create protection mechanisms from toxic assets. Imagine a stablecoin is hacked and somebody wants to sell it to omnipool it's the easiest target since you have liquidity paired with all of the other assets and could be a risk to LPs. 

We are working on other mechanisms inside of the pool to amend this but this is really first line of defense.

Once we see unusually high volume of tokens flowing in or out of our parachain we can slow them down and if governance sees that this is indeed volume from a hack they can stop these messages without interrupting the protocol. 

The reason it's done like this is that it is as generic as possible and other chains can use this for different things. I.E. storing messages for later execution, handling unsupported messages or generally just deferring XCM (you could order XCM by priority)

The question is: do you want us to continue the work here and eventually merge it back or should we do it just for ourselves?

apopiak:
Another option would be to implement deferring of messages into the Substrate `message_queue` pallet (that only supports `yield`ing now, which AFAICT stops progress on that particular page of messages).

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2638473

Polkadot-Forum:

https://forum.polkadot.network/t/deferred-execution-of-xcmp-messages/2513/1

PR #2299, opened 2023-03-09
Update substrate and polkadot

slumber:
Align substrate head

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -500,7 +500,7 @@ dependencies = [
 [[package]]
 name = "binary-merkle-tree"
 version = "4.0.0-dev"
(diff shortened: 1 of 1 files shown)


bkchr (approved):
Any particular reason?

slumber:
> Any particular reason?

Building a companion for async backing and it's just easier to bump everywhere than to find an intersection point

bkchr:
> > Any particular reason?
> 
> Building a companion for async backing and it's just easier to bump everywhere than to find an intersection point

But for that you don't need this pr? You can just bump in your branch as you like. Going to close this, as you already started to force push because there are again changes in Substrate/Polkadot.

slumber:
I force pushed because of accident commit made to the wrong branch.
But the pr can be discarded since I pulled lockfile into companion

PR #2302, opened 2023-03-10
Cleanup dependencies + dead code

skunert:
- Clean up some unused dependencies.
- Remove `AuxStore` trait impl since we removed the subsystem that required it (also done in #2270, but I am waiting on that one)
- Remove some `polkadot-service` dependencies where possible. In the long run I would like to remove as many `polkadot-service` dependencies as possible to allow compilation of collator nodes without compiling polkadot-runtime (in rpc and light-client mode). Currently not possible as the minimal node depends on it.


dmitry-markin (approved):

melekes (approved):
👍

PR #2186, opened 2023-02-15
Fellowship into Collectives

muharem:

PR #2305, opened 2023-03-10
XCM tweaks for OpenGov

muharem:
Tweaks for https://github.com/paritytech/cumulus/pull/2186
Required for the test cases https://github.com/paritytech/parachains-integration-tests/pull/96

- Statemint allows unpaid execution for Fellows plurality 
- SafeCallFilter updated for Collectives and Statemint


parachains/runtimes/assets/statemint/src/xcm_config.rs
```diff
@@ -237,6 +241,9 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 				pallet_uniques::Call::set_collection_max_supply { .. } |
 				pallet_uniques::Call::set_price { .. } |
 				pallet_uniques::Call::buy_item { .. },
+			) |
+			RuntimeCall::PolkadotXcm(
+				pallet_xcm::Call::force_xcm_version { .. } | pallet_xcm::Call::send { .. },
```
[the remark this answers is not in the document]
muharem: yes, we. can just use `force_open_hrmp_channel`
removed send

parachains/runtimes/assets/statemint/src/xcm_config.rs
```diff
@@ -258,6 +263,8 @@ pub type Barrier = DenyThenTry<
 				AllowExplicitUnpaidExecutionFrom<ParentOrParentsPlurality>,
 				// Subscriptions for version tracking are OK.
 				AllowSubscriptionsFrom<ParentOrSiblings>,
+				// Fellows plurality gets free execution.
+				AllowExplicitUnpaidExecutionFrom<FellowsPlurality>,
```
KiChjang, line 267: Huh, does this mean that even the root of Polkadot doesn't get free execution and only the technical fellowship does?

joepetrowski: Root should, but wouldn't `AllowExplicitUnpaidExecutionFrom<ParentOrParentsPlurality>` allow that?

KiChjang: Ah, sorry I missed it -- it would seem like we should just have 1 `AllowExplicitUnpaidExecutionFrom` barrier instead of 2.

KiChjang (approved):

joepetrowski (approved):

PR #2308, opened 2023-03-11
Improve build times by disabling wasm-builder in `no_std`

bkchr:
There is no need to compile wasm-builder in `no_std` as we skip the compilation of the crate any way. So, we can also directly disable wasm-builder as dependency in `no_std` to improve build times.

### DIFF
--- a/parachains/runtimes/testing/rococo-parachain/build.rs
+++ b/parachains/runtimes/testing/rococo-parachain/build.rs
@@ -1,25 +1,26 @@
-// Copyright 2019-2021 Parity Technologies (UK) Ltd.
-// This file is part of Cumulus.
+// Copyright (C) Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
 
-// Substrate is free software: you can redistribute it and/or modify
-// it under the terms of the GNU General Public License as published by
-// the Free Software Foundation, either version 3 of the License, or
-// (at your option) any later version.
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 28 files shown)


xlc:
parachain teams should do similar change to improve build time?

bkchr:
> parachain teams should do similar change to improve build time?

You don't need to, but you can.

parachains/runtimes/testing/rococo-parachain/build.rs
```diff
@@ -1,25 +1,26 @@
-// Copyright 2019-2021 Parity Technologies (UK) Ltd.
-// This file is part of Cumulus.
+// Copyright (C) Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: Apache-2.0
```
mrcnski, line 2: We're not putting the year anymore?

I think all the licenses should be updated in a dedicated PR, as this distracts from other changes here and is incomplete in itself. We have a [license scanner](https://github.com/paritytech/license-scanner) tool for this which we already used in Polkadot. cc @rzadp

bkchr: Not really sure that it distracts from the changes. Yes we are not putting the year anymore.

koute (approved):

skunert (approved):

parachains/runtimes/testing/rococo-parachain/build.rs
rzadp: Hey, feel free to ask me to set up the licence scanner CI check here if it has proven useful in Polkadot.

PR #2316, opened 2023-03-14
Benchmarks for pallet-bridge-assets-transfer

svyatonik:
Important things in this PR:
- I've added `open_outbound_hrmp_channel_for_benchmarks` to the `cumulus-pallet-parachain-system`. Why: I need to "
open" the HRMP channel between Statemin{t/e} and sibling bridge hub in benchmarks. There's that `asset_test_utils::mock_open_hrmp_channel`, but it isn't available in no-std environment because of proof generation (e.g. `sp_state_machine::prove_read`). I could try to fix it to use `no-std` primitives only (there could be some caveats rhough) - feel free to insist on that && I'll do some exploration;
- I've seen @bkontur that you were planning to make `transfer_asset_via_bridge` weight depend on the number of transferred assets, but imo the weight of the call doesn't depend on that at all - you just schedule the XCM message in the queue and that's just it. Please comment if you think that's wrong;
- I had to add this `BenchmakrHelper` - otherwise there's no way to get the proper arguments for the `transfer_asset_via_bridge` call (all other benchmarks can be measured without that), because we're using the whole XCM stuff (including exporters, executers, ...) there.

What is left here:
- [x] to run benchmarks using bench bot


svyatonik:
/cmd queue -c bench $ pallet assets statemine pallet_bridge_assets_transfer

ggwpez:
bot help

svyatonik:
bot bench $ pallet statemine assets pallet_bridge_assets_transfer

parachains/pallets/bridge-assets-transfer/src/lib.rs
```diff
@@ -67,6 +69,28 @@ pub mod pallet {
...
+		fn bridge_config() -> (NetworkId, BridgeConfig);
+		/// Prepare environment for assets transfer and return transfer origin and assets
+		/// to transfer. After this function is called, we expect `transfer_asset_via_bridge`
+		/// to succeed, so in proper environment, it should:
+		///
+		/// - deposit enough funds (fee from `bridge_config()` and transferred assets) to the sender account;
+		///
+		/// - ensure that the `BridgeXcmSender` is properly configured for the transfer;
+		///
+		/// - be close to the worst possible scenario - i.e. if some account may need to be created during
```
svyatonik, line 88: I've also been thinking about supporting different bridges here - e.g. in our case message will be sent to bridge hub first. But what if in other case message needs to be sent using some other queueing mechanism? E.g. when it is deployed at the same chain where `pallet-bridge-messages` is deployed or where eth bridge is deployed? Right now the assumption is that we won't do any refunds + the worst case needs to be "prepared" here. If we need to support such ^^^ case, then we may need to change sending interface a bit (to return actual used weight)

svyatonik: > as we expect eth bridge deployed on bridge-hub (now).

Sorry - my comment is ambiguous. It was about the case when it works in a different way :) I mean - in our case it doesn't matter, but since anyone may use this pallet at their own parachain and use in a different way, we may think about it in advance. But let's just stay with our own case for now

> @svyatonik Slava, what do you mean by assumption is that we won't do any refunds? Does it mean, that we dont do any withdrawal for BuyExecution instruction here?

No - I mean that if there are several bridges and using first is `1_000_000` weight units and second is `5_000_000`, then we could return worst case from there and then return actual used weight from the call (i.e. "refund" the excessive cost of the call). But again - for our case it isn't an issue, since we're going to use the same transport everywhere

svyatonik: TLDR: nvm for now - it is irrelevant for our planned deployment

bkontur (approved):
lgtm, thank you :)

parachains/pallets/bridge-assets-transfer/src/lib.rs
bkontur: > Sorry - my comment is ambiguous. It was about the case when it works in a different way :) I mean - in our case it doesn't matter, but since anyone may use this pallet at their own parachain and use in a different way, we may think about it in advance. But let's just stay with our own case for now

yes, I was also thinking about different usages of this pallet, with `type BridgeXcmSender: SendXcm` you can send it everywhere (to sibling, to relay, to some local pallets..).
what would you suggest here to better support "use in a different way"?

svyatonik: Something like having a dedicated trait that returns actual/unspent weight when XCM is sent (and of course is using some `SendXcm` implementation inside). So the weight could be different for every possible route (HRMP/local pallet call). Then we could return post-dispatch weight from the call. But yeah - let's not do this now :)

svyatonik:
bot bench $ pallet statemine assets pallet_bridge_assets_transfer

svyatonik:
Actually, after thinking about it a bit, there should be some performance dependency on number of sent assets - every asset has its configuration and its own ledger. So at least additional db reads should be there. Gonna check that

svyatonik:
@bkontur So right now pallet rejects transfers if count of wrapped assets is not equal to `1`. I've added prototype for benchmarking multi assets - we'll need to work with it once pallet is completed.

PR #2342, opened 2023-03-17
Increase xcm fees when busy

gilescope:
This is the corollary to https://github.com/paritytech/polkadot/pull/6843 . Xcmp fees upwards and horizontally get exponentially more expensive as the queues increase.

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -53,6 +53,7 @@ use sp_runtime::{
 		InvalidTransaction, TransactionLongevity, TransactionSource, TransactionValidity,
 		ValidTransaction,
 	},
+	FixedU128, Saturating,
 };
 use sp_std::{cmp, collections::btree_map::BTreeMap, prelude::*};
 use xcm::latest::XcmHash;
@@ -90,6 +91,8 @@ pub use cumulus_pallet_parachain_system_proc_macro::register_validate_block;
 pub use relay_state_snapshot::{MessagingStateSnapshot, RelayChainStateProof};
 
 pub use pallet::*;
+
+pub const MULTIPLICATIVE_FEE_FACTOR_UMP: FixedU128 = FixedU128::from_rational(101, 100); // 1.01
 
 /// Something that can check the associated relay block number.
 ///
@@ -261,6 +264,9 @@ pub mod pallet {
 
 				UpwardMessages::<T>::put(&up[..num]);
 				*up = up.split_off(num);
+				if num > 0 {
+					Self::decrement_ump_fee_factor();
+				}
 			});
 
 			// Sending HRMP messages is a little bit more involved. There are the following
@@ -691,6 +697,16 @@ pub mod pallet {
 	#[pallet::storage]
 	pub(super) type CustomValidationHeadData<T: Config> = StorageValue<_, Vec<u8>, OptionQuery>;
 
+	frame_support::parameter_types! {
+		pub InitialFactor: FixedU128 = FixedU128::from_u32(1);
+	}
+
+	/// The number to multiply the base delivery fee by.
+	#[pallet::storage]
+	#[pallet::getter(fn delivery_fee_factor_ump)]
+	pub(crate) type DeliveryFeeFactorUMP<T: Config> =
+		StorageValue<_, FixedU128, ValueQuery, InitialFactor>;
+
 	#[pallet::inherent]
 	impl<T: Config> ProvideInherent for Pallet<T> {
 		type Call = Call<T>;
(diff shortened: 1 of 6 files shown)


pallets/parachain-system/src/lib.rs
```diff
@@ -691,6 +697,16 @@ pub mod pallet {
...
 	pub(super) type CustomValidationHeadData<T: Config> = StorageValue<_, Vec<u8>, OptionQuery>;
 
+	frame_support::parameter_types! {
+		pub InitialFactor: FixedU128 = FixedU128::from_u32(1);
+	}
+
+	/// The number to multiply the base delivery fee by.
+	#[pallet::storage]
+	#[pallet::getter(fn delivery_fee_factor_ump)]
+	pub(crate) type DeliveryFeeFactorUMP<T: Config> =
```
KiChjang, line 707:
```suggestion
	pub(crate) type DeliveryFeeFactorUmp<T: Config> =
```

pallets/parachain-system/src/lib.rs
```diff
@@ -1069,7 +1085,7 @@ impl<T: Config> Pallet<T> {
 		match Self::host_configuration() {
 			Some(cfg) =>
 				if message.len() > cfg.max_upward_message_size as usize {
-					return Err(MessageSendError::TooBig)
+					Self::increment_ump_fee_factor();
```
KiChjang, line 1088: This is the wrong place to put the increment. The condition here checks whether the individual message size is greater than the limit, and does _not_ check for the queue size.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -374,6 +376,16 @@ pub mod pallet {
...
 	pub(super) type QueueSuspended<T: Config> = StorageValue<_, bool, ValueQuery>;
+
+	frame_support::parameter_types! {
+		pub InitialFactor: FixedU128 = FixedU128::from_u32(1);
+	}
+
+	/// The number to multiply the base delivery fee by.
+	#[pallet::storage]
+	#[pallet::getter(fn delivery_fee_factor_xcmp)]
+	pub(crate) type DeliveryFeeFactorXCMP<T: Config> =
```
KiChjang, line 387:
```suggestion
	pub(crate) type DeliveryFeeFactorXcmp<T: Config> =
```

KiChjang: Since you're in the XCMP pallet in this file, it doesn't require specifying which pallet that this delivery fee factor belongs to.

pallets/xcmp-queue/src/lib.rs
```diff
@@ -1129,6 +1163,9 @@ impl<T: Config> XcmpMessageSource for Pallet<T> {
 
 		<OutboundXcmpStatus<T>>::put(statuses);
 
+		if !result.is_empty() {
```
KiChjang, line 1166: It should check whether the number of messages in queue is less than the maximum limit.

pallets/parachain-system/src/lib.rs
```diff
@@ -261,6 +264,9 @@ pub mod pallet {
 
 				UpwardMessages::<T>::put(&up[..num]);
 				*up = up.split_off(num);
+				if num > 0 {
```
KiChjang, line 267: This should check whether the UMP queue is less than then max limit.

xlc:
Please make sure there is an easy way to estimate XCM fee https://github.com/paritytech/polkadot/issues/6883

Also please consider the scenario that the XCM is triggered indirectly. For example, from a smart contract. The smart contract will pay for the XCM fee so it will need to know how much to charge from the user for the XCM fee.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2546897

gilescope:
This just does not translate well from the polkadot dmp PR. The polkadot DMP pallet does not page, but holds a list of messages where as the xcmp-queue holds pages of messages. Establishing how many messages are in a page when it gets sent would require decoding all the page or keeping a tally separately, so it's probably nicer to bump the xcm fees based on the number of pages queued up which is readily available information. I.e. once one page is full then xcm messages get more expensive, and once two pages get full then the channel suspends while it clears down (2 pages is the current `suspendThreshold`).

gilescope:
Let's split off the xcmp xcm fees into a separate PR as they're going to have to be treated differently given that it's paged. But the UMP can be treated in the same way as the relay is treating the DMP.

parachains/runtimes/assets/statemine/src/lib.rs
```diff
@@ -475,7 +477,8 @@ impl cumulus_pallet_xcmp_queue::Config for Runtime {
 	>;
 	type ControllerOriginConverter = xcm_config::XcmOriginToTransactDispatchOrigin;
 	type WeightInfo = weights::cumulus_pallet_xcmp_queue::WeightInfo<Runtime>;
-	type PriceForSiblingDelivery = ();
+	type PriceForSiblingDelivery =
+		SiblingExponentialPrice<FeeAssetId, BaseDeliveryFee, TransactionByteFee, XcmpFeeFactor>;
```
bkontur, line 481: when https://github.com/paritytech/polkadot/pull/7005 and https://github.com/paritytech/polkadot/pull/7585,
I am goging to reuse here lots of stuff from `runtime/common` e.g. `ExponentialPrice / PriceForParachainDelivery ...`

Issue #2320, opened 2023-03-15
Burning fee allowed for system parachain?


xlc:
https://github.com/paritytech/cumulus/blob/66887505e6c5ef89ce9f88fc4ec368eb858b1b92/parachains/runtimes/assets/statemint/src/xcm_config.rs#L314

To my understanding, this will burn all XCM fees.

Two questions:

- Shouldn't fee be sent to collator?
- Can system parachain allowed to burn relaychain tokens (without notifying relaychain)? Wouldn't this cause imbalances? i.e. balance mismatch between relaychain's checking account and parachain's total issuance.

joepetrowski:
Yes, I agree

gavofyork:
Yes, when `FeeManager` is used, they would be better sent to a proper account.

PR #2358, opened 2023-03-21
FeeManager that puts fees in a central account

gilescope:
Similar to XcmFeesTo32ByteAccount, an implementation of FeeManager that routes XCM fees to a central account.

Associated issue: https://github.com/paritytech/cumulus/issues/2320

### DIFF
--- a/primitives/utility/src/lib.rs
+++ b/primitives/utility/src/lib.rs
@@ -33,7 +33,7 @@ use sp_runtime::{traits::Saturating, SaturatedConversion};
 use sp_std::{marker::PhantomData, prelude::*};
 use xcm::{latest::prelude::*, WrapVersion};
 use xcm_builder::TakeRevenue;
-use xcm_executor::traits::{MatchesFungibles, TransactAsset, WeightTrader};
+use xcm_executor::traits::{FeeManager, FeeReason, MatchesFungibles, TransactAsset, WeightTrader};
 
 pub trait PriceForParentDelivery {
 	fn price_for_parent_delivery(message: &Xcm<()>) -> MultiAssets;
@@ -270,7 +270,7 @@ impl<
 }
 
 /// XCM fee depositor to which we implement the TakeRevenue trait
-/// It receives a Transact implemented argument, a 32 byte convertible acocuntId, and the fee receiver account
+/// It receives a Transact implemented argument, a 32 byte convertible accountId, and the fee receiver account
 /// FungiblesMutateAdapter should be identical to that implemented by WithdrawAsset
 pub struct XcmFeesTo32ByteAccount<FungiblesMutateAdapter, AccountId, ReceiverAccount>(
 	PhantomData<(FungiblesMutateAdapter, AccountId, ReceiverAccount)>,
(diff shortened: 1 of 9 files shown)


primitives/utility/src/lib.rs
```diff
@@ -297,6 +297,43 @@ impl<
...
+
+impl<FungiblesMutateAdapter, AccountId, ReceiverAccount> FeeManager
+	for XcmFeeManagerToAccount<FungiblesMutateAdapter, AccountId, ReceiverAccount>
+where
+	FungiblesMutateAdapter: xcm_executor::traits::TransactAsset,
+	AccountId: Clone + Into<[u8; 32]>,
+	ReceiverAccount: frame_support::traits::Get<Option<AccountId>>,
+{
+	fn is_waived(_origin: Option<&MultiLocation>, _: FeeReason) -> bool {
+		false
```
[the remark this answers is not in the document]
joepetrowski: It does take an `origin`, so it is possible.

primitives/utility/src/lib.rs
```diff
@@ -297,6 +297,43 @@ impl<
...
+{
+	_phantom: PhantomData<(FungiblesMutateAdapter, AccountId, ReceiverAccount)>,
+}
+
+impl<FungiblesMutateAdapter, AccountId, ReceiverAccount> FeeManager
+	for XcmFeeManagerToAccount<FungiblesMutateAdapter, AccountId, ReceiverAccount>
+where
+	FungiblesMutateAdapter: xcm_executor::traits::TransactAsset,
+	AccountId: Clone + Into<[u8; 32]>,
+	ReceiverAccount: frame_support::traits::Get<Option<AccountId>>,
```
bkontur, line 312: what about to change `ReceiverAccount: frame_support::traits::Get<Option<MultiLocation>>`?
it will be more general and xcm-like,

and `let dest = X1(AccountId32 { network: None, id: RELAY_TREASURY_PALLET_ID.into_account_truncating() }).into();`  can be configured also in runtime

xlc: I don't think that's a good idea. And if you really need this feature for whatever reason, just implement it on a new struct.

primitives/utility/src/lib.rs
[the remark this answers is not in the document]
gilescope: Fee free for relay chain governance seems sensible. I can't envision a reason not to have that. Have updated the code accordingly.

primitives/utility/src/lib.rs
```diff
@@ -297,6 +297,43 @@ impl<
...
+{
+	fn is_waived(origin: Option<&MultiLocation>, _: FeeReason) -> bool {
+		matches!(origin, Some(MultiLocation { parents: 1, interior: Here }))
```
xlc, line 315: This isn't exactly what I wanted. This makes it only useable for system parachain.
It will be better if you pass a generic `WavedOrigin: Contains<(Option<MultiLocation>, FeeReason)>` and have a `ParentOrigin`

gilescope: Tbh I think that it's simpler to just write a different FeeManager implementation if you want custom logic. It's only 2 functions and if we're delegating one of them, are we implementing much? I think I'd prefer it stays simpler and just used as a reference example of how to implement a FeeManager. Indirection is not without cost in terms of comprehension.

xlc: you have a point. it is just a lot of XCM is built with this pattern.

primitives/utility/src/lib.rs
```diff
@@ -297,6 +297,43 @@ impl<
...
+	fn handle_fee(fee: MultiAssets) {
+		if let Some(receiver) = ReceiverAccount::get() {
+			let dest = X1(AccountId32 { network: None, id: receiver.into() }).into();
+			for asset in fee.into_inner() {
+				let ok = FungiblesMutateAdapter::deposit_asset(
+					&asset,
+					&dest,
+					// We aren't able to track the XCM that initiated the fee deposit, so we create a
+					// fake message hash here
+					&XcmContext::with_message_hash([0; 32]),
```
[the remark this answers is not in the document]
gilescope: Keith has a PR to fix this here: https://github.com/paritytech/polkadot/pull/7005

PR #2438, opened 2023-04-06
(async backing) parachain-system: track limitations for unincluded blocks

### LINKED ISSUE #1866: Computation of real message bandwidth available in asynchronous backing
The parachain-system pallet should keep track of the 'unincluded segment' of the parachain. This is all recent blocks which have not been included in a referenced relay-parent.

For each block in the unincluded segment, parachain-system should track:
  * Outgoing UMP message count, size
  * Outgoing HRMP message count, size per recipient
  * Stated HRMP watermark
  * DMP messages processed

We also keep a struct aggregating the data from the whole unincluded segment:
  * Total outgoing UMP message count, size in the unincluded segment
  * Total outgoing HRMP message count, size per recipient in the unincluded segment
  * Highest HRMP watermark in the unincluded segment
  * Total DMP messages processed

Note that if the unincluded segment of the parachain is empty, this is equivalent to the protocol before asynchronous backing, where we only build parablocks once the parent parablock has been included and therefore the unincluded segment is always empty.

On initialization of parachain-system, we:
  1. Read a storage proof of the latest head of the para included under the relay-parent
  2. Discard everything in the unincluded segment up to and including that head
  3. Update the aggregate struct

The aggregate struct will be used to further limit message bandwidth and processing beyond just what is stated by the relay-parent storage proofs

slumber:
Resolves #1866

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -301,7 +301,7 @@ pub mod pallet {
 
 			// If `MaxUnincludedLen` is present in the storage, parachain head
 			// is always expected to be included into the relay storage proof.
-			let para_head = <MaxUnincludedLen<T>>::get().map(|_max_len| {
+			let para_head_with_len = <MaxUnincludedLen<T>>::get().map(|max_len| {
 				let relay_chain_state = Self::relay_state_proof()
 					.expect("relay state proof must be present in storage");
 				let validation_data =
@@ -313,38 +313,54 @@ pub mod pallet {
 					relay_chain_state,
 				)
 				.expect("Invalid relay chain state proof");
-				relay_state_proof
-					.read_included_para_head()
-					.expect("Invalid para head in relay chain state proof")
+				(
+					relay_state_proof
+						.read_included_para_head()
+						.expect("Invalid para head in relay chain state proof"),
+					max_len,
+				)
 			});
 			weight += T::DbWeight::get().reads(1);
 
 			// Update unincluded segment related storage values.
-			if let Some(para_head) = para_head {
+			if let Some((para_head, max_len)) = para_head_with_len {
 				// Weight used for reading para head.
 				weight += T::DbWeight::get().reads(2);
 
-				let dropped: Vec<Ancestor> = <UnincludedSegment<T>>::mutate(|chain| {
-					// Drop everything up to the block with an included para head, if present.
-					let idx = chain
-						.iter()
-						.position(|block| block.para_head() == &para_head)
-						.map_or(0, |idx| idx + 1); // inclusive.
+				if max_len > 0u32.into() {
+					let (dropped, left_count): (Vec<Ancestor>, u32) =
+						<UnincludedSegment<T>>::mutate(|chain| {
+							// Drop everything up to the block with an included para head, if present.
+							let idx = chain
+								.iter()
+								.position(|block| block.para_head() == &para_head)
+								.map_or(0, |idx| idx + 1); // inclusive.
 
-					chain.drain(..idx).collect()
-				});
-				weight += T::DbWeight::get().reads_writes(1, 1);
-
-				if !dropped.is_empty() {
-					<AggregatedUnincludedSegment<T>>::mutate(|agg| {
-						let agg = agg.as_mut().expect(
-							"dropped part of the segment wasn't empty, hence value exists; qed",
-						);
-						for block in dropped {
-							agg.subtract(&block);
-						}
-					});
+							let left_count = (idx..chain.len()).count() as u32;
+							let dropped = chain.drain(..idx).collect();
+							(dropped, left_count)
+						});
 					weight += T::DbWeight::get().reads_writes(1, 1);
+
+					// sanity-check there's place for the block at finalization phase.
+					//
+					// TODO: this potentially restricts parachains from decreasing `MaxUnincludedLen` value.
+					assert!(
+						max_len > left_count.into(),
+						"no space left for the block in the unincluded segment"
+					);
+
+					if !dropped.is_empty() {
+						<AggregatedUnincludedSegment<T>>::mutate(|agg| {
+							let agg = agg.as_mut().expect(
+								"dropped part of the segment wasn't empty, hence value exists; qed",
+							);
+							for block in dropped {
+								agg.subtract(&block);
+							}
+						});
+						weight += T::DbWeight::get().reads_writes(1, 1);
+					}
 				}
 			}
 
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -348,7 +348,7 @@ pub mod pallet {
 				weight += T::DbWeight::get().reads_writes(1, 1);
 
 				// Weight used during finalization.
-				weight += T::DbWeight::get().reads_writes(4, 2);
+				weight += T::DbWeight::get().reads_writes(2, 2);
 			}
 			weight += T::DbWeight::get().reads(1);
 
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -65,7 +65,7 @@ pub mod validate_block;
 #[cfg(test)]
 mod tests;
 
-use unincluded_segment::{BlockTracker, SegmentTracker};
+use unincluded_segment::{Ancestor, SegmentTracker};
 
 /// Register the `validate_block` function that is used by parachains to validate blocks on a
 /// validator.
@@ -319,7 +319,7 @@ pub mod pallet {
 
 			// Update unincluded segment related storage values.
 			if let Some(para_head) = para_head {
-				let dropped: Vec<BlockTracker> = <UnincludedSegment<T>>::mutate(|chain| {
+				let dropped: Vec<Ancestor> = <UnincludedSegment<T>>::mutate(|chain| {
 					// Drop everything up to the block with an included para head, if present.
 					let idx = chain
 						.iter()
@@ -592,7 +592,7 @@ pub mod pallet {
 	}
 
 	#[pallet::storage]
-	pub(super) type UnincludedSegment<T: Config> = StorageValue<_, Vec<BlockTracker>, ValueQuery>;
+	pub(super) type UnincludedSegment<T: Config> = StorageValue<_, Vec<Ancestor>, ValueQuery>;
 
 	#[pallet::storage]
 	pub(super) type AggregatedUnincludedSegment<T: Config> =
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -1096,7 +1096,8 @@ impl<T: Config> Pallet<T> {
 
 			// sanity-check there's place for the block at finalization phase.
 			//
-			// TODO: this potentially restricts parachains from decreasing `MaxUnincludedLen` value.
+			// If this fails, the max segment len is reached and parachain should wait
+			// for ancestor's inclusion.
 			assert!(
 				max_len > left_count.into(),
 				"no space left for the block in the unincluded segment"
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -302,6 +302,8 @@ pub mod pallet {
 				// This is correct because:
 				// - inherent never contains messages that were previously processed.
 				// - current implementation always attempts to exhaust each message queue.
+				//
+				// <https://github.com/paritytech/cumulus/issues/2472>
 				let limits = TotalBandwidthLimits::new(&relevant_messaging_state);
 
 				let hrmp_outgoing = outbound_messages
@@ -437,7 +439,7 @@ pub mod pallet {
 			// the inherent shouldn't contain messages that were already processed by any of the
 			// ancestors.
 			//
-			// This invariant should be upheld by the node-side.
+			// This invariant should be upheld by the `ProvideInherent` implementation.
 			let ParachainInherentData {
 				validation_data: vfp,
 				relay_chain_state,
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -335,7 +335,7 @@ pub mod pallet {
 							"dropped part of the segment wasn't empty, hence value exists; qed",
 						);
 						for block in dropped {
-							agg.used_bandwidth.subtract(block.used_bandwidth());
+							agg.subtract(&block);
 						}
 					});
 					weight += T::DbWeight::get().reads_writes(1, 1);
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -320,10 +320,11 @@ pub mod pallet {
 			// Update unincluded segment related storage values.
 			if let Some(para_head) = para_head {
 				let dropped: Vec<BlockTracker> = <UnincludedSegment<T>>::mutate(|chain| {
-					let idx =
-						chain.iter().take_while(|block| block.para_head() != &para_head).count();
-					// Drop the block with an included para head too.
-					let idx = chain.len().min(idx + 1);
+					// Drop everything up to the block with an included para head, if present.
+					let idx = chain
+						.iter()
+						.position(|block| block.para_head() == &para_head)
+						.map_or(0, |idx| idx + 1); // inclusive.
 
 					chain.drain(..idx).collect()
 				});
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -312,7 +312,7 @@ pub mod pallet {
 				)
 				.expect("Invalid relay chain state proof");
 				relay_state_proof
-					.read_para_head()
+					.read_included_para_head()
 					.expect("Invalid para head in relay chain state proof")
 			};
 			weight += T::DbWeight::get().reads(2);
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -596,12 +596,24 @@ pub mod pallet {
 		Unauthorized,
 	}
 
+	/// Maximum number of latest included block descendants the runtime is allowed to accept. In other words,
+	/// these are ancestor of the block being currently executed, not yet sent to the relay chain runtime.
+	///
+	/// This value is optional, but once set to `Some` by the governance, should never go back to `None`.
+	/// Requires latest included para head to be present in the relay chain storage proof.
 	#[pallet::storage]
 	pub(super) type MaxUnincludedLen<T: Config> = StorageValue<_, T::BlockNumber, OptionQuery>;
 
+	/// Latest included block descendants the runtime accepted. In other words, these are
+	/// ancestor of the block being currently executed, not yet sent to the relay chain runtime.
+	///
+	/// The segment length is limited by [`MaxUnincludedLen`].
 	#[pallet::storage]
 	pub(super) type UnincludedSegment<T: Config> = StorageValue<_, Vec<Ancestor>, ValueQuery>;
 
+	/// Storage field that keeps track of bandwidth used by the unincluded segment
+	/// along with the latest HRMP watermark. Used for limiting the acceptance of new
+	/// blocks with respect to relay chain constraints.
 	#[pallet::storage]
 	pub(super) type AggregatedUnincludedSegment<T: Config> =
 		StorageValue<_, SegmentTracker, OptionQuery>;
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -343,74 +343,6 @@ pub mod pallet {
 				weight += T::DbWeight::get().writes(1);
 			}
 
-			// If `MaxUnincludedLen` is present in the storage, parachain head
-			// is always expected to be included into the relay storage proof.
-			let para_head_with_len = <MaxUnincludedLen<T>>::get().map(|max_len| {
-				let relay_chain_state = Self::relay_state_proof()
-					.expect("relay state proof must be present in storage");
-				let validation_data =
-					Self::validation_data().expect("validation data must be present in storage");
-
-				let relay_state_proof = RelayChainStateProof::new(
-					T::SelfParaId::get(),
-					validation_data.relay_parent_storage_root,
-					relay_chain_state,
-				)
-				.expect("Invalid relay chain state proof");
-				(
-					relay_state_proof
-						.read_included_para_head()
-						.expect("Invalid para head in relay chain state proof"),
-					max_len,
-				)
-			});
-			weight += T::DbWeight::get().reads(1);
-
-			// Update unincluded segment related storage values.
-			if let Some((para_head, max_len)) = para_head_with_len {
-				// Weight used for reading para head.
-				weight += T::DbWeight::get().reads(2);
-
-				if !max_len.is_zero() {
-					// Weight used by `on_finalize` in case the len is non-zero.
-					weight += T::DbWeight::get().reads_writes(4, 2);
-
-					let (dropped, left_count): (Vec<Ancestor>, u32) =
-						<UnincludedSegment<T>>::mutate(|chain| {
-							// Drop everything up to the block with an included para head, if present.
-							let idx = chain
-								.iter()
-								.position(|block| block.para_head() == &para_head)
-								.map_or(0, |idx| idx + 1); // inclusive.
-
-							let left_count = (idx..chain.len()).count() as u32;
-							let dropped = chain.drain(..idx).collect();
-							(dropped, left_count)
-						});
-					weight += T::DbWeight::get().reads_writes(1, 1);
-
-					// sanity-check there's place for the block at finalization phase.
-					//
-					// TODO: this potentially restricts parachains from decreasing `MaxUnincludedLen` value.
-					assert!(
-						max_len > left_count.into(),
-						"no space left for the block in the unincluded segment"
-					);
-
-					if !dropped.is_empty() {
-						<AggregatedUnincludedSegment<T>>::mutate(|agg| {
-							let agg = agg.as_mut().expect(
-								"dropped part of the segment wasn't empty, hence value exists; qed",
-							);
-							for block in dropped {
-								agg.subtract(&block);
-							}
-						});
-						weight += T::DbWeight::get().reads_writes(1, 1);
-					}
-				}
-			}
-
 			// Remove the validation from the old block.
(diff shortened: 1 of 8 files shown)


pallets/parachain-system/src/unincluded_segment.rs
```diff
@@ -0,0 +1,117 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+use codec::{Decode, Encode};
+use cumulus_primitives_core::{relay_chain, ParaId};
+use scale_info::TypeInfo;
+use sp_std::collections::btree_map::BTreeMap;
+
+pub struct BandwidthLimits {}
```
slumber, line 22: Not all limits mentioned in the issue are available to parachains. AbridgedHostConfiguration only includes per-candidate limit for hrmp, this struct can only target (as of now) aggregated changes for ump queue.

pallets/parachain-system/src/lib.rs
```diff
@@ -544,6 +582,19 @@ pub mod pallet {
...
+	/// This value is optional since it requires extra relay-chain state proof.
+	#[pallet::storage]
+	pub(super) type ParaHead<T: Config> = StorageValue<_, relay_chain::HeadData, OptionQuery>;
+
+	#[pallet::storage]
+	pub(super) type UnincludedSegment<T: Config> = StorageValue<_, Vec<BlockTracker>, ValueQuery>;
+
+	#[pallet::storage]
+	pub(super) type AggregatedUnincludedSegment<T: Config> =
+		StorageValue<_, SegmentTracker, OptionQuery>;
```
slumber, line 596: Does the pallet require another runtime api exposed for appending into this queue?

slumber:
Also #1866 mentions
>UMP messages processed

Was it supposed to be dmp?

slumber:
And do we want it in master?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2654330

pallets/parachain-system/src/lib.rs
```diff
@@ -544,6 +582,19 @@ pub mod pallet {
 		Unauthorized,
 	}
 
+	/// Latest parachain head data included in the relay chain.
+	///
+	/// This value is optional since it requires extra relay-chain state proof.
+	#[pallet::storage]
+	pub(super) type ParaHead<T: Config> = StorageValue<_, relay_chain::HeadData, OptionQuery>;
```
rphmeier, line 589: This shouldn't need to hit storage - the latest included head should be read out of the state proof in the paras-inherent.

pallets/parachain-system/src/lib.rs
```diff
@@ -296,6 +299,34 @@ pub mod pallet {
...
+			let para_head = ParaHead::<T>::take();
+			weight += T::DbWeight::get().reads_writes(1, 1);
+
+			// Update unincluded segment related storage values.
+			if let Some(para_head) = para_head {
+				let dropped: Vec<BlockTracker> = <UnincludedSegment<T>>::mutate(|chain| {
+					let idx =
+						chain.iter().take_while(|block| block.para_head() != &para_head).count();
+					// Drop the block with an included para head too.
+					let idx = chain.len().min(idx + 1);
```
rphmeier, line 311: This seems inaccurate.

If we have an unincluded segment of `[A, B, C]` and then `B` is included, we'd get `idx = 1 + 1 = 2` and drain `[A, B]` from the front (correct so far).

But then if `B` is still the best included block at the next call here, we'd get `idx = 0 + 1 = 2` and drain `[C]`.

The fix is to only prune the unincluded segment when the best included parablock appears within it and do nothing otherwise.

pallets/parachain-system/src/lib.rs
rphmeier: Appending should only happen within the runtime, so we shouldn't need a runtime API for that.

pallets/parachain-system/src/unincluded_segment.rs
```diff
@@ -0,0 +1,117 @@
...
+				.expect("entry's been inserted earlier with `append`; qed");
+			entry.subtract(channel);
+		}
+
+		self.hrmp_outgoing.retain(|_, channel| !channel.is_empty());
+	}
+}
+
+#[derive(Encode, Decode, TypeInfo)]
+pub struct BlockTracker {
```
rphmeier, line 98: bikeshed: I don't like the name `BlockTracker` here. Maybe `UnincludedAncestor`

pallets/parachain-system/src/relay_state_snapshot.rs
```diff
@@ -235,6 +237,17 @@ impl RelayChainStateProof {
 			.map_err(Error::Config)
 	}
 
+	/// Read parachain [head data](`relay_chain::HeadData`) from the relay chain state proof.
+	///
+	/// Returns an error if anything failed at reading or decoding.
+	pub fn read_para_head(&self) -> Result<Option<relay_chain::HeadData>, Error> {
```
rphmeier, line 243: please rename this function to something more descriptive e.g. `read_included_para_head` and adjust docs

pallets/parachain-system/src/lib.rs
rphmeier: i.e. `on_finalize` should do the work of appending to the unincluded segment and updating the aggregate tracker.

pallets/parachain-system/src/unincluded_segment.rs
rphmeier: Note that the `HrmpChannel` structs which are contained in the relay-chain storage proof has current sizes and limits contained within them.

pallets/parachain-system/src/lib.rs
```diff
@@ -296,6 +299,34 @@ pub mod pallet {
 				weight += T::DbWeight::get().writes(1);
 			}
 
+			let para_head = ParaHead::<T>::take();
+			weight += T::DbWeight::get().reads_writes(1, 1);
+
+			// Update unincluded segment related storage values.
+			if let Some(para_head) = para_head {
```
rphmeier, line 306: `para_head` should not be allowed to be `None` once the parachain has been configured to support asynchronous backing. Otherwise, collators could simply choose to omit the `para_head` from the proof and break the logic.

We should have a 'switch' for parachain governance to enable the unincluded-segment functionality; after that point, it should be invalid for the relay-chain storage proof _not_ to contain the para head.

rphmeier: It should also be invalid for a parachain block to execute `on_initialize` when the unincluded segment is full and has not had anything drained.

pallets/parachain-system/src/lib.rs
slumber: I tried to account for the case of our fork being abandoned. Correct approach would include looking for the parent too, right?

pallets/parachain-system/src/lib.rs
slumber: We can introduce a config value for `MaxUnincludedLen`. If it's non-zero, then the switch is on

pallets/parachain-system/src/lib.rs
slumber: [ffe3dcb](https://github.com/paritytech/cumulus/pull/2438/commits/ffe3dcb5fc491350887323b1777341efc90e23f5)

pallets/parachain-system/src/unincluded_segment.rs
slumber: It's more descendant, not ancestor, isn't it?

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2656662

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2657029

pallets/parachain-system/src/lib.rs
rphmeier: I think the best thing is a storage item which can be set by governance. There are situations where you might want `MaxUnincludedLen` to be 0 but use the relay-chain state proof. Maybe an `Option<BlockNumber>` is best - it starts as `None`, but there should be no way to set it back to `None` once it's `Some`.

pallets/parachain-system/src/unincluded_segment.rs
rphmeier: From the perspective of the current block, they are ancestors. It makes more sense for me to look at them as ancestors rather than descendants - descendants are a tree, ancestors are a linked list.

pallets/parachain-system/src/lib.rs
rphmeier: So with that in mind, the correct approach is to clear everything up to and including the last included parablock, and to clear nothing if it is not present.

pallets/parachain-system/src/unincluded_segment.rs
rphmeier: i.e. I like to view the runtime state as the "perspective" of the block being executed. From that perspective they are ancestors. From the perspective of the included block, they are descendents, but that is more of a relay-chain perspective than a parachain runtime one.

pallets/parachain-system/src/lib.rs
rphmeier: i.e. @burdges suggested that we should have pluggable logic for the relay chain to set max-unincluded-len dynamically during on-initialize based on the relay-chain state, like how many cores are available, how far in the future, etc.

slumber: [cdf3ba9](https://github.com/paritytech/cumulus/pull/2438/commits/cdf3ba9f0a6b44e43c72406df0eaa063f3fb4521)

pallets/parachain-system/src/unincluded_segment.rs
slumber: [b1daf2d](https://github.com/paritytech/cumulus/pull/2438/commits/b1daf2d9a513f6b61833d1790328fe55afec7496)

(Although it's `Ancestor`, not `UnincludedAncestor`. Personally too much `Unincluded`-prefixed names worsen the readability)

pallets/parachain-system/src/lib.rs
slumber: [6549be6](https://github.com/paritytech/cumulus/pull/2438/commits/6549be6216b23d34ff3140d0f57d006c6512dfa0)

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2662582

pallets/parachain-system/src/lib.rs
slumber: I.e. this PR is missing the following in `on_finalize`:

If `MaxUnincludedLen` is present in the storage and `> 0`, then
construct `TotalBandwidthLimits` and try to append the block into unincluded segment.

?

slumber:
Do we want to account for `max_ancestry_len` here?

pallets/parachain-system/src/lib.rs
rphmeier: I'd expect some check in `on_initialize` which bails early if `hypothetical_segment_len_after_this_block > unincluded_segment_len` (the hypothetical is based on the pruning).

Then after `on_initialize` we'd always have the invariant that `unincluded_segment_len < max_unincluded_segment_len`, so we can unconditionally push the new bandwidth usage to the unincluded segment in `on_finalize`.

pallets/parachain-system/src/unincluded_segment.rs
rphmeier: I think that's fine, within the scope of the `unincluded_segment` module the context is clear.

rphmeier:
> Do we want to account for max_ancestry_len here?

This comment doesn't point to any specific code, not sure what you mean by this.

slumber:
> > Do we want to account for max_ancestry_len here?
> 
> This comment doesn't point to any specific code, not sure what you mean by this.

It's a general question, we have max unincluded length which is mapped to `max_candidate_depth`

But async backing parameters also have `allowed_ancestry_len`, which we're not accounting for in this PR

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2669366

pallets/parachain-system/src/lib.rs
rphmeier: I was thinking that we'd have the reduction in some kind of hook that runs before this check. Otherwise it does fail if the maximum gets reduced, true.

rphmeier: Yeah, let's defer updating logic to another PR, so this one is just about the core mechanism

rphmeier:
> But async backing parameters also have allowed_ancestry_len, which we're not accounting for in this PR

We could account for it in the runtime, but `allowed_ancestry_len` is a parameter of the relay chain block where the parachain block gets backed, whereas `candidate_depth` is a property of the relay chain block that is set as the relay-parent. We don't know ahead of time at one point a fresh parablock will get posted to the relay chain, so I'd imagined that the consensus/authoring layer that chooses relay parents would account for that instead.

pallets/parachain-system/src/lib.rs
```diff
@@ -282,6 +295,40 @@ pub mod pallet {
...
+					.collect();
+				let dmp_processed_count = ProcessedDownwardMessages::<T>::get();
+				let used_bandwidth = UsedBandwidth {
+					ump_msg_count,
+					ump_total_bytes,
+					hrmp_outgoing,
+					dmp_processed_count,
+				};
+				// The bandwidth constructed was ensured to satisfy relay chain constraints.
+				let ancestor = Ancestor::new_unchecked(used_bandwidth, todo!());
```
slumber, line 320: We need a head data of this block, but it's only available during the call to `fetch_collation_info` runtime api (because the value itself is produced by the consensus implementation on the node side).
If I attempt to store it, changes would be discarded anyway.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2676659

pallets/parachain-system/src/lib.rs
slumber: cc @bkchr

pallets/parachain-system/src/lib.rs
```diff
@@ -428,6 +547,7 @@ pub mod pallet {
 			<RelayStateProof<T>>::put(relay_chain_state);
 			<RelevantMessagingState<T>>::put(relevant_messaging_state.clone());
 			<HostConfiguration<T>>::put(host_config);
+			<PendingDownwardMessages<T>>::put(downward_messages.clone());
```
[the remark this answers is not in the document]
rphmeier: If it's cleared in `on_finalize` then it never hits storage in practice, only the in-memory overlay.

pallets/parachain-system/src/lib.rs
rphmeier: Yeah, it's impossible to get that value at this stage. What should be done is to leave a "pending" value which is then updated with `system::parent_hash()` during `on_initialize` of the next block.

pallets/parachain-system/src/lib.rs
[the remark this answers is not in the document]
slumber: It's inserted in `set_validation_data`, fetched in `on_finalize`, cleared in `on_initialize`

pallets/parachain-system/src/lib.rs
```diff
@@ -282,6 +295,40 @@ pub mod pallet {
...
+				let ancestor = Ancestor::new_unchecked(used_bandwidth, todo!());
+
+				// Check in `on_initialize` guarantees there's space for this block.
+				UnincludedSegment::<T>::append(ancestor);
+
+				let watermark = HrmpWatermark::<T>::get();
+				AggregatedUnincludedSegment::<T>::mutate(|agg| {
+					let agg = agg.get_or_insert_with(SegmentTracker::default);
+					agg.append(&ancestor, watermark, &limits)
+						.expect("unincluded segment limits exceeded");
```
rphmeier, line 329: Add a comment here to indicate that this bails because the block is invalid.

rphmeier: There are checks that user code that queues HRMP messages doesn't overrun this limit in a friendly way, right?

user code should never encounter this panic if possible.

cc @ggwpez

rphmeier: (ah, this is for segment length)

pallets/parachain-system/src/lib.rs
```diff
@@ -296,15 +343,84 @@ pub mod pallet {
...
+
+							let left_count = (idx..chain.len()).count() as u32;
+							let dropped = chain.drain(..idx).collect();
+							(dropped, left_count)
+						});
+					weight += T::DbWeight::get().reads_writes(1, 1);
+
+					// sanity-check there's place for the block at finalization phase.
+					//
+					// TODO: this potentially restricts parachains from decreasing `MaxUnincludedLen` value.
```
rphmeier, line 394: Yes, it does - parachains manually adjusting `MaxUnincludedLen` would potentially brick their chain. This is fine as long as there is a warning in the storage item doc comment.

We will need good ways for parachains to adjust this.

primitives/parachain-inherent/src/client_side.rs
```diff
@@ -104,6 +104,7 @@ async fn collect_relay_storage_proof(
 	relevant_keys.push(relay_well_known_keys::hrmp_egress_channel_index(para_id));
 	relevant_keys.push(relay_well_known_keys::upgrade_go_ahead_signal(para_id));
 	relevant_keys.push(relay_well_known_keys::upgrade_restriction_signal(para_id));
+	relevant_keys.push(relay_well_known_keys::para_head(para_id));
```
rphmeier, line 107: @bkchr do you think it's an issue if we merge this into master and proofs start carrying the `para_head` in the proof even if it's unused?

Alternatively, we could refactor this bit of code and create a new version of the `ParachainInherentData`. The old version would not have this member, and the new version would.

rphmeier (commented):
I just realized something - I don't think that we can actually access the relay-chain state proof within the `on_initialize` area of code. It only gets handled later, in the `set_inherent_data` inherent. The logic we currently have in `on_initialize` needs to go there, actually.

pallets/parachain-system/src/lib.rs
slumber: I basically referenced https://github.com/paritytech/cumulus/pull/2438#discussion_r1163461339, need to link to a github issue too before the merge

primitives/parachain-inherent/src/client_side.rs
slumber: There's feature branch for cumulus if we don't want this hustle

slumber:
> I just realized something - I don't think that we can actually access the relay-chain state proof within the `on_initialize` area of code. It only gets handled later, in the `set_inherent_data` inherent.

Data put into storage by `set_validation_data` (= create_inherent) is available in `on_finalize` and dropped in `on_initialize`, why isn't state proof accessible?

pallets/parachain-system/src/lib.rs
slumber: [8f246c9](https://github.com/paritytech/cumulus/pull/2438/commits/8f246c9439fb223a370706cad71c3d60ff87eb17)

pallets/parachain-system/src/lib.rs
slumber: I'm honestly not sure, the code performs the check w.r.t. to host config for current block, so I can imagine this bailing if the unincluded chain becomes long enough.

pallets/parachain-system/src/lib.rs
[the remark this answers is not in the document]
rphmeier: two things stand out to me:
  * only the length of the list is used, not the messages themselves
  * it could be cleared in `on_finalize` too, right?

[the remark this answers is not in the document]
rphmeier: it also doesn't seem well-defined to me what `pending_downward_messages` _is_ in this context.

i.e. is it all pending downward messages that were present in the relay-parent state? If so, then we need to update `process_inbound_downward_messages` to avoid processing them in duplicate.

or is it only the pending downward messages which haven't been processed? In that case, we shouldn't have to make any changes to the processing logic.

For efficiency's sake, I'd favor the latter - we shouldn't include the same downward messages into parachain blocks multiple times.

horizontal logic probably needs the same, no?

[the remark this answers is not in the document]
slumber: I followed the pattern of cleaning everything in `on_initialize`

As for the messages, yes, I suggested in the first comments to either put only len or 
> Second option is to introduce migration and save its len along with ProcessedDownwardMessages and rename it into something like DownwardQueueSize.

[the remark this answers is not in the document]
slumber: > i.e. is it all pending downward messages that were present in the relay-parent state? If so, then we need to update process_inbound_downward_messages to avoid processing them in duplicate.

These are all messages from the downward queue. They're not used for processing in the current code, so it's not an issue. We should probably just use len.
But the thing is, according to the code each candidate exhaust the dm queue:
```rust
let dm_count = downward_messages.len() as u32;
// ...
ProcessedDownwardMessages::<T>::put(dm_count);
```
Do we even need to track this limitation at all?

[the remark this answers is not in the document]
rphmeier: If we assume that `ParachainInherentData` only contains messages which have _not_ been posted to any prior parachain block (which is the way it should be to avoid redundancy), I don't know that that current bandwidth limits approach works.

The bandwidth limits for DMP/HRMP-in should be based on the _total_ size of the incoming queues at the relay-parent, whereas some (or most) of the messages may have already been processed.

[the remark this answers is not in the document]
rphmeier: > according to the code each candidate exhaust the dm queue

Yes, this looks currently guaranteed. We could avoid tracking this property as a result, and reintroduce the concept if we get to the point where we implement things more piecemeal.

pallets/parachain-system/src/lib.rs
slumber: @rphmeier 
```rust
/// A means of figuring out what outbound XCMP messages should be being sent.
pub trait XcmpMessageSource {
	/// Take a single XCMP message from the queue for the given `dest`, if one exists.
	fn take_outbound_messages(maximum_channels: usize) -> Vec<(ParaId, Vec<u8>)>;
}
```
I believe this trait not only should account for max messages taken from the storage, but also for segment limitations if we want to make this panic correct.
Probably primitives for segment should be extracted into common crate to be accessible from both pallets.

rphmeier: Thanks. Can you file an issue about that?

pallets/parachain-system/src/lib.rs
```diff
@@ -960,6 +1055,68 @@ impl<T: Config> Pallet<T> {
...
+
+					let left_count = (idx..chain.len()).count() as u32;
+					let dropped = chain.drain(..idx).collect();
+					(dropped, left_count)
+				});
+			weight_used += T::DbWeight::get().reads_writes(1, 1);
+
+			// sanity-check there's place for the block at finalization phase.
+			//
+			// TODO: this potentially restricts parachains from decreasing `MaxUnincludedLen` value.
```
rphmeier, line 1099: I think the answer here, now that we've moved this to run in the inherent, is "it doesn't".

`on_initialize` always runs before inherents, so any pallet setting the `MaxUnincludedLen` in code that executes in the `on_initialize` of any pallet would run before this, so it is safe.

slumber: What about scenario when there're 4 blocks in the unincluded segment, and the len is reduced to e.g. 1?

pallets/parachain-system/src/lib.rs
```diff
@@ -282,6 +295,38 @@ pub mod pallet {
 					.map(|(recipient, data)| OutboundHrmpMessage { recipient, data })
 					.collect::<Vec<_>>();
 
+			if MaxUnincludedLen::<T>::get().map_or(false, |max_len| !max_len.is_zero()) {
+				// NOTE: these limits don't account for the amount of processed messages from
+				// downward and horizontal queues.
+				//
+				// This is correct because:
+				// - inherent never contains messages that were previously processed.
+				// - current implementation always attempts to exhaust each message queue.
```
rphmeier, line 304: Thanks for this comment. I filed https://github.com/paritytech/cumulus/issues/2472 which this could link to.

pallets/parachain-system/src/lib.rs
```diff
@@ -364,6 +429,12 @@ pub mod pallet {
 				"ValidationData must be updated only once in a block",
 			);
 
+			// NOTE: the inherent data is expected to be unique, even if this block is built
+			// in the context of the same relay parent as the previous one. In particular,
+			// the inherent shouldn't contain messages that were already processed by any of the
+			// ancestors.
+			//
+			// This invariant should be upheld by the node-side.
```
rphmeier, line 437: A comment for the benefit of other reviewers: the "node-side" here could also include the `ProvideInherent` implementation. That is probably the best way to ensure this in practice. Just trimming some MQCs.

rphmeier (approved):
Looks good!

pallets/parachain-system/src/lib.rs
rphmeier: That is fine. For example, assume that there is some unconditional code that sets the maximum length to 1 in `on_initialize` at a certain block height. For more dynamically controlled systems, the reasoning is the same but slightly more complicated.

As long as the unincluded segment has length > 1, we can't author a new block. But new relay-parents will "drain" the unincluded segment from the back, so eventually the unincluded segment will be short enough to push a new block onto. This is the desired behavior of shrinking the maximum segment length.

pallets/parachain-system/src/lib.rs
ggwpez: >cc @ggwpez

I have not worked with the XCMP outbound side yet, so cant help you here.

pallets/parachain-system/src/lib.rs
rphmeier: It's rather that we are creating a block validity rule based on the relay-parent. It'll be impossible to create valid blocks on top of some `(para parent, relay parent)` pairs, and that failure will manifest in the inherent that handles the storage proof. Anything that runs in `on_initialize` could alter the block validity condition (e.g. altering the maximum unincluded segment length). We will probably also want hooks for people to write dynamic block validity rules based on the relay-parent in question.

rphmeier (approved):

PR #2447, opened 2023-04-12
[Feature] XCM-Emulator

ruseinov:
The original tool has been built by [Shaun](https://github.com/shaunxw) of Acala and lives in [here](https://github.com/shaunxw/xcm-simulator/tree/master/xcm-emulator).
Important: once this gets accepted - make sure to tip Shaun for his good work.
tip: 16PNhcMEvGbTYdCht9vLy6gj8qehRehEpb2Ef

**What is this for?**
This tool is designed for testing XCM configuration and message passing between pre-defined runtimes, e.g. Polkadot and Statemint. With up-to-date chain specs it allows for verifying whether specific XCM messages work in live networks.

**Why does it have to live in cumulus repo?**
Mainly because we want it to target the latest code and relative cumulus paths as opposed to a certain release. Otherwise it's not possible to use it for testing cumulus runtimes that target polkadot master. This copy already diverges from the original tool. Furthermore, by running test examples in CI we'll be able to verify that the tool is still up-to-date with the latest polkadot and cumulus changes. 
Initially this was introduced to [polkadot repo](https://github.com/paritytech/polkadot/pull/7037), but that requires ugly patch overrides for transitive dependencies on both cumulus and polkadot.

Future work on this PR is [here](https://github.com/paritytech/cumulus/pull/2502).


ruseinov:
bot fmt

gilescope (approved):

bkontur (approved):
I would maybe just move all `yayoi` stuff to `yayoi.rs` under `example/yayoi.rs`, because we wont probably use it anyway

ruseinov:
> I would maybe just move all `yayoi` stuff to `yayoi.rs` under `example/yayoi.rs`, because we wont probably use it anyway

I guess the idea of this being a separate crate has to do with the fact that it's similar to importing a runtime. As part of the next iteration we could completely remove `test_runtime` and just use one of the preconfigured runtimes for this.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2671144

joepetrowski:
Do not merge yet, I'd like to review.

muharem (commented):
In the PR we make xcm-emulator's polkadot/substrate/cumulus dependencies point to `master` branch. Why cant we do the same for original crate? Or may be CI jobs do it for us? We do probably something similar for releases. It could be done even with our help in the original repo.
I am in favor for a solution which will let us maintain and develop only one xcm-emulator.
I think everyone will benefit from it.

vstam1:
I would like to argue that we can completely deprecate the xcm-simulator from the Polkadot repo. Please correct me if I'm wrong, but I see the xcm-emulator as a superset of the xcm-simulator. We could create an example with a simple relay-para setup in the xcm-emulator that allows developers to quickly test ideas. The problem now is that the Kusama runtime is pretty restrictive in it's configuration, making it hard to test simple ideas in the emulator. Creating a testing relay, like the testing para runtime would allow us to create both an xcm playground (for testing ideas and instructions) and complex xcm tests with production level runtimes (testing configurations). 

Would like to know what you think.

ruseinov:
> I would like to argue that we can completely deprecate the xcm-simulator from the Polkadot repo. Please correct me if I'm wrong, but I see the xcm-emulator as a superset of the xcm-simulator. We could create an example with a simple relay-para setup in the xcm-emulator that allows developers to quickly test ideas. The problem now is that the Kusama runtime is pretty restrictive in it's configuration, making it hard to test simple ideas in the emulator. Creating a testing relay, like the testing para runtime would allow us to create both an xcm playground (for testing ideas and instructions) and complex xcm tests with production level runtimes (testing configurations).
> 
> Would like to know what you think.

It's not completely correct. XCM-Simulator is designed to work with mock runtimes, therefore it's more suitable for all sorts of examples that don't involve fully-fledged live runtimes. XCM-Emulator, on the other hand, is designed specifically to test said live runtimes or something very close to those. 

I would say this: 
1. Use XCM-Simulator for simple xcm examples that don't care about the transport level and complex runtime configurations.
2. Use XCM-Emulator to test the real deal aka live runtimes. There are still some caveats that have to do with the fact that we're testing things in an environment without block production, but it allows for testing quite a bit. 

The truth is - both are valuable. And I foresee that both types of test examples are going to be very useful for the community. I'm hopeful that upcoming XCM-Emulator integration-tests for assets parachains, properly documented, are going to help other parachain teams understand the configurations and flow better.

vstam1:
Yeah I see what you mean. The thing I currently do not like is that the tools are located in different repos and that they overlap quite a bit. I was looking at the configurations of the test_runtime in this pr and the parachain runtime in the xcm-simulator. The biggest difference is the transport layer. How I understand the combination of Xcm and transport layer, is that xcm execution should be independent from the transport layer. Quickly like to check with @KiChjang and @franciscoaguirre? 

So if we can match both runtimes, and create a test_relay, like we have the test_runtime, we could replace the simulator with this test_network in the emulator. I am not sure if it is possible, but I think we should at least check it. I am in favor of having one well documented tool for both an xcm playground and full integration tests then two separate tools. I also think that if we choose to keep both tools (simulator for simple xcm examples and emulator for testing live runtime configurations) that we should move them both to cumulus. What do you think @ruseinov?

Polkadot-Forum:

https://forum.polkadot.network/t/test-frameworks-and-utilities-for-polkadot-and-friends/2612/3

ruseinov:
> So if we can match both runtimes, and create a test_relay, like we have the test_runtime, we could replace the simulator with this test_network in the emulator. I am not sure if it is possible, but I think we should at least check it. I am in favor of having one well documented tool for both an xcm playground and full integration tests then two separate tools. I also think that if we choose to keep both tools (simulator for simple xcm examples and emulator for testing live runtime configurations) that we should move them both to cumulus. What do you think @ruseinov?

I think that consolidation of those two tools or potential deprecation of xcm-simulator is not in scope of this PR. These two are fundamentally different, even though they have a very similar API. Looking at this code and what it does makes me think that it might actually be easier to have two tools and document them separately. When it comes to moving both tools to cumulus - I'd rather have both in polkadot when the repos are consolidated (monorepo) or XCM could potentially be a shared module not belonging to either of those. For now having this in cumulus is a necessity dictated by the dependency issues that otherwise arise.

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+				RuntimeEvent::Ump(polkadot_runtime_parachains::ump::Event::ExecutedUpward(
+					_,
+					Outcome::Incomplete(_, XcmError::NoPermission)
+				))
+			)));
+		});
+	}
+
+	#[test]
+	fn xcmp() {
```
joepetrowski, line 267: Do these tests care about the transport? 
```suggestion
	fn para_to_para() {
```

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+	fn xcmp_through_a_parachain() {
+		use test_runtime::{PolkadotXcm, Runtime, RuntimeCall};
+
+		Network::reset();
+
+		// The message goes through: Pumpkin --> Mushroom --> Octopus
+		let remark = RuntimeCall::System(frame_system::Call::<Runtime>::remark_with_event {
+			remark: "Hello from Pumpkin!".as_bytes().to_vec(),
+		});
+		let send_xcm_to_octopus = RuntimeCall::PolkadotXcm(pallet_xcm::Call::<Runtime>::send {
```
joepetrowski, line 308: Where do these names come from? Why not continue with "Parachain A/B/C/..."?

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+			send_xcm_to_octopus.get_dispatch_info().weight,
+			Weight::from_parts(110000010, 10000010)
+		);
+		ParachainA::execute_with(|| {
+			assert_ok!(PolkadotXcm::send_xcm(
+				Here,
+				MultiLocation::new(1, X1(Parachain(2))),
+				Xcm(vec![Transact {
+					origin_kind: OriginKind::SovereignAccount,
+					require_weight_at_most: 110_000_010.into(),
```
joepetrowski, line 326: Are these hardcoded weights things that people expect to hold? People will copy these values as acceptable things to use in their XCM programs, so we should ensure that they come from a well-known weights table, and exceeding it is a breaking change.

This should also be a 2D weight.

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+
+		ParachainC::execute_with(|| {
+			use test_runtime::{RuntimeEvent, System};
+			// execution would fail, but good enough to check if the message is received
+			System::events().iter().for_each(|r| println!(">>> {:?}", r.event));
+
+			assert!(System::events().iter().any(|r| matches!(
+				r.event,
+				RuntimeEvent::XcmpQueue(cumulus_pallet_xcmp_queue::Event::Fail { .. })
+			)));
```
joepetrowski, line 350: Why do we expect execution to fail? "good enough to check if the message is received" doesn't seem very meaningful.

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+use frame_support::{pallet_prelude::Weight, traits::GenesisBuild};
+use sp_runtime::AccountId32;
+
+use xcm_emulator::{decl_test_network, decl_test_parachain, decl_test_relay_chain};
+
+decl_test_relay_chain! {
+	pub struct KusamaNet {
```
joepetrowski, line 23:
```suggestion
	pub struct KusamaRelay {
```

joepetrowski: Why is this even called "kusama"? It does not seem to be reading in any actual network state. In fact, it's setting the default XCM version to 3 when it is really 2 on Kusama. So this is clearly _not_ Kusama nor an accurate representation of it.

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+				kusama_runtime::RuntimeOrigin::root(),
+				Some(3)
+			));
+		});
+
+		kusama_send_rmrk("Kusama", 1);
+		parachain_receive_and_reset_events(true);
+	}
+
+	fn kusama_send_rmrk(msg: &str, count: u32) {
```
joepetrowski, line 394:
```suggestion
	fn send_remark_with_event_to_child_parachain_a(msg: &str, count: u32) {
```

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,438 @@
...
+						),
+						call: remark.encode().into(),
+					}]),
+				));
+			}
+		});
+	}
+
+	fn parachain_receive_and_reset_events(received: bool) {
+		ParachainA::execute_with(|| {
```
joepetrowski, line 419: Why is this function specific to `ParachainA`?
```suggestion
	fn parachain_a_verify_remark_event_and_reset_all_events(received: bool) {
		ParachainA::execute_with(|| {
```

xcm/xcm-emulator/Cargo.toml
```diff
@@ -0,0 +1,30 @@
+[package]
+name = "xcm-emulator"
+description = "Test kit to emulate cross-chain message passing and XCM execution"
```
joepetrowski, line 3:
```suggestion
description = "Test kit to emulate XCM program execution."
```

xcm/xcm-emulator/README.md
```diff
@@ -0,0 +1,5 @@
+# xcm-emulator
+
+xcm-emulator uses relay chain and parachain runtimes.
+Users could plug in Kusama, Statemine, Polkadot and other pre-configured runtimes.
+With up-to-date chain specs, it's able to verify if specific XCM messages work in live networks.
```
joepetrowski, line 5: Not OK for documentation.

xcm/xcm-emulator/test-runtime/Cargo.toml
```diff
@@ -0,0 +1,71 @@
+[package]
+name = "test-runtime"
+description = "A simple runtime for cross-chain messages tests."
```
joepetrowski, line 3:
```suggestion
description = "A simple runtime for cross-chain message tests."
```

joepetrowski (changes requested):

xcm/xcm-emulator/example/src/lib.rs
ruseinov: It's called Kusama, because it's using the Kusama runtime definition. Note that those are just the examples that came with xcm-emulator in the original repo, unchanged. Those are meant to be used as guidance for builders in terms of structure and code, nothing more. 

XCM-Emulator is not designed to be reading the chain state yet, it simply allows for using pre-defined runtime definitions to test message passing between them. I'm sure this can be further enhanced to do more advanced testing, however this PR's aim is to move XCM-Emulator into the Cumulus repo and making sure it compiles/runs with the current versions of Cumulus and Polkadot. 

Furthermore, we are setting Kusama's XCM version to `3` in [parachains-integration-tests](https://github.com/paritytech/parachains-integration-tests/tree/master/examples/statemine), those are the tests that are used for release checks, so that is consistent with what we are doing here.

xcm/xcm-emulator/example/src/lib.rs
ruseinov: It's a refactoring bug, messages definitely need to be fixed.

xcm/xcm-emulator/example/src/lib.rs
ruseinov: The XCM-Emulator is using XCMP/UMP sinks to pass messages around, so it's relevant. Another nice thing about having those names is that they correspond to those in `parachains-integration-tests` repo and potentially to the future parachain tests.

xcm/xcm-emulator/example/src/lib.rs
ruseinov: Only because `kusama_send_rmrk` is currently sending only to `parachain_a`.

ruseinov:
I've opened another one [here](https://github.com/paritytech/cumulus/pull/2502), which is where I'm going to fix the review comments.

ruseinov:
@joepetrowski I have addressed your comments in [this PR](https://github.com/paritytech/cumulus/pull/2502). However, I stuck to `dmp/xcmp` test names to be consistent with those in [parachains-integration-tests repo](https://github.com/paritytech/parachains-integration-tests). If this does not work for some reason - I'm happy to apply whatever gets suggested instead.

NachoPal:
> @joepetrowski I have addressed your comments in [this PR](https://github.com/paritytech/cumulus/pull/2502). However, I stuck to `dmp/xcmp` test names to be consistent with those in [parachains-integration-tests repo](https://github.com/paritytech/parachains-integration-tests). If this does not work for some reason - I'm happy to apply whatever gets suggested instead.

I've been thinking about it and I think that gathering test under `dmp`, `ump` and `hrmp` definitions will break user stories flow. I think it is better to create tests based on concepts and use cases, e.g _Teleport Asset_ - that would include both `dmp` and `ump` (from  Relay to Para and its way back).

ruseinov:
> > @joepetrowski I have addressed your comments in [this PR](https://github.com/paritytech/cumulus/pull/2502). However, I stuck to `dmp/xcmp` test names to be consistent with those in [parachains-integration-tests repo](https://github.com/paritytech/parachains-integration-tests). If this does not work for some reason - I'm happy to apply whatever gets suggested instead.
> 
> I've been thinking about it and I think that gathering test under `dmp`, `ump` and `hrmp` definitions will break user stories flow. I think it is better to create tests based on concepts and use cases, e.g _Teleport Asset_ - that would include both `dmp` and `ump` (from Relay to Para and its way back).

Fair point, perhaps we should regroup those tests then. Might also be a good idea to do that for integration tests too, this way it's less confusing and more consistent.

xcm/xcm-emulator/example/src/lib.rs
gilescope: These tests should work the same way when HRMP exists. I think Joe's right on this one - it's not that particular.

xcm/xcm-emulator/Cargo.toml
gilescope:
```suggestion
description = "Test kit to emulate XCM program execution"
```

tonyalaribe:
bot rebase

xcm/xcm-emulator/README.md
```diff
@@ -0,0 +1,9 @@
+# xcm-emulator
+
+XCM-Emulator is a tool to emulate XCM program execution using pre-configured runtimes, including those
+used to run on live networks, such as Kusama, Polkadot, Statemine et cetera.  
+This allows for testing cross-chain message passing, verifying outcomes, weights and side-effects.
```
gilescope, line 5:
```suggestion
XCM-Emulator is a tool to emulate XCM program execution using
pre-configured runtimes, including those used to run on live
networks, such as Kusama, Polkadot, Statemine et cetera.
This allows for testing cross-chain message passing, verifying
outcomes, weights and side-effects. It is faster than spinning up
a zombienet and as all the chains are in one process debugging using Clion is easy.

## Limitations

As the channels are mocked, using xcm-emulator tests to test
channel setup would not be appropriate.

## Alternatives

If you just wish to test execution of various XCM instructions
against the XCM VM then the `xcm-simulator` (in the polkadot
repo) is the perfect tool for this.
```

xcm/xcm-emulator/example/src/lib.rs
gilescope: Switched to 2d weight. A well-known weights table sounds like a good idea but I'd argue that should be for a separate follow on PR.

xcm/xcm-emulator/example/src/lib.rs
gilescope: This has been tightened up so it needs to be an origin fail.

xcm/xcm-emulator/README.md
gilescope: Extended the documentation to point out that xcm-simulator also exists and when to use one over the other.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2786888

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2787359

bkontur (approved):

xcm/xcm-emulator/test-runtime/Cargo.toml
```diff
@@ -0,0 +1,71 @@
+[package]
+name = "test-runtime"
```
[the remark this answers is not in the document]
gilescope: good idea

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2788453

muharem (approved):
@ruseinov @gilescope thank you for docs

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+			// 	r.event,
+			// 	RuntimeEvent::System(frame_system::Event::Remarked { sender: _, hash: _ })
+			// )));
+			assert!(System::events().iter().any(|r| matches!(
+				r.event,
+				RuntimeEvent::Ump(polkadot_runtime_parachains::ump::Event::ExecutedUpward(
+					_,
+					Outcome::Incomplete(_, XcmError::NoPermission)
+				))
+			)));
```
joepetrowski, line 269: This seems just to be testing `SafeCallFilter`? What is the value of "testing" a `remark_with_event`?

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+use frame_support::{pallet_prelude::Weight, traits::GenesisBuild};
+use sp_runtime::AccountId32;
+
+use xcm_emulator::{decl_test_network, decl_test_parachain, decl_test_relay_chain};
+
+// Setup relay chain.
```
joepetrowski, line 22:
```suggestion
// Set up relay chain.
```

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+decl_test_relay_chain! {
+	pub struct KusamaRelay {
+		Runtime = kusama_runtime::Runtime,
+		XcmConfig = kusama_runtime::xcm_config::XcmConfig,
+		new_ext = kusama_ext(),
+	}
+}
+
+// Setup ParachainA.
```
joepetrowski, line 31:
```suggestion
// Set up ParachainA.
```

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+	pub struct ParachainA {
+		Runtime = test_runtime::Runtime,
+		RuntimeOrigin = test_runtime::RuntimeOrigin,
+		XcmpMessageHandler = test_runtime::XcmpQueue,
+		DmpMessageHandler = test_runtime::DmpQueue,
+		new_ext = parachain_ext(1),
+	}
+}
+
+// Setup ParachainB.
```
joepetrowski, line 42:
```suggestion
// Set up ParachainB.
```

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+	pub struct ParachainB {
+		Runtime = test_runtime::Runtime,
+		RuntimeOrigin = test_runtime::RuntimeOrigin,
+		XcmpMessageHandler = test_runtime::XcmpQueue,
+		DmpMessageHandler = test_runtime::DmpQueue,
+		new_ext = parachain_ext(2),
+	}
+}
+
+// Setup ParachainC.
```
joepetrowski, line 53:
```suggestion
// Set up ParachainC.
```

xcm/xcm-emulator/example/src/lib.rs
```diff
@@ -0,0 +1,448 @@
...
+	pub struct ParachainC {
+		Runtime = test_runtime::Runtime,
+		RuntimeOrigin = test_runtime::RuntimeOrigin,
+		XcmpMessageHandler = test_runtime::XcmpQueue,
+		DmpMessageHandler = test_runtime::DmpQueue,
+		new_ext = parachain_ext(3),
+	}
+}
+
+// Setup a network with all the declared parachains and a relay chain.
```
joepetrowski, line 64:
```suggestion
// Set up a network with all the declared parachains and a relay chain.
```

xcm/xcm-emulator/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Alternatives
+
+If you just wish to test execution of various XCM instructions
+against the XCM VM then the `xcm-simulator` (in the polkadot
+repo) is the perfect tool for this.
+
+## How to use
+
+1. Setup a relay chain:
```
joepetrowski, line 23:
```suggestion
1. Set up a relay chain:
```

xcm/xcm-emulator/README.md
```diff
@@ -0,0 +1,110 @@
...
+		Runtime = test_runtime::Runtime,
+		RuntimeOrigin = test_runtime::RuntimeOrigin,
+		XcmpMessageHandler = test_runtime::XcmpQueue,
+		DmpMessageHandler = test_runtime::DmpQueue,
+		new_ext = parachain_ext(1),
+	}
+}
+```
+
+3. Setup a network with all the declared parachains and a relay chain:
```
joepetrowski, line 48:
```suggestion
3. Set up a network with all the declared parachains and a relay chain:
```

joepetrowski (commented):
Some small things remaining.

xcm/xcm-emulator/README.md
```diff
@@ -0,0 +1,110 @@
...
+pre-configured runtimes, including those used to run on live
+networks, such as Kusama, Polkadot, Statemine et cetera.
+This allows for testing cross-chain message passing, verifying
+outcomes, weights and side-effects. It is faster than spinning up
+a zombienet and as all the chains are in one process debugging using Clion is easy.
+
+## Limitations
+
+As the channels are mocked, using xcm-emulator tests to test
+channel setup would not be appropriate.
```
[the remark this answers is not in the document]
ruseinov: It’s using `*Sink` to facilitate message passing as opposed to `xcm-simulator` that is using mocks. Opening of channels is not required though.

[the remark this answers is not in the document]
bkontur: > Opening of channels is not required though.

what channels? e.g. hrmp channel needs to be opened, else parachain's XcmRouter->XcmpQueue sending messages wont work.

If "As the channels are mocked" mean ump/dmp/hrmp, I wouldnt say they are mocked. Just to read/write mesasges from these queues is orchestrated directly here with some thread_local queues without `relay-chain-interface / relay-chain-inprocess-interface` code, so lets say it uses testing/simplified transport layer between relaychain and parachain.

xcm/xcm-emulator/README.md
```diff
@@ -0,0 +1,110 @@
+# xcm-emulator
+
+XCM-Emulator is a tool to emulate XCM program execution using
+pre-configured runtimes, including those used to run on live
+networks, such as Kusama, Polkadot, Statemine et cetera.
+This allows for testing cross-chain message passing, verifying
+outcomes, weights and side-effects. It is faster than spinning up
```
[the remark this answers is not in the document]
gilescope: It does not test the 'physical' transport of the messages as that is substituted, but the logical transportation of the messages is tested - I.e. the routing as well as the execution of the messages in the context of the foreign chain is tested.

xcm/xcm-emulator/README.md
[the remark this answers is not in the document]
ruseinov: > Could rewrite as: "As the messages do not physically go through the same messaging infrastructure there is some code that is not being tested compared to using slower E2E tests. In future it may be possible to run these XCM emulated tests as E2E tests (without changes)."

Sounds perfect. It's indeed not 100% the same, but not fully mocked either.

xcm/xcm-emulator/example/src/lib.rs
ruseinov: Yes, it's currently hitting the `SafeCallFilter`, but once that's removed - this could be reinstated. Since the purpose of those example tests is to demonstrate common use-cases, not necessarily focusing on the outcome - it could be ok to keep it as is.

According to the discussion [here](https://github.com/paritytech/polkadot/pull/6824#issuecomment-1464913946), as far as I understand it, there's no real reason why the `remark_with_event` call is disabled in the first place, as the actual event that's stored contains a hash and not the original unbounded `Vec`, correct me if I'm wrong.

That said, there's an alternative, which is to `force_create_asset`. It's a bit of a more complicated scenario, but it works.

joepetrowski: Yes, I understand all that, but this error returns [quite early](https://github.com/paritytech/polkadot/blob/release-v0.9.42/xcm/xcm-executor/src/lib.rs#L550) in the `Transact` instruction. So the only thing this tests is that this `ensure` statement works. As the error indicates, the execution is "incomplete".

ruseinov: Right, I'll raise a PR to change that to `force_create_assets`, going to be less confusing. Though it would be nice to see `remark_with_event` whitelisted too, but not as part of this PR.

ruseinov: Fixed [here](https://github.com/paritytech/cumulus/pull/2566)

xcm/xcm-emulator/README.md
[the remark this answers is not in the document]
ruseinov: Fixed [here](https://github.com/paritytech/cumulus/pull/2447)

NachoPal:
I merged https://github.com/paritytech/cumulus/pull/2593 with the following improvements (most of them aimed to ease tests development and apply DRY principle):

- `decl_test_relay_chain` , `decl_test_parachain` and `decl_test_network` modified to accept multiple struct items avoiding the need of multiple declarations of the same macro for each item.
- Multi-networks declarations are possible now (previously `xcm-emulator` was only capable of working for one network)
    - This new capability potentially unlocks Bridged XCM communication (mocked transport layer not implemented yet)
- Redefined `decl_test_relay_chain` and `decl_test_parachain` struct argument:
    - New `genesis` field expecting `Storage`. This improves the previous way of declaring the initial storage state of the test externality.
    - New `on_init` field to execute arbitrary actions during struct item initialisation.
    - New `runtime` field  (this field gathers previous struct types definitions + new ones). Mandatory `xcm-emulator` types dependancies.
    - New `pallets_extra` argument. Place to add specific runtime pallets types to be accessible through the struct item implementation.
- Proper structure for `xcm-emulator` with `RelayChain` , `Parachain` , `Network` and `NetworkComponent` traits.
- Improvement of `TestExt` trait.
- Helper methods for relay an parachains items struct implementations:
    - `parent_location()`
    - `child_location_of()`
    - `account_id_of()`
    - `account_data_of()`
    - `sovereign_account_id_of()`
    - `para_id()`
    - `fund_accounts()`
    - `events()`
- `assert_expected_events` macro. It helps checking received events with pattern matching and conditional assertion of theirs inner fields. [Example](https://github.com/paritytech/cumulus/blob/342babdf31733202d8c57abc91cd19b69c008e5f/parachains/integration-tests/emulated/assets/statemint/src/tests/reserve_transfer.rs#L34-L38).
    - Added events logger to inspect received events.
- Handy helper functions `within_threshold` & `weight_within_threshold` for events assessment.

In the PR I also added the base for the System Parachains *emulated* test suite:

- `integration-tests-common` package
    - `[lib.rs](http://lib.rs)`
        - Relay Chains, System Parachains and Networks declarations (for Polkadot & Kusama networks)
        - Parameters declaration of expected senders and receivers accounts for each chain. These are meant to be across all test to improve readability and prevent developers from declaring each time their own accounts.
    - `[constants.rs](http://constants.rs)`
        - System Parachain’s genesis definitions plus other common and Parachain’s specific constants.
- A few Statemint tests have been added as a reference of how to write tests to keep tests structure consistent across the whole System Parachains test suite.

`xcm-emulator` example has been removed. Now that `xcm-emulator` lives in Cumulus, System Parachains tests work as great examples.

Previous e2e test have been moved under `parachains/integration-tests/e2e`

_**Note**_: I will add proper documentation about how to use `xcm-emulator` once the PR has been approved and I make sure no interface changes are needed.

gilescope:
On the whole it looks good - I think it's only as we develop tests against it that we will see where we need to tweak it. Can't wait for it to be merged.

parachains/integration-tests/emulated/common/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+			LocationToAccountId: statemint_runtime::xcm_config::LocationToAccountId,
+			System: statemint_runtime::System,
+			Balances: statemint_runtime::Balances,
+			ParachainSystem: statemint_runtime::ParachainSystem,
+			ParachainInfo: statemint_runtime::ParachainInfo,
+		},
+		pallets_extra = {
+			PolkadotXcm: statemint_runtime::PolkadotXcm,
+			Assets: statemint_runtime::Assets,
+		}
```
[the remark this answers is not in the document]
NachoPal: > New `runtime` field (this field gathers previous struct types definitions + new ones). Mandatory xcm-emulator types dependancies.
New `pallets_extra` argument. Place to add specific runtime pallets types to be accessible through the struct item implementation.

`runtime` is a fixed and expected number of types that `xcm-emulator` lib needs as a dependancy in its implementation.
`pallets_extra` is optional an you can add as many types you want. If you do not add them `xcm-emulator` implementation will still work.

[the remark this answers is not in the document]
gilescope: So the types listed in `runtime` are all mandatory for the emulator to work and are the minimum pallets that we can get away with, and then everything else is put in `pallets_extra`?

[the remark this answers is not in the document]
NachoPal: Besides the conceptual difference. The need of having both separately is coming also from declarative macros arguments handling limitations.

joepetrowski (approved):

joepetrowski:
bot merge

NachoPal:
@bkontur @ggwpez @mordamax Could any of you please drop a medium tip here?

mordamax:
@NachoPal sure,
@ruseinov is that the address you use? Polkadot address: 127o5MyqLCjZzggX5kTUokpvspKU4SgC1eRioJMYckVF4Ndm
We can try kusama thorough opengov if you want

bkchr:
I think @shaunxw already got some tip for it back in the days? If not, he can just request tip on chain.

xlc:
He got it for simulator, not emulator

PR #2449, opened 2023-04-13
Align BridgeHub runtimes with other SP runtimes + reused test for teleport native tokens + some nits

bkontur:
Align BridgeHub runtimes with other SP runtimes
Reused teleports_for_native_asset_works test to all bridge-hub runtime
Some small nits


parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
```diff
@@ -96,17 +96,16 @@ pub type XcmOriginToTransactDispatchOrigin = (
...
+	// `RuntimeOrigin::Signed` origin of the same 32-byte value.
 	SignedAccountId32AsNative<RelayNetwork, RuntimeOrigin>,
 	// Xcm origins can be represented natively under the Xcm pallet's Xcm origin.
 	XcmPassthrough<RuntimeOrigin>,
 );
 
 match_types! {
-	// TODO: map gov2 origins here - after merge https://github.com/paritytech/cumulus/pull/1895
-	pub type ParentOrParentsExecutivePlurality: impl Contains<MultiLocation> = {
+	pub type ParentOrParentsPlurality: impl Contains<MultiLocation> = {
```
serban300, line 106: Nit: `ParentOrParentsPlurality` and `ParentOrSiblings` seem to be duplicated for all the bridge hubs, and maybe for other runtimes as well. Could we define them once in a common place instead and reuse them ?

parachains/runtimes/bridge-hubs/test-utils/Cargo.toml
```diff
@@ -0,0 +1,50 @@
+[package]
+name = "bridge-hub-test-utils"
+version = "0.1.0"
+authors = ["Parity Technologies <admin@parity.io>"]
+edition = "2021"
+description = "Utils for BridgeHub testing"
+
+[dependencies]
```
serban300, line 8: Nit: Do we need all the other dependencies apart from `asset-test-utils` ?

parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
bkontur: @joepetrowski @gilescope @muharem 
good point, `ParentOrParentsPlurality / ParentOrSiblings` looks the same everywhere now and probably wont change ever (maybe we will add another ones to the tuple like `FellowsPlurality`),

wdyt about moving `ParentOrParentsPlurality / ParentOrSiblings` to the `parachains-common`?

muharem: In this case it should in some specialized module/file which will be under the same review/merge rules.
I am personally fine with such duplications.

parachains/runtimes/bridge-hubs/test-utils/Cargo.toml
bkontur: ok, removed, they will come later [here](https://github.com/paritytech/cumulus/pull/2111/files#diff-b70b5564f1d2014dc102ea781326a768361b8ccb9faacfc3517fc1acb3e99814)

serban300 (approved):

parachains/runtimes/bridge-hubs/bridge-hub-polkadot/src/xcm_config.rs
```diff
@@ -197,7 +199,7 @@ impl xcm_executor::Config for XcmConfig {
 		MaxInstructions,
 	>;
 	type Trader =
-		UsingComponents<WeightToFee, DotRelayLocation, AccountId, Balances, ToAuthor<Runtime>>;
+		UsingComponents<WeightToFee, DotRelayLocation, AccountId, Balances, ToStakingPot<Runtime>>;
```
bkontur, line 202: @joepetrowski 
I am not sure here, why others (assets, collectives) have `ToStakingPot` and here we have `ToAuthor`,

can you, please, confirm this change?
`ToAuthor`->`ToStakingPot`  and upper `CurrencyAdapter<Balances, ()>` -> `CurrencyAdapter<Balances, DealWithFees<Runtime>>` (DealWithFees uses ToStakingPot)

parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
joepetrowski: Yeah, same

bkontur:
bot rebase

parachains/runtimes/bridge-hubs/bridge-hub-polkadot/src/xcm_config.rs
joepetrowski: I agree this should go to the staking pot since it is not up to the individual author to process incoming messages.

parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/xcm_config.rs
```diff
@@ -163,10 +163,10 @@ pub type Barrier = DenyThenTry<
 		AllowKnownQueryResponses<PolkadotXcm>,
 		WithComputedOrigin<
 			(
-				// Allow anything to pay for execution.
+				// If the message is one that immediately attemps to pay for execution, then allow it.
 				AllowTopLevelPaidExecutionFrom<Everything>,
-				// Parent and its exec plurality get free execution.
-				AllowExplicitUnpaidExecutionFrom<ParentOrParentsExecutivePlurality>,
+				// Parent and its plurality (i.e. governance bodies) gets free execution.
```
joepetrowski, line 168:
```suggestion
				// Parent and its pluralities (i.e. governance bodies) get free execution.
```

joepetrowski (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2700837

parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
```diff
@@ -165,10 +164,10 @@ pub type Barrier = DenyThenTry<
 		AllowKnownQueryResponses<PolkadotXcm>,
 		WithComputedOrigin<
 			(
-				// Allow anything to pay for execution.
+				// If the message is one that immediately attemps to pay for execution, then allow it.
 				AllowTopLevelPaidExecutionFrom<Everything>,
-				// Parent and its exec plurality get free execution.
-				AllowExplicitUnpaidExecutionFrom<ParentOrParentsExecutivePlurality>,
+				// Parent and its pluralities (i.e. governance bodies) get free execution.
+				AllowExplicitUnpaidExecutionFrom<ParentOrParentsPlurality>,
```
bkchr, line 170: Why don't we have `Fellowship` here as well as on the Polkadot hub?

bkchr (approved):

parachains/runtimes/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
bkontur: I did it according to statemine, but I think it is related to the fact that Kusama does not have Collectives parachain, polkadot's `FellowsPlurality` points to the Collectives parachain [here](https://github.com/paritytech/cumulus/pull/2449/files#diff-167b6926f30b1edd29856107ea200c33c40ddea60c85128cfe175e6c3749860bR114-R115)

bkontur:
bot rebase

bkontur:
bot merge

PR #2461, opened 2023-04-18
Optimize level monitor reconstruction

davxy:
Partially address https://github.com/paritytech/substrate/issues/13864

---

The reconstruction of the level monitor used by cumulus was using the `tree_route` function.

While this function perfs are acceptable for single very long branches (tested with a branch with an order of 3M with time ~2s on my "humble" machine). The limits hits when there are several leaves (tested with ~2K leaves).
Using this function to reconstruct the monitor computes all the 2K routes. Even if these are just small branches composed of 1 single block.

---

This PR adopts the strategy of stopping the backtrack as soon as we hit something already imported (i.e. we hit an already imported branch).

---

With this modification we should manage all the scenarios that realistically we find in the wild.

(obviously the worst case is when every branch starts from the last finalized block and each has some gigantic length... but this is quite improbable if not manually crafted for some reason).

### DIFF
--- a/client/consensus/common/src/level_monitor.rs
+++ b/client/consensus/common/src/level_monitor.rs
@@ -48,17 +48,17 @@ pub enum LevelLimit {
 
 /// Support structure to constrain the number of leaves at each level.
 pub struct LevelMonitor<Block: BlockT, BE> {
-	// Max number of leaves for each level.
+	/// Max number of leaves for each level.
 	level_limit: usize,
-	// Monotonic counter used to keep track of block freshness.
+	/// Monotonic counter used to keep track of block freshness.
 	pub(crate) import_counter: NumberFor<Block>,
-	// Map between blocks hashes and freshness.
+	/// Map between blocks hashes and freshness.
 	pub(crate) freshness: HashMap<Block::Hash, NumberFor<Block>>,
-	// Blockchain levels cache.
+	/// Blockchain levels cache.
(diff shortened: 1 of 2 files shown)


arkpar (approved):

skunert (approved):
The logic looks good to me!

I noticed that the previous implementation would not add branches to the level monitor that are rooted below the finalized block. The new implementation will add all blocks above the finalized block, so not all blocks in the level monitor are guaranteed descendants of the finalized block. I think it's hard to avoid this but I wanted to leave this observation here in case it's unwanted.

Minor nit: We should have the struct field comments as doc comments: https://github.com/paritytech/cumulus/blob/083f76ce51faeb67a091cdc9a792f2f571978c0d/client/consensus/common/src/level_monitor.rs#L51-L62

davxy:
> I noticed that the previous implementation would not add branches to the level monitor that are rooted below the finalized block. The new implementation will add all blocks above the finalized block, so not all blocks in the level monitor are guaranteed descendants of the finalized block. I think it's hard to avoid this but I wanted to leave this observation here in case it's unwanted.

If ever happens, dead branches blocks will be removed as the finalized blocks advances (in the `block_imported` method).
We were keeping information for finalized branches also before (i.e. the cleanup in `block_imported` is performed per level and not to the top).  Only the initialization was indeed not considering the *stale* branches (as you've noted).

Because (IIRC) the blocks will be held by the backend even if are stale (now is probably even more correct)

davxy:
bot merge

PR #2462, opened 2023-04-18
Parachain node should not recover blocks while syncing

skunert:
While syncing a parachain full node recently, I noticed that we are trying to recover blocks while the parachain node is still syncing and way behind the chain tip.
In this PR, I propose not to recover anything while sync is in progress. In cases where the relay chain node is already synced but the parachain node is not, we already receive import notifications from the relay chain and try to recover the contained candidates. They can not be imported anyway since we are missing the parents while syncing.

### DIFF
--- a/client/pov-recovery/src/lib.rs
+++ b/client/pov-recovery/src/lib.rs
@@ -47,7 +47,8 @@
 
 use sc_client_api::{BlockBackend, BlockchainEvents, UsageProvider};
 use sc_consensus::import_queue::{ImportQueueService, IncomingBlock};
-use sp_consensus::{BlockOrigin, BlockStatus};
+use sc_network_sync::SyncingService;
+use sp_consensus::{BlockOrigin, BlockStatus, SyncOracle};
 use sp_runtime::traits::{Block as BlockT, Header as HeaderT, NumberFor};
 
 use polkadot_node_primitives::{AvailableData, POV_BOMB_LIMIT};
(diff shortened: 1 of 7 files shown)


client/pov-recovery/src/lib.rs
```diff
@@ -597,16 +606,27 @@ where
...
+	sync_service: Arc<SyncingService<Block>>,
 ) -> RelayChainResult<impl Stream<Item = (CommittedCandidateReceipt, SessionIndex)>> {
 	let import_notification_stream = relay_chain_client.import_notification_stream().await?;
 
 	let filtered_stream = import_notification_stream.filter_map(move |n| {
 		let client_for_closure = relay_chain_client.clone();
+		let sync_oracle = sync_service.clone();
 		async move {
 			let hash = n.hash();
+			if sync_oracle.is_major_syncing() {
```
altonen, line 621: Is it going to download the candidate later if it now skips them since it's syncing or what is going to happen with this pending candidate?

altonen (approved):
Code-wise LGTM

davxy (approved):

bkchr (approved):

client/pov-recovery/src/lib.rs
skunert: We are maintaining a list of pending candidates that we have seen from the relay chain but that have not yet been finalized. So the list of candidates we maintain here is limited anyway. During sync, we skip them now, but once we reach the tip, the list will soon be complete again.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2701277

PR #2470, opened 2023-04-20
[Polkadot Companion] for 7101

mrcnski:
PR: https://github.com/paritytech/polkadot/pull/7101


s0me0ne-unkn0wn (approved):

Cargo.lock
mrcnski, line 1: Should companions include changes to `Cargo.lock`? 🤔

slumber: I think you can always re-generate it once polkadot pr lands

slumber (approved):

PR #2475, opened 2023-04-21
Allow xcm transactions to statemint from computed origins on relay chain

tonyalaribe:
Related to https://github.com/paritytech/substrate/pull/13607 

To support treasury spends over XCM originating from the relay chain, we update the xcm barrier to allow unpaid execution from these computed origins.

### DIFF
--- a/parachains/runtimes/assets/statemine/src/xcm_config.rs
+++ b/parachains/runtimes/assets/statemine/src/xcm_config.rs
@@ -38,11 +38,12 @@ use sp_runtime::traits::ConvertInto;
 use xcm::latest::prelude::*;
 use xcm_builder::{
 	AccountId32Aliases, AllowExplicitUnpaidExecutionFrom, AllowKnownQueryResponses,
-	AllowSubscriptionsFrom, AllowTopLevelPaidExecutionFrom, CurrencyAdapter, EnsureXcmOrigin,
-	FungiblesAdapter, IsConcrete, LocalMint, NativeAsset, NoChecking, ParentAsSuperuser,
-	ParentIsPreset, RelayChainAsNative, SiblingParachainAsNative, SiblingParachainConvertsVia,
-	SignedAccountId32AsNative, SignedToAccountId32, SovereignSignedViaLocation, TakeWeightCredit,
-	UsingComponents, WeightInfoBounds, WithComputedOrigin,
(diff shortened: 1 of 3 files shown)


parachains/runtimes/assets/statemine/src/xcm_config.rs
```diff
@@ -360,6 +361,8 @@ pub type Barrier = DenyThenTry<
 				AllowExplicitUnpaidExecutionFrom<ParentOrParentsPlurality>,
 				// Subscriptions for version tracking are OK.
 				AllowSubscriptionsFrom<ParentOrSiblings>,
+				// Allow unpaid execution when from a computed origin or plurarity on the relay chain.
+				AllowUnpaidExecutionFrom<ParentOrParentsPlurality>,
```
bkontur, line 365: @tonyalaribe 
what xcm transactions/scenarios do you have? Isnt it possible to extend them with `UnpaidExecution` instruction?

tonyalaribe: My final XCM includes the `UnpaidExecution` instruction
```rust
Xcm(
      [
            SetAppendix(Xcm([ReportError(QueryResponseInfo { destination: MultiLocation { parents: 1, interior: Here }, query_id: 9, max_weight: <wasm:stripped> })])), 
            UnpaidExecution { weight_limit: Unlimited, check_origin: None }, DescendOrigin(X1(AccountId32 { network: None, id: [] })), 
           TransferAsset { 
                    assets: MultiAssets([MultiAsset { id: Concrete(MultiLocation { parents: 1, interior: Here }), fun: Fungible(5) }]), 
                    beneficiary: MultiLocation { parents: 0, interior: X1(AccountId32 { network: None, id: [] })}
             }
      ])

```

But that existing Barrier `AllowExplicitUnpaidExecutionFrom<ParentOrParentsPlurality>,` doesn't match it because the barrier expects the UnpaidExecution to be the first instruction which is not possible since SetAppendix always takes the first position in my scenario

tonyalaribe: image


But I imagine a likely better solution could be for me to modify AllowExplicitUnpaidExecutionFrom to support the `UnpaidExecution` instructions being in other position (not position 1). I'll explore this and see if I can get it to work

PR #2497, opened 2023-04-29
Skip av-store in collator overseer

skunert:
Followup for https://github.com/paritytech/polkadot/pull/7131

Also mark the consensus task as blocking, since it is importing and finalizing which are performing blocking operations.

### DIFF
--- a/client/service/src/lib.rs
+++ b/client/service/src/lib.rs
@@ -125,7 +125,7 @@ where
 
 	task_manager
 		.spawn_essential_handle()
-		.spawn("cumulus-consensus", None, consensus);
+		.spawn_blocking("cumulus-consensus", None, consensus);
 
 	let pov_recovery = PoVRecovery::new(
 		recovery_handle,
@@ -218,7 +218,7 @@ where
 
 	task_manager
 		.spawn_essential_handle()
-		.spawn("cumulus-consensus", None, consensus);
+		.spawn_blocking("cumulus-consensus", None, consensus);
 
 	let pov_recovery = PoVRecovery::new(
 		recovery_handle,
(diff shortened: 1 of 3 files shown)


client/service/src/lib.rs
```diff
@@ -125,7 +125,7 @@ where
 
 	task_manager
 		.spawn_essential_handle()
-		.spawn("cumulus-consensus", None, consensus);
+		.spawn_blocking("cumulus-consensus", None, consensus);
```
davxy, line 128: Why is this is now classified as blocking? 

[here](https://github.com/paritytech/substrate/blob/74b2c92066ec3abcb612faa9272f246ae339fab3/client/service/src/config.rs#L153-L158) I can read that `TaskType::Blocking` is generally used when the task may block somehow the execution for pecause of expensive CPU ops or `sleep`s.

What are the synchronous ops that may "block" the executor here?

skunert: This task receives finality notifications from the relay chain and finalizes accordingly.
Therefore it calls `finalize_block` on the backend [here](https://github.com/paritytech/cumulus/blob/master/client/consensus/common/src/parachain_consensus.rs#L71), which does potentially expensive tree-route calculations and db access. I prefer to let the TaskManager spawn a blocking task for this.

davxy (approved):

client/service/src/lib.rs
davxy: Got it! Indeed traversing the chain can be quite expensive. Maybe a day we can make this `tree_route` async by spawning a dedicated task and waiting for the result on a channel (I already saw this pattern around the project IIRC)

michalkucharczyk (approved):

Issue #2476, opened 2023-04-21
Aura alterations for Asynchronous Backing compatibility


rphmeier:
The bulk of the changes here should also apply to SASSAFRAS.

## High Level

(feel free to skip ahead to the proposed modifications section)

As a recap (more details #2301), we need to prepare Cumulus collators for more advanced parachains protocol features such as Asynchronous Backing, [Elastic Scaling](https://forum.polkadot.network/t/parachain-scaling-by-parablock-splitting/341), and [Blockspace Regions](https://forum.polkadot.network/t/unifying-bulk-and-mid-term-blockspace-with-strided-regions/2228). Both of these changes impact the collator logic and protocols in similar ways: they enable more blockspace to be utilized, and they enable this blockspace to be consumed optimistically, based on predictions of what will happen on the relay chain.

Collator-selection algorithms are not responsible for consensus safety (i.e. security) but just consensus liveness (not halting). Liveness is a spectrum, though. From a formal standpoint, consensus algorithms usually focus on guaranteeing some minimal level of liveness, whereas in practice we care about maximizing throughput, minimizing latency, and reducing waste.

To be more specific, we can clarify two desirable properties for a collator-selection algorithm:
  1. **Maximize utilization of available blockspace** (Goal L1)
  2. **Minimize effort in fulfiling L1** (Goal L2)

The phrase "available blockspace" here means something very specific. It refers to the number of cores and the proportion of each of those cores allocated to the parachain. This deliberately ignores the Pay-as-you-Go model, as Aura and similar collator-selection mechanisms are focused on dividing up authorship rights within already allocated blockspace regions, as opposed to Pay-as-you-Go, which is focused on purchasing and consuming blockspace in small increments.

The available blockspace can be roughly condensed into two variables: the maximum capacity *C* and the expected velocity *V*. Capacity refers to the amount of blocks that can be pending at any given time. e.g. for 1 core this might be 1 + 4 (the 4 comes from some lookahead performed by Polkadot's validators). The velocity of blockspace refers to the rate of block processing by the relay chain. Roughly speaking, with one core, V = 1, with two cores V = 2 and so on. Velocity calculations are more subtle when we take into account parachains sharing cores with strided blockspace regions.

[...]

PR #2501, opened 2023-05-02
refactor unincluded segment length into a ConsensusHook

rphmeier:
This introduces the notion of a `ConsensusHook` for managing the unincluded segment capacity and performing consensus-specific checks. We'll provide some basic hooks, including an `ExpectParentIncluded` hook which is used to smooth over any transition periods where the collators aren't yet posting the included block's hash.

The idea is that we can write a custom `ConsensusHook` for the logic described in #2476 (probably fixed velocity=1, capacity=2/4, enforces max velocity + 1 blocks per slot).

I also fixed a bug in UMP queue size calculation and fixed some tests accordingly. Also relaxed the watermark rules to make them make sense, related to https://github.com/paritytech/polkadot/pull/7188

The vision with the consensus hook is something like this: it will be able to handle the relay-chain state proof and extract information like the number of cores currently assigned or upcoming on-demand claims and adjust the velocity and capacity of the unincluded segment. It will allow us to do stuff like dynamically switch between on-demand and Aura/SASSAFRAS. and so on.

### DIFF
--- a/client/collator/src/lib.rs
+++ b/client/collator/src/lib.rs
@@ -385,9 +385,11 @@ mod tests {
 		TestClientBuilder, TestClientBuilderExt,
 	};
 	use cumulus_test_runtime::{Block, Header};
+	use cumulus_test_relay_sproof_builder::RelayStateSproofBuilder;
 	use futures::{channel::mpsc, executor::block_on, StreamExt};
 	use polkadot_node_subsystem_test_helpers::ForwardSubsystem;
 	use polkadot_overseer::{dummy::dummy_overseer_builder, HeadSupportsParachains};
+	use polkadot_primitives::HeadData;
 	use sp_consensus::BlockOrigin;
 	use sp_core::{testing::TaskExecutor, Pair};
 	use sp_runtime::traits::BlakeTwo256;
@@ -415,10 +417,14 @@ mod tests {
 			_: PHash,
 			validation_data: &PersistedValidationData,
 		) -> Option<ParachainCandidate<Block>> {
+			let mut sproof = RelayStateSproofBuilder::default();
+			sproof.included_para_head = Some(HeadData(parent.encode()));
+			sproof.para_id = cumulus_test_runtime::PARACHAIN_ID.into();
+
 			let builder = self.client.init_block_builder_at(
 				parent.hash(),
 				Some(validation_data.clone()),
-				Default::default(),
+				sproof,
 			);
 
 			let (block, _, proof) = builder.build().expect("Creates block").into_inner();
--- a/client/collator/src/lib.rs
+++ b/client/collator/src/lib.rs
@@ -384,8 +384,8 @@ mod tests {
 		Client, ClientBlockImportExt, DefaultTestClientBuilderExt, InitBlockBuilder,
 		TestClientBuilder, TestClientBuilderExt,
 	};
-	use cumulus_test_runtime::{Block, Header};
 	use cumulus_test_relay_sproof_builder::RelayStateSproofBuilder;
+	use cumulus_test_runtime::{Block, Header};
 	use futures::{channel::mpsc, executor::block_on, StreamExt};
 	use polkadot_node_subsystem_test_helpers::ForwardSubsystem;
 	use polkadot_overseer::{dummy::dummy_overseer_builder, HeadSupportsParachains};
--- a/client/consensus/common/src/tests.rs
+++ b/client/consensus/common/src/tests.rs
@@ -581,9 +581,7 @@ fn do_not_set_best_block_to_older_block() {
 
 	let blocks = (0..NUM_BLOCKS)
 		.into_iter()
-		.map(|i| {
-			build_and_import_block(client.clone(), true)
-		})
+		.map(|_| build_and_import_block(client.clone(), true))
 		.collect::<Vec<_>>();
 
 	assert_eq!(NUM_BLOCKS as u32, client.usage_info().chain.best_number);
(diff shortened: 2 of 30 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2758988

pallets/parachain-system/src/consensus_hook.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl UnincludedSegmentCapacity {
+	pub(crate) fn get(&self) -> u32 {
+		match self.0 {
+			UnincludedSegmentCapacityInner::ExpectParentIncluded => 1,
+			UnincludedSegmentCapacityInner::Value(v) => v.get(),
+		}
+	}
+
+	pub(crate) fn is_expecting_included_parent(&self) -> bool {
```
slumber, line 35: Capacity expecting included parent seems confusing, or perhaps tangential?

Why isn't it `is_set` or something?

pallets/parachain-system/src/lib.rs
```diff
@@ -237,14 +247,27 @@ pub mod pallet {
...
+				&relevant_messaging_state,
+				host_config.max_upward_queue_count,
+				host_config.max_upward_queue_size,
+			);
+			let mut bandwidth_out = None;
+			if let Some(segment) = AggregatedUnincludedSegment::<T>::get() {
+				let mut b = total_bandwidth_out.clone();
+				b.subtract(segment.used_bandwidth());
+				bandwidth_out = Some(b);
+			}
```
slumber, line 261:
```suggestion
			let bandwidth_out = AggregatedUnincludedSegment::<T>::get().map(|segment| {
				let mut b = total_bandwidth_out.clone();
				b.subtract(segment.used_bandwidth());
				bandwidth_out = Some(b);
			});
```

pallets/parachain-system/src/consensus_hook.rs
rphmeier: it's more of a "is demanding" than "is expecting", true.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2766695

pallets/parachain-system/src/tests.rs
```diff
@@ -457,17 +466,19 @@ fn unincluded_segment_works() {
 			|| {},
 			|| {
 				let segment = <UnincludedSegment<Test>>::get();
-				// Block 123 was popped from the segment, the len is still 2.
+				// Block 123 was popped from the segment, the len is still 1.
```
slumber, line 469: Why? The test passes and it's 2.

pallets/parachain-system/src/tests.rs
```diff
@@ -621,15 +632,18 @@ fn send_upward_message_num_per_candidate() {
...
+	// 	2,
+	// 	|| {
+	// 		assert_eq!(UnincludedSegment::<Test>::get().len(), 0);
+	// 		/* do nothing within block */
+	// 	},
+	// 	|| {
+	// 		let v = UpwardMessages::<Test>::get();
+	// 		assert_eq!(v, vec![b"message 2".to_vec()]);
+	// 	},
+	// );
```
slumber, line 646: Does it require debugging?

pallets/parachain-system/src/tests.rs
rphmeier: I changed the comment while debugging, but 2 is correct.

just to explain why it's correct for posterity, an inclusion delay of 1 means:
  * parent is not included (unincluded segment is 1 in pre-state)
  * block we just built isn't included (unincluded segment is 2 in post-state)

client/collator/src/lib.rs
```diff
@@ -415,10 +417,14 @@ mod tests {
 			_: PHash,
 			validation_data: &PersistedValidationData,
 		) -> Option<ParachainCandidate<Block>> {
+			let mut sproof = RelayStateSproofBuilder::default();
+			sproof.included_para_head = Some(HeadData(parent.encode()));
+			sproof.para_id = cumulus_test_runtime::PARACHAIN_ID.into();
+
```
slumber, line 423: I believe there's more suitable place for this code
https://github.com/paritytech/cumulus/blob/b67b0d6bbc310b8e84d9981bf228b35fc1e8552d/test/client/src/block_builder.rs#L67-L73

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2791468

client/collator/src/lib.rs
rphmeier: That loses generality, e.g. setting the included para head to _not_ be in there.

client/consensus/common/src/tests.rs
```diff
@@ -548,7 +581,10 @@ fn do_not_set_best_block_to_older_block() {
 
 	let blocks = (0..NUM_BLOCKS)
 		.into_iter()
-		.map(|_| build_and_import_block(client.clone(), true))
+		.map(|i| {
+			println!("{}", i);
```
slumber, line 585:
```suggestion

```

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2830823

BradleyOlson64 (commented):
This proved to be a great primer on the Cumulus side of Async Backing 👍

PR #2520, opened 2023-05-04
Bump transaction_version

EgorPopelyaev:
This PR bumps `transaction_version` based on the [Extrinsic ordering check](https://github.com/paritytech/cumulus/issues/2507#issuecomment-1534403237) in #2507

### DIFF
--- a/parachains/runtimes/bridge-hubs/bridge-hub-polkadot/src/lib.rs
+++ b/parachains/runtimes/bridge-hubs/bridge-hub-polkadot/src/lib.rs
@@ -131,7 +131,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 9420,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version:2,
+	transaction_version: 2,
 	state_version: 1,
 };
 


bkontur (approved):

gilescope (approved):

muharem (approved):

jsidorenko (approved):

joepetrowski (approved):

chevdor:
In Polkadot, `transfer` has been removed and replaced by the new `transfer_allow_death`. Also a stub has been added for `transfer`.

Here it looks like this stub is missing.

```
[≠] pallet 10: Balances -> 25 change(s)
  - calls changes:
    [≠]  0: transfer_allow_death ( dest: AccountIdLookupOf<T>, value: T::Balance, )  )
        [Name(StringChange("transfer_allow_death", "transfer"))]
    [≠]  1: set_balance_deprecated ( who: AccountIdLookupOf<T>, new_free: T::Balance, old_reserved: T::Balance, )  )
        [Name(StringChange("set_balance_deprecated", "set_balance")), Signature(SignatureChange { args: [Changed(2, [Name(StringChange("old_reserved", "new_reserved"))])] })]
    [-] "upgrade_accounts"
    [-] "transfer"
    [-] "force_set_balance"
```

bkchr:
>  0: transfer_allow_death ( dest: AccountIdLookupOf<T>, value: T::Balance, )  )
        [Name(StringChange("transfer_allow_death", "transfer"))]

It mentions the rename her? The index stays the same and the signature.

chevdor:
> [Name(StringChange("transfer_allow_death", "transfer"))]

That does not make sense. The diff was done the wrong way around.

bkchr:
I don't know if the output makes sense or not. I just know that the old function, with a new name is there.

chevdor:
> I don't know if the output makes sense or not. I just know that the old function, with a new name is there.

yep, that's fine:
- `transfer` with index 0 "disappears" and gets replaced/renamed as `transfer_allow_death`
- the new `transfer` stub shows up with index 7

@EgorPopelyaev is re-running the diffs for the bridges tio be sure but that looks alright.

chevdor:
None of the tx_version bumps are required expect for collectives due to:
```
[≠] pallet 50: Alliance -> 1 change(s)
  - calls changes:
    [-] "close_old_weight"

[≠] pallet 51: AllianceMotion -> 2 change(s)
  - calls changes:
    [-] "close_old_weight"
```
details at: https://github.com/paritytech/cumulus/issues/2507#issuecomment-1534412351

EgorPopelyaev:
I'll leave the changes only for collectives and revert the rest.

muharem:
> * `transfer` with index 0 "disappears" and gets replaced/renamed as `transfer_allow_death`

Does not this require a transaction version bump?
As I remember `transfer` and `transfer_allow_death` is the same function, just renamed. But as I client I might refer it by name, not by index.

bkchr:
> As I remember `transfer` and `transfer_allow_death` is the same function, just renamed. But as I client I might refer it by name, not by index.

That is right. However, `transaction_version` doesn't protect against this. It protects against a user signing a tx with a call at runtime X that gets applied at runtime Y and the call gets decoded to something different. It just ensures that people don't execute calls they don't wanted to execute.

PR #2522, opened 2023-05-04
Fix Clippy

ggwpez:
Importing Polkadot Clippy config. Please add the CI job here @alvicsam


pallets/collator-selection/src/weights.rs
```diff
@@ -39,93 +39,91 @@ pub trait WeightInfo {
...
+		Weight::from_parts(71_461_000_u64, 0)
+			.saturating_add(T::DbWeight::get().reads(3_u64))
+			.saturating_add(T::DbWeight::get().writes(4_u64))
 	}
 	fn new_session(r: u32, c: u32) -> Weight {
-		Weight::from_parts(0 as u64, 0)
+		Weight::from_parts(0_u64, 0)
 			// Standard Error: 1_010_000
-			.saturating_add(Weight::from_parts(109_961_000 as u64, 0).saturating_mul(r as u64))
+			.saturating_add(Weight::from_parts(109_961_000_u64, 0).saturating_mul(r as u64))
```
ggwpez, line 75: Clippy does not seem to like the template.

bkchr (approved):

.cargo/config.toml
```diff
@@ -0,0 +1,32 @@
...
+  "-Aclippy::unit_arg",                  # styalistic.
+  "-Aclippy::option-map-unit-fn",        # styalistic
+  "-Aclippy::bind_instead_of_map",       # styalistic
+  "-Aclippy::erasing_op",                # E.g. 0 * DOLLARS
+  "-Aclippy::eq_op",                     # In tests we test equality.
+  "-Aclippy::while_immutable_condition", # false positives
+  "-Aclippy::needless_option_as_deref",  # false positives
+  "-Aclippy::derivable_impls",           # false positives
+  "-Aclippy::stable_sort_primitive",     # prefer stable sort
+]
```
chevdor, line 32:
```suggestion
]

```

chevdor (approved):

parachains/runtimes/assets/statemint/src/xcm_config.rs
```diff
@@ -166,85 +166,82 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 			}
 		}
 
-		match call {
+		matches!(
```
chevdor, line 169: Are you sure about that one?
Using the `matches!` macro we lose the completeness check.
[Color](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6673dbf7f2227c8b1e6059051b0e371b)

chevdor (changes requested):
Taking back my ✅
It is risky swapping `match` for the `matches!` macro where you did it.

chevdor (approved):

bkchr:
> Taking back my ✅
> It is risky swapping `match` for the `matches!` macro where you did it.

Why? The macro compiles down to the exact same code..

chevdor:
> > Taking back my ✅
> > It is risky swapping `match` for the `matches!` macro where you did it.
> 
> Why? The macro compiles down to the exact same code..

Yes but you will miss some errors if new variants have been introduced.

In that case, `match` will complain (unless a catch all `_` was used) and `matches!` will happily compile.

bkchr:
> Yes but you will miss some errors if new variants have been introduced.
> 
> In that case, `match` will complain (unless a catch all `_` was used) and `matches!` will happily compile.

It will just return `false` for new variants that are not explicitly matched on. So, it is secure and nothing breaks or allows something accidentally.

skunert (approved):

alvicsam (approved):

bkchr:
bot rebase

ggwpez:
bot rebase

Issue #2544, opened 2023-05-08
Ensure only the first V + 1 blocks from a given author at a slot are accepted by the runtime


Sophia-Gold:
From #2476: 

> What I'm imagining is an extension to the logic of `pallet-aura-ext` which does these things in `on_initialize`:
>   * Track the slot of the current in a storage item, as well as how many blocks have been authored at that slot. `(Slot, u32)`
>   * if the slot is the same as the previous value of the storage item, the counter is incremented.
>   * if the slot is less than the previous value, then panic
>   * if the slot is greater than the previous value, then update the storage item to `(new_slot, 1)`.
>   * To handle the case where the storage item was not present directly after an upgrade, set the value to `(current_slot, 1)`.
> 
> Along with a `ConsensusHook` implementation like this (living in the `pallet-aura-ext` crate):
> 
> ```rust
> struct FixedVelocityConsensusHook<const N: u32>;
> ```
> 
> * If `counter > N + 1`, panic.

rphmeier:
We need a second constant `C` which is the maximum capacity, actually. so `FixedVelocityConsensusHook<V, C>` where both velocity & capacity need to be >= 1

slumber:
#2551

PR #2551, opened 2023-05-09
aura-ext: limit the number of authored blocks per slot

slumber:
Resolves https://github.com/paritytech/cumulus/issues/2544

### DIFF
--- a/pallets/aura-ext/src/consensus_hook.rs
+++ b/pallets/aura-ext/src/consensus_hook.rs
@@ -20,15 +20,20 @@
 //! The velocity `V` refers to the rate of block processing by the relay chain.
 
 use super::pallet;
-use frame_support::pallet_prelude::*;
-use sp_std::marker::PhantomData;
+use cumulus_pallet_parachain_system::{
+	consensus_hook::{ConsensusHook, UnincludedSegmentCapacity},
+	relay_state_snapshot::RelayChainStateProof,
+};
+use sp_std::{marker::PhantomData, num::NonZeroU32};
 
-/// A consensus hook for a fixed block processing velocity.
+/// A consensus hook for a fixed block processing velocity and unincluded segment capacity.
 pub struct FixedVelocityConsensusHook<T, const V: u32, const C: u32>(PhantomData<T>);
 
-impl<T: pallet::Config, const V: u32, const C: u32> FixedVelocityConsensusHook<T, V, C> {
-	/// Validates the number of authored blocks within the slot with respect to the `V + 1` limit.
-	pub fn validate_slot() -> Weight {
+impl<T: pallet::Config, const V: u32, const C: u32> ConsensusHook
+	for FixedVelocityConsensusHook<T, V, C>
+{
+	// Validates the number of authored blocks within the slot with respect to the `V + 1` limit.
+	fn on_state_proof(_state_proof: &RelayChainStateProof) -> UnincludedSegmentCapacity {
 		// Ensure velocity is non-zero.
 		let velocity = V.max(1);
 
@@ -39,6 +44,8 @@ impl<T: pallet::Config, const V: u32, const C: u32> FixedVelocityConsensusHook<T
 			panic!("authored blocks limit is reached for the slot")
 		}
 
-		T::DbWeight::get().reads(1)
+		NonZeroU32::new(sp_std::cmp::max(C, 1))
+			.expect("1 is the minimum value and non-zero; qed")
+			.into()
 	}
 }
--- a/pallets/aura-ext/src/consensus_hook.rs
+++ b/pallets/aura-ext/src/consensus_hook.rs
@@ -1,0 +1,44 @@
+// Copyright 2023 Parity Technologies (UK) Ltd.
+// This file is part of Cumulus.
+
+// Cumulus is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Cumulus is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! The definition of a [`FixedVelocityConsensusHook`] for consensus logic to manage
+//! block velocity.
+//!
+//! The velocity `V` refers to the rate of block processing by the relay chain.
+
+use super::pallet;
+use frame_support::pallet_prelude::*;
+use sp_std::marker::PhantomData;
+
+/// A consensus hook for a fixed block processing velocity.
+pub struct FixedVelocityConsensusHook<T, const V: u32, const C: u32>(PhantomData<T>);
+
+impl<T: pallet::Config, const V: u32, const C: u32> FixedVelocityConsensusHook<T, V, C> {
+	/// Validates the number of authored blocks within the slot with respect to the `V + 1` limit.
+	pub fn validate_slot() -> Weight {
+		// Ensure velocity is non-zero.
+		let velocity = V.max(1);
+
+		let authored = pallet::Pallet::<T>::slot_info()
+			.map(|(_slot, authored)| authored)
+			.expect("slot info is inserted on block initialization");
+		if authored > velocity + 1 {
+			panic!("authored blocks limit is reached for the slot")
+		}
+
+		T::DbWeight::get().reads(1)
+	}
+}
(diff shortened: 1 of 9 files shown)


pallets/aura-ext/src/lib.rs
```diff
@@ -68,7 +77,32 @@ pub mod pallet {
...
 
-			T::DbWeight::get().reads_writes(2, 1)
+			let new_slot = Aura::<T>::current_slot();
+			let velocity = T::ConsensusHook::block_processing_velocity();
+
+			let (new_slot, authored) = match SlotInfo::<T>::get() {
+				Some((slot, authored)) if slot == new_slot => {
+					// New slot is the same as previous one, increment the
+					// the counter if within the limit.
+					if authored < velocity.get() {
```
slumber, line 87: To match the issue description this should actually be
```rust
if authored <= velocity.get()
```

But it seems counter-intuitive to me. If there's a single core assigned to para, why would we allow `V + 1` blocks? This doesn't align with current behavior either.

@rphmeier

pallets/aura-ext/src/consensus_hook.rs
```diff
@@ -0,0 +1,63 @@
...
+	fn from(value: NonZeroU32) -> Self {
+		Self(value)
+	}
+}
+
+/// The consensus hook for managing number of blocks authored per slot.
+pub trait ConsensusHook {
+	/// This hook is called on each block initialization to determine if it's valid to author
+	/// another block.
+	fn block_processing_velocity() -> BlockProcessingVelocity;
```
slumber, line 45: Is there any argument we could expect? Looks like the only option is some constant value, and if para takes another core they'd have to upgrade their runtime

pallets/aura-ext/src/lib.rs
```diff
@@ -52,7 +55,13 @@ pub mod pallet {
 
 	/// The configuration trait.
 	#[pallet::config]
-	pub trait Config: pallet_aura::Config + frame_system::Config {}
+	pub trait Config: pallet_aura::Config + frame_system::Config {
+		/// Hook used to determine current velocity of block processing
+		/// by the relay chain.
+		///
+		/// Defines the limit of blocks authored per slot to `V`.
```
slumber, line 62: ditto here

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2809336

rphmeier (commented):
I don't understand why this exposes a new trait `ConsensusHook`. 

Instead, we should be exporting one or more `struct`s from this pallet that can be plugged in as `pallet_parachain_system::ConsensusHook`s.

The only thing people can reasonably do right now is have a constant velocity which we clamp to '1', although in the future we will want to do stuff like read out the state of blockspace regions, claims queue, etc.

pallets/aura-ext/src/lib.rs
```diff
@@ -68,7 +77,32 @@ pub mod pallet {
 			// Fetch the authorities once to get them into the storage proof of the PoV.
 			Authorities::<T>::get();
 
-			T::DbWeight::get().reads_writes(2, 1)
+			let new_slot = Aura::<T>::current_slot();
+			let velocity = T::ConsensusHook::block_processing_velocity();
+
+			let (new_slot, authored) = match SlotInfo::<T>::get() {
```
rphmeier, line 83: The only things `on_initialized` should be doing here is:
  1. ensure that the slot doesn't move backwards
  2. update the storage item.

The actual checks need to be done in the `pallet_parachain_system::ConsensusHook` implementation.

pallets/aura-ext/src/lib.rs
rphmeier: Note that if this is implemented as a `pallet_parachain_system::ConsensusHook` with a fixed capacity, the rules we get allow authors to author N blocks, where N is the lesser of:
  * the amount of space remaining in the unincluded segment
  * V+1

There is more elaboration on the capacity/velocity duality in https://github.com/paritytech/cumulus/issues/2476 . But in effect, this is about building a backlog.

This _does_ align with current behavior as long as the unincluded segment capacity is 1. But this is a general piece of code for taking full advantage of asynchronous backing in Aura.

pallets/aura-ext/src/lib.rs
rphmeier: Should be v+1, see #2476

rphmeier:
I think we had a bit of an architectural mental model mismatch.

I do see where you're coming from, with introducing a separate `ConsensusHook` type for Aura that determines the velocity. That _could_ be combined with a `pallet_parachain_system::FixedCapacityConsensusHook` to get effectively the same behavior. At the moment, the determination of velocity and capacity are not very tightly linked.

But looking forward, both velocity and capacity will both need to be determined dynamically, e.g. by reading relay-parent state and seeing how many cores a parachain is allocated onto or similar things. That's why I chose the architecture I did: consensus engines should have one piece of code for determining capacity and velocity. `parachain-system` only needs to care about `Capacity`, not velocity. The consensus hook itself is responsible for defining velocity rules.

pallets/aura-ext/src/consensus_hook.rs
rphmeier: Yeah. It'll have to be constant for now. This'll be a relatively easy amendment once we have completed the design of elastic scaling.

pallets/parachain-system/src/consensus_hook.rs
```diff
@@ -100,3 +102,15 @@ impl<const N: u32> ConsensusHook for FixedCapacityUnincludedSegment<N> {
 ///
 /// This is a simple type alias around a fixed-capacity unincluded segment with a size of 1.
 pub type RequireParentIncluded = FixedCapacityUnincludedSegment<1>;
+
+impl<T: pallet_aura_ext::Config, const V: u32, const C: u32> ConsensusHook
+	for aura_consensus_hook::FixedVelocityConsensusHook<T, V, C>
+{
+	fn on_state_proof(_state_proof: &RelayChainStateProof) -> UnincludedSegmentCapacity {
+		Self::validate_slot();
```
slumber, line 110: This uses weight so we should probably modify consensus hook to return it

pallets/parachain-system/Cargo.toml
```diff
@@ -34,6 +34,7 @@ xcm = { git = "https://github.com/paritytech/polkadot", default-features = false
 cumulus-pallet-parachain-system-proc-macro = { path = "proc-macro", default-features = false }
 cumulus-primitives-core = { path = "../../primitives/core", default-features = false }
 cumulus-primitives-parachain-inherent = { path = "../../primitives/parachain-inherent", default-features = false }
+cumulus-pallet-aura-ext = { path = "../aura-ext", default-features = false }
```
slumber, line 37: Should it be feature-gated?

pallets/parachain-system/src/consensus_hook.rs
rphmeier: This implementation should be in the `pallet-aura-ext` crate. The dependency is backwards.

pallets/aura-ext/src/consensus_hook.rs
```diff
@@ -0,0 +1,44 @@
...
+use super::pallet;
+use frame_support::pallet_prelude::*;
+use sp_std::marker::PhantomData;
+
+/// A consensus hook for a fixed block processing velocity.
+pub struct FixedVelocityConsensusHook<T, const V: u32, const C: u32>(PhantomData<T>);
+
+impl<T: pallet::Config, const V: u32, const C: u32> FixedVelocityConsensusHook<T, V, C> {
+	/// Validates the number of authored blocks within the slot with respect to the `V + 1` limit.
+	pub fn validate_slot() -> Weight {
```
rphmeier, line 31: This implementation doesn't need to be public.

pallets/parachain-system/Cargo.toml
rphmeier: Dependency is backwards. Aura-ext should depend on parachain-system and implement `ConsensusHook` in its own code.

This is a general principle in Rust development: expose traits in one crate and implement them in others.

pallets/aura-ext/src/consensus_hook.rs
```diff
@@ -0,0 +1,44 @@
...
+//! The definition of a [`FixedVelocityConsensusHook`] for consensus logic to manage
+//! block velocity.
+//!
+//! The velocity `V` refers to the rate of block processing by the relay chain.
+
+use super::pallet;
+use frame_support::pallet_prelude::*;
+use sp_std::marker::PhantomData;
+
+/// A consensus hook for a fixed block processing velocity.
```
rphmeier, line 26: docs: mention capacity, please.

pallets/aura-ext/src/consensus_hook.rs
rphmeier: This function should just be the body of 

`fn on_state_proof`

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2817212

pallets/parachain-system/src/lib.rs
```diff
@@ -58,7 +58,7 @@ use sp_std::{cmp, collections::btree_map::BTreeMap, prelude::*};
 use xcm::latest::XcmHash;
 
 mod migration;
-mod relay_state_snapshot;
+pub mod relay_state_snapshot;
```
rphmeier, line 61: Either:
  * re-export the necessary types for this module and leave it private
  * add module docs to `relay_state_snapshot` and move this declaration alongside the other public modules.

rphmeier (approved):
Looks good modulo nit. Thanks.

PR #2552, opened 2023-05-10
Relay-parent digest logs for parachains

rphmeier:
**After runtimes update to incorporate this PR, parachains using `pallet-parachain-system` will include a new digest in every header indicating their relay-parent's state root and number. Later on, this digest is intended to change to the relay-parent hash.**

Partially addresses #303 in the absence of https://github.com/paritytech/polkadot/issues/7191 .

This introduces a (currently unused) `CumulusDigestItem` which will have the relay-parent hash once that's available. In the meantime, there is a `#[doc(hidden)]` digest item representing the relay-parent storage root. This serves as a good enough unique identifier to implement https://github.com/paritytech/cumulus/issues/2476 .

For consideration - adding this private digest is not ideal, but it does let us implement some much better logic around asynchronous backing for parachains without needing to go through an execution API upgrade.

### DIFF
--- a/primitives/core/src/lib.rs
+++ b/primitives/core/src/lib.rs
@@ -18,6 +18,11 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
+use crate::substrate::{
+	generic::{Digest, DigestItem},
+	traits::Block as BlockT,
+	ConsensusEngineId,
+};
 use codec::{Decode, Encode};
 use polkadot_parachain::primitives::HeadData;
 use scale_info::TypeInfo;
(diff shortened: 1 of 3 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2810082

primitives/core/src/lib.rs
```diff
@@ -198,6 +208,83 @@ impl<B: BlockT> ParachainBlockData<B> {
 	}
 }
 
+/// A consensus engine ID indicating that this is a Cumulus Parachain.
+pub const CUMULUS_CONSENSUS_ID: ConsensusEngineId = *b"CMLS";
+
+/// Consensus header digests for Cumulus parachains.
+#[derive(Clone, RuntimeDebug, Decode, Encode, PartialEq)]
+pub enum CumulusDigestItem {
+	/// A digest item indicating that the parachain block has the provided relay-parent.
```
bkchr, line 217:
```suggestion
	/// A digest item indicating the relay-parent a parachain block was build on.
```

bkchr (approved):

primitives/core/src/lib.rs
```diff
@@ -41,6 +46,11 @@ pub mod relay_chain {
 	pub use polkadot_primitives::*;
 }
 
+/// A module that re-exports relevant Substrate primitive types.
+pub mod substrate {
```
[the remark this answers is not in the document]
rphmeier: It's generally best practice to re-export types that are dependent on in the crate so they don't need more dependencies.

We don't do this as much as we should, and note that this mirrors the `pub mod relay_chain` in this same module.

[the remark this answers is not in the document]
rphmeier: I can also re-export `Digest`/`DigestItem`/`ConsensusEngineID` directly from the `cumulus-primitives-core` crate if that's preferable. We need to re-export before this is merged in one way or another. This way is cleaner IMO.

primitives/core/src/lib.rs
rphmeier:
```suggestion
	/// A digest item indicating the relay-parent a parachain block was built against
```

primitives/core/src/lib.rs
```diff
@@ -198,6 +208,80 @@ impl<B: BlockT> ParachainBlockData<B> {
...
+	})
+}
+
+/// Utilities for handling the relay-parent storage root as a digest item.
+///
+/// This is not intended to be part of the public API, as it is a workaround for
+/// <https://github.com/paritytech/cumulus/issues/303> via
+/// <https://github.com/paritytech/polkadot/issues/7191>.
+///
+/// Runtimes using the parachain-system pallet are expected produce this digest item,
```
rphmeier, line 250:
```suggestion
/// Runtimes using the parachain-system pallet are expected to produce this digest item,
```

primitives/core/src/lib.rs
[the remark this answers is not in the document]
bkchr: > I can also re-export `Digest`/`DigestItem`/`ConsensusEngineID` directly from the `cumulus-primitives-core` crate if that's preferable.

Sounds good 👍

Generally you don't export every type of the public API. Most often when you have some trait or similar.

primitives/core/src/lib.rs
```diff
@@ -198,6 +208,80 @@ impl<B: BlockT> ParachainBlockData<B> {
...
+#[doc(hidden)]
+pub mod rpsr_digest {
+	use super::{relay_chain, ConsensusEngineId, Decode, Digest, DigestItem, Encode};
+
+	/// A consensus engine ID for relay-parent storage root digests.
+	pub const RPSR_CONSENSUS_ID: ConsensusEngineId = *b"RPSR";
+
+	/// Construct a digest item for relay-parent storage roots.
+	pub fn relay_parent_storage_root_item(
+		relay_parent_storage_root: relay_chain::Hash,
```
bkchr, line 269: If we would also add the relay parent number, it could be easier find find relevant relay block.

rphmeier: That is true. Do you have a use-case in mind?

bkchr: I thought you wanted to use this as intermediate solution until we can securely return the relay hash. So, it would be easier right now to find the hash.

eskimor (approved):

primitives/core/src/lib.rs
rphmeier: that's not needed for what I'm doing, as we have a very small number of relay chain blocks we can do a linear search against. However, I will add this.

rphmeier:
bot merge

PR #2566, opened 2023-05-11
[Fix] Re-work UMP tests to execute a safe call

ruseinov:


pepoviola:
Hi @ruseinov, the zombienet test are failing because is using an old version. I can create a pr in the base branch or you can set the version to `v1.3.52`.

NachoPal:
@ruseinov We decided that `examples` does not make sense anymore since `xcm-emulator` will live in Cumulus and System Parachain tests are good examples.

Closing this in favor of https://github.com/paritytech/cumulus/pull/2576

PR #2576, opened 2023-05-15
Nacho/improve xcm emulator v2

NachoPal:


bkontur:
@NachoPal 
I am not sure about name of the directory:
`parachains/integration-tests/xcm-emulator`
maybe something like:
`parachains/integration-tests/emulated`
or other better name :)


```
parachains/integration-tests/e2e
parachains/integration-tests/emulated
```

Issue #2472, opened 2023-04-20
runtime: Incoming Message-Queue Chains (MQCs) should be synchronized before importing messages


rphmeier:
DMP and XCMP both use [Message Queue Chains](https://paritytech.github.io/polkadot/book/glossary.html?highlight=MQC#glossary) to allow parachains to get a compact representation of the message queue state.

Parachain runtimes have access to limited state of their relay-parent, which they use to get the MQC head. They then need to walk the linked list backwards until a familiar link is encountered. Once the segments of the MQC are joined, a limited number of incoming messages can be processed.

It will actually be the responsibility of the parachain block author to provide the MQC links and messages that will be processed, and the runtime will just set rules on what the parachain block authors need to provide.

## Example

If the parachain's most recently known MQC head is 'a' and the new MQC head (read out of the state of the relay-parent) is 'e', then the block author needs to provide the missing links `[d c b]` and then some prefix of the list of corresponding messages `[B C D E]`.

rphmeier:
Ensuring that the inherent doesn't provide duplicate messages:
  * The `ParachainInherent` generation on the node-side should gather _all_ pending messages at the relay-parent from DMP and HRMP channels
  * The `ProvideInherent` implementation for `pallet-parachain-system` should cut off all messages which have already been processed, i.e. only provide the postfix of the available messages which weren't processed in the unincluded segment.

Note that with the current behavior, where all messages are processed, this postfix will always be empty.

ggwpez:
I am not sure if this makes it easier for you here, but in https://github.com/paritytech/cumulus/pull/2157 all the incoming DMP/XCMP messages are either immediately handled or enqueued into the MQ pallet and can thereby be treated as "processed".

rphmeier:
That is mostly orthogonal, although it does increase the amount of messages that the chain _can_ 'process' in one block.

PR #2590, opened 2023-05-17
parachain-system: drop processed messages from inherent data

slumber:
Partially addresses https://github.com/paritytech/cumulus/issues/2472

### DIFF
--- a/pallets/parachain-system/src/tests.rs
+++ b/pallets/parachain-system/src/tests.rs
@@ -241,7 +241,6 @@ struct BlockTests {
 	ran: bool,
 	relay_sproof_builder_hook:
 		Option<Box<dyn Fn(&BlockTests, RelayChainBlockNumber, &mut RelayStateSproofBuilder)>>,
-	persisted_validation_data_hook: Option<Box<dyn Fn(&BlockTests, &mut PersistedValidationData)>>,
 	inherent_data_hook:
 		Option<Box<dyn Fn(&BlockTests, RelayChainBlockNumber, &mut ParachainInherentData)>>,
 	inclusion_delay: Option<usize>,
@@ -302,14 +301,6 @@ impl BlockTests {
 		self
 	}
 
-	fn with_validation_data<F>(mut self, f: F) -> Self
-	where
-		F: 'static + Fn(&BlockTests, &mut PersistedValidationData),
-	{
-		self.persisted_validation_data_hook = Some(Box::new(f));
-		self
-	}
-
 	fn with_inherent_data<F>(mut self, f: F) -> Self
 	where
 		F: 'static + Fn(&BlockTests, RelayChainBlockNumber, &mut ParachainInherentData),
@@ -363,14 +354,11 @@ impl BlockTests {
 				}
 				let (relay_parent_storage_root, relay_chain_state) =
 					sproof_builder.into_state_root_and_proof();
-				let mut vfp = PersistedValidationData {
+				let vfp = PersistedValidationData {
 					relay_parent_number,
 					relay_parent_storage_root,
 					..Default::default()
 				};
-				if let Some(ref hook) = self.persisted_validation_data_hook {
-					hook(self, &mut vfp);
-				}
 
 				<ValidationData<Test>>::put(&vfp);
 				NewValidationCode::<Test>::kill();


pallets/parachain-system/src/lib.rs
```diff
@@ -946,6 +948,84 @@ impl<T: Config> GetChannelInfo for Pallet<T> {
...
+		// Read MQC from relay chain state.
+		let relay_state_proof = RelayChainStateProof::new(
+			T::SelfParaId::get(),
+			validation_data.relay_parent_storage_root,
+			relay_chain_state.clone(),
+		)
+		.expect("Invalid relay chain state proof");
+		let relevant_messaging_state = relay_state_proof
+			.read_messaging_state_snapshot()
+			.expect("Invalid messaging state in relay chain state proof");
```
slumber, line 979: This code is repeated in `set_validation_data`. Can leave it as is or refactor to avoid recreating state proof.

pallets/parachain-system/src/lib.rs
```diff
@@ -946,6 +948,84 @@ impl<T: Config> GetChannelInfo for Pallet<T> {
...
+
+		let relay_dmq_head = MessageQueueChain::from(relevant_messaging_state.dmq_mqc_head);
+		let relay_hrmp_heads =
+			relevant_messaging_state.ingress_channels.iter().map(|(para_id, channel)| {
+				let mqc_head = channel.mqc_head.unwrap_or_default();
+				(para_id, MessageQueueChain::from(mqc_head))
+			});
+
+		// DMQ.
+		let dmq_processed_num = (0..=downward_messages.len())
```
slumber, line 989: Issue mentions going backwards in queue, but that would mean O(n**2) traverse for regular cases when the queue isn't exhausted.

slumber:
> The ParachainInherent generation on the node-side should gather all pending messages at the relay-parent from DMP and HRMP channels

I believe this is already the case.

pallets/parachain-system/src/lib.rs
```diff
@@ -946,6 +948,84 @@ impl<T: Config> GetChannelInfo for Pallet<T> {
...
+			.find(|processed| {
```
rphmeier, line 990: this algorithm seems very weird.

Can we not just do a single loop over downward messages with a helper function like this:

```rust
let mut current_mqc_head = last_dmq_head;
let target_mqc_head = relay_dmq_head;

for (i, messages) in 0..messages.iter().enumerate() {
    if current_mqc_head = target_mqc_head { return i }
    current_mqc_head.extend_downwards(messages[i]);
}

panic!("messages in inherent are not consistent with MQC");
```

slumber: As long as we don't know how many messages were already processed, I don't think O(N^2) can be avoided.

In your example, suppose `current_mqc_head` points to somewhere in the middle of the queue, it will certainly panic.

slumber: I don't think `LastRelayChainBlockNumber` is guaranteed to change?

We can use small optimization and compare last mqc with relay mqc, assuming the whole queue was exhausted previously, otherwise, perform search.

rphmeier: > I don't think LastRelayChainBlockNumber is guaranteed to change?

It feels safe to rely on this being updated every block, if that's what you mean.

If the `LastRelayChainBlockNumber = N`, then we can safely omit all messages `sent_at <= N` as they've presumably already been processed. The MQC head and pending messages are drawn out of a child relay-chain block M, so we'd end up providing all messages where `N < sent_at <= M`. If the current relay-parent is the same as the previous, then N = M and no messages will have `N < sent_at <= N` and we'd provide an empty list, which is the desired behavior.

rphmeier:
This doesn't actually close #2472 but addresses the workaround solution for now: https://github.com/paritytech/cumulus/issues/2472#issuecomment-1535535569

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2874733

pallets/parachain-system/src/lib.rs
```diff
@@ -971,6 +973,40 @@ impl<T: Config> GetChannelInfo for Pallet<T> {
 }
 
 impl<T: Config> Pallet<T> {
+	/// Updates inherent data to only contain messages that weren't already processed
+	/// by the runtime based on MQC heads.
+	///
+	/// # Panics
+	///
+	/// Panics if MQC heads don't match even after exhausting incoming message queue, either
```
rphmeier, line 981: This function doesn't panic, docs here are referring to the panic in `set_validation_data`?

slumber: Leftover from previous version of PR, my bad

rphmeier (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2874942

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2882925

PR #2594, opened 2023-05-18
parachain-system: ignore go ahead signal once upgrade is processed

### LINKED ISSUE #2580: Asynchronous backing: handle code upgrade signals properly
When a parachain wants to upgrade its code, it submits the new validation code to the relay chain. After some checks and a mandatory waiting period, the relay-chain sets a signal in its storage - either `GoAhead` or `Abort`. It is illegal to create a candidate that attempts to upgrade the code if there is already a pending code upgrade.

The `GoAhead` signal is cleared in the relay chain runtime after the first candidate which used a relay-parent _at or after_ the relay-chain block where it was set is included. This means that the go-ahead signal may hang around in the relay-chain state for a few blocks afterwards.

pallet-parachain-system currently has this code, which panics if the `GoAhead` signal is set and there is no pending upgrade:

https://github.com/paritytech/cumulus/blob/a82d5afbd7ea477f4160408905c4e4835a1e5970/pallets/parachain-system/src/lib.rs#L404-L423

This means that if the parachain has scheduled a code upgrade, one candidate might see the `GoAhead` signal and process the code upgrade correctly, but the subsequent candidates (at the same relay-parent or within a few blocks of it) will be invalid. They will become valid again only once the relay-parent has advanced to a point where the `GoAhead` signal has vanished from the relay-chain state.

The issue here is just to avoid small amounts of downtime after parachains upgrade their code. The fix is to handle the code upgrade signals "properly" i.e. to ignore them when there is no pending code upgrade or the unincluded segment contains an applied/aborted code upgrade.

slumber:
Closes #2580

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -225,6 +225,7 @@ pub mod pallet {
 		fn on_finalize(_: T::BlockNumber) {
 			<DidSetValidationCode<T>>::kill();
 			<UpgradeRestrictionSignal<T>>::kill();
+			let relay_upgrade_go_ahead = <UpgradeGoAhead<T>>::take();
 
 			assert!(
 				<ValidationData<T>>::exists(),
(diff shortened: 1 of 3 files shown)


rphmeier (commented):
With this approach, we should be careful that we don't submit another code upgrade until the previous go-ahead signal has gone away (i.e. the block which handled the go-ahead signal is included). Otherwise we could be left with an ambiguous interpretation of the go-ahead. This should be tested.

rphmeier (commented):
There is also the related problem that the original issue didn't pick up - that the upgrade restriction signal isn't set until the block scheduling a code upgrade has been included. This means that the unincluded segment should also avoid scheduling code upgrades if either of the following hold:
  1. The upgrade restriction signal disallows it
  2. The unincluded segment schedules a code upgrade

The current logic only does (1) but we should do (2) as well, in this PR.

slumber:
> There is also the related problem that the original issue didn't pick up - that the upgrade restriction signal isn't set until the block scheduling a code upgrade has been included. This means that the unincluded segment should also avoid scheduling code upgrades if either of the following hold:
> 
> 1. The upgrade restriction signal disallows it
> 2. The unincluded segment schedules a code upgrade
> 
> The current logic only does (1) but we should do (2) as well, in this PR.

2 is handled by `PendingValidationCode` storage item.

rphmeier:
> Scheduling another upgrade in the same segment shouldn't be a problem, although, it's very likely there'll be a restriction signal because of upgrade cooldown, e.g. 14400 blocks ~ 24 hours on Polkadot.

What I mean is that depending on the configuration of Polkadot _right now_ is not future proof. If the current code breaks if parachains are "allowed" to submit a code upgrade every block by Polkadot, then the PR implementation is broken.

slumber:
> What I mean is that depending on the configuration of Polkadot _right now_ is not future proof. If the current code breaks if parachains are "allowed" to submit a code upgrade every block by Polkadot, then the PR implementation is broken.

I didn't try to justify it with Polkadot config, this was "although"
Initial statement is that to me it looks correct to submit another upgrade after go-ahead being processed but before the processing block's inclusion.

slumber:
> which is the case. Is this tested?

https://github.com/paritytech/cumulus/blob/974dc4616a9bd3ca82641cc3f80bc317c188e13c/pallets/parachain-system/src/tests.rs#L424-L439

This works regardless of blocks being included or not.

rphmeier (approved):

PR #2608, opened 2023-05-20
Switch to `relay_dispatch_queue_remaining_capacity`

bkchr:
This switches the parachain runtimes to use `relay_dispatch_queue_remaining_capacity` when possible. If the data is not yet available on the relay chain it falls back to `relay_dispatch_queue_size`. It will require that all parachains migrate to `relay_dispatch_queue_remaining_capacity` before we can start removing the call to `relay_dipatch_queue_size`.

Besides that the pr adapts the xcm exumulator to make it work with the message queue.


bkchr:
Left a comment in the tracking issue: https://github.com/paritytech/polkadot/issues/6283

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2864799

pallets/parachain-system/src/relay_state_snapshot.rs
```diff
@@ -24,6 +24,17 @@ use sp_state_machine::{Backend, TrieBackend, TrieBackendBuilder};
...
 
+/// The capacity of the upward message queue of a parachain on the relay chain.
+// The field order should stay the same as the data can be found in the proof to ensure both are
+// have the same encoded representation.
+#[derive(Clone, Encode, Decode, TypeInfo, Default)]
+pub struct RelayDispachQueueSize {
+	/// The number of additional messages that can be enqueued.
+	pub remaining_count: u64,
+	/// The total size of additional messages that can be enqueued.
+	pub remaining_size: u64,
```
ggwpez, line 35: I think the relay still uses `u32` for those (although should be changed to `u64` eventually), so how can this work?

pallets/parachain-system/src/relay_state_snapshot.rs
```diff
@@ -164,20 +172,49 @@ impl RelayChainStateProof {
...
+		//
+		// When the relay chain and all parachains support `relay_dispatch_queue_remaining_capacity`,
+		// this code here needs to be removed and above needs to be changed to `read_entry` that
+		// returns an error if `relay_dispatch_queue_remaining_capacity` can not be found/decoded.
+		//
+		// For now we just fallback to the old dispatch queue size if there is an error.
+		let relay_dispatch_queue_size = match relay_dispatch_queue_size {
+			Ok(Some(r)) => r,
+			_ => {
+				let res = read_entry::<(u32, u32), _>(
```
ggwpez, line 202: >It will require that all parachains migrate to relay_dispatch_queue_remaining_capacity before we can start removing the call to relay_dipatch_queue_size.

We cannot monitor this on the relay chain, or? Can we somehow know, how many parachains already support this?

bkchr: No, this needs to be done manually. Aka we wait one year at least.

pallets/parachain-system/src/relay_state_snapshot.rs
bkchr: I can check again, but it is u64 afair. While I don't get why this is u64 as we will never have soo many remaining messages.

bkchr: You were right, it is `u32`. Fixed!

bkchr:
bot fmt

ggwpez (approved):

sam0x17 (approved):

PR #2610, opened 2023-05-22
fix warning

xlc:
```
warning: calls to `std::mem::drop` with a reference instead of an owned value does nothing
   --> client/consensus/common/src/parachain_consensus.rs:328:4
    |
328 |             drop(unset_best_header);
    |             ^^^^^-----------------^
    |                  |
    |                  argument has type `&<Block as sp_runtime::traits::Block>::Header`
    |
    = note: use `let _ = ...` to ignore the expression or result
    = note: `#[warn(drop_ref)]` on by default
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -236,9 +236,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.71"
+version = "1.0.69"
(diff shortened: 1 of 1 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2868286

PR #2616, opened 2023-05-22
Use default for intergration test relay runtimes

gilescope:
If people are using zombienet with a relay chain that has sudo added to it (for testing) then this is one additional place that `sudo: Default::default()` does not need to be added (it's sufficient to add the sudo pallet to the polkadot repo E.g. the `-fast-sudo` branches).


jsidorenko (approved):

bkontur:
I dont know if it is worth or desired to check this way all genesis chain specs here `polkadot-parachain/src/chain_spec/*.rs,
**pros** is when you add new pallet then you dont need to change it here, but this can be also **cons**

bkontur (approved):

gilescope:
bot merge

PR #2626, opened 2023-05-23
Add Glutton parachain to pr-custom-review

NachoPal:
(Added to the exclusion as the parachain has no economic value)

### DIFF
--- a/.github/pr-custom-review.yml
+++ b/.github/pr-custom-review.yml
@@ -6,7 +6,7 @@ action-review-team: ci
 rules:
   - name: Runtime files
     check_type: changed_files
-    condition: ^parachains/runtimes/assets/(statemine|statemint)/src/[^/]+\.rs$|^parachains/runtimes/bridge-hubs/(bridge-hub-kusama|bridge-hub-polkadot)/src/[^/]+\.rs$|^parachains/runtimes/collectives/collectives-polkadot/src/[^/]+\.rs$|^parachains/common/src/[^/]+\.rs$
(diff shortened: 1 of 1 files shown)


bkontur (approved):

gilescope (approved):

.github/pr-custom-review.yml
```diff
@@ -6,7 +6,7 @@ action-review-team: ci
 rules:
   - name: Runtime files
     check_type: changed_files
-    condition: ^parachains/runtimes/assets/(statemine|statemint)/src/[^/]+\.rs$|^parachains/runtimes/bridge-hubs/(bridge-hub-kusama|bridge-hub-polkadot)/src/[^/]+\.rs$|^parachains/runtimes/collectives/collectives-polkadot/src/[^/]+\.rs$|^parachains/common/src/[^/]+\.rs$
+    condition: ^parachains/runtimes/assets/(statemine|statemint)/src/[^/]+\.rs$|^parachains/runtimes/bridge-hubs/(bridge-hub-kusama|bridge-hub-polkadot)/src/[^/]+\.rs$|^parachains/runtimes/collectives/collectives-polkadot/src/[^/]+\.rs$|^parachains/runtimes/glutton/glutton-kusama/src/[^/]+\.rs$|^parachains/common/src/[^/]+\.rs$
```
joepetrowski, line 9: > (Added to the exclusion as the parachain has no economic value)

This seems to do the opposite

NachoPal: Yeah, not my comment, sorry. I followed the same rules as the rest of the paras, but if it is not necessary because of "no economic value", then we just need to close this PR.

joepetrowski: The PR title and description contradict each other. The exclusion only applies to things that are getting two reviews from locks developers. I don't see the need to change it, I think 2 core devs are OK.

NachoPal: > The PR title and description contradict each other. 

Yeah, that happened because PR title and description comment are not from the same person :P

Agree, closing the PR.

PR #2649, opened 2023-05-28
Support westend integration tests

gilescope:
Things appear in westend before kusama, (example asset-conversion pallet) so it's helpful if we can write integration tests targeting westend.

Also updates emulator to handle host version changes (westend is on v5 of the host interface)


NachoPal:
> @NachoPal any problems with this?

All good

NachoPal (approved):

parachains/integration-tests/emulated/assets/asset-hub-westend/src/tests/transact.rs
```diff
@@ -0,0 +1,58 @@
+use crate::*;
+
+#[test]
+fn transact_sudo_from_relay_to_assets_para() {
+	// Init tests variables
+	// Call to be executed in Assets Parachain
+	const ASSET_ID: u32 = 1;
+
+	let call = <AssetHubWestend as Para>::RuntimeCall::Assets(pallet_assets::Call::<
```
joepetrowski, line 9: Would be nice to do this with foreign assets and register PAL.

parachains/integration-tests/emulated/assets/asset-hub-westend/src/tests/transact.rs
```diff
@@ -0,0 +1,58 @@
...
+	.encode()
+	.into();
+
+	// XcmPallet send arguments
+	let sudo_origin = <Westend as Relay>::RuntimeOrigin::root();
+	let assets_para_destination: VersionedMultiLocation =
+		Westend::child_location_of(AssetHubWestend::para_id()).into();
+
+	let weight_limit = WeightLimit::Unlimited;
+	let require_weight_at_most = Weight::from_parts(1000000000, 200000);
```
joepetrowski, line 27: This can be a follow up project (maybe good for someone new), but we should have some core set of "user stories" that other chains would use transact for, and then set the weight values in these tests as `actual * some_factor_of_safety`, and if the tests fail in a new release because of weight, then we need to mark it as a breaking change. That way people can rely on these weights in tests to have some future proof-ness in their runtimes.

parachains/integration-tests/emulated/assets/asset-hub-westend/src/tests/reserve_transfer.rs
```diff
@@ -0,0 +1,65 @@
+use crate::*;
+
+#[test]
+fn reserve_transfer_native_asset_from_relay_to_assets() {
+	// Init tests variables
+	let amount = POLKADOT_ED * 1000;
```
joepetrowski, line 6:
```suggestion
	let amount = WESTEND_ED * 1000;
```

parachains/integration-tests/emulated/assets/asset-hub-westend/src/tests/teleport.rs
```diff
@@ -0,0 +1,62 @@
+use crate::*;
+
+#[test]
+fn teleport_native_assets_from_relay_to_assets_para() {
+	// Init tests variables
+	let amount = POLKADOT_ED * 1000;
```
joepetrowski, line 6:
```suggestion
	let amount = WESTEND_ED * 1000;
```

joepetrowski (approved):

parachains/integration-tests/emulated/assets/asset-hub-westend/src/tests/transact.rs
gilescope: Absolutely. That kind of thing will be coming in future PRs.

PR #2658, opened 2023-05-30
aura-ext: check slot in consensus hook and remove all `CheckInherents` logic

### LINKED ISSUE #2588: Aura Consensus Hook should enforce the relay-parent slot is equal to the Aura slot
https://github.com/paritytech/cumulus/pull/2551 introduced a `ConsensusHook` for Aura based on https://github.com/paritytech/cumulus/pull/2501 .

We require for Aura that the slot for parachains is set based on the slot of the relay-parent, as it's the only reliable indication of time we have access to within parachain runtimes. These checks are currently done through the `parachain_system::CheckInherents` trait that is evaluated during `validate_block`. Moving this check to the consensus hook would make it run during all block executions - both during `validate_block` and vanilla `execute_block` as used by parachain nodes. It is also a refactoring to make the logic more clear.

slumber:
Resolves #2588

This PR also removes the `CheckInherents` functionality used during Cumulus block execution, as that is now replaced by the `ConsensusHook` interface of `parachain_system`.


pallets/aura-ext/src/consensus_hook.rs
```diff
@@ -19,28 +19,42 @@
...
 
+const RELAY_CHAIN_MILLIS_PER_BLOCK: u64 = 6000;
+
 /// A consensus hook for a fixed block processing velocity and unincluded segment capacity.
 pub struct FixedVelocityConsensusHook<T, const V: u32, const C: u32>(PhantomData<T>);
 
 impl<T: pallet::Config, const V: u32, const C: u32> ConsensusHook
 	for FixedVelocityConsensusHook<T, V, C>
+where
+	<T as pallet_timestamp::Config>::Moment: Into<u64>,
```
slumber, line 39: `Aura::slot_duration` returns `T::Moment` with a definition
```rust
type Moment: Parameter
			+ Default
			+ AtLeast32Bit // From<u16> + From<u32>
			+ Scale<Self::BlockNumber, Output = Self::Moment>
			+ Copy
			+ MaxEncodedLen
			+ scale_info::StaticTypeInfo;
```
And `SlotDuration` cannot be built out of it.

slumber: Couldn't come up with any other way. Ideally, I would expect `Aura::slot_duration` to return `SlotDuration`.
As you can see from the type definition, it's not convertible to any numeric primitive, so we have to enforce it in boundary.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2934988

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2947621

pallets/aura-ext/src/consensus_hook.rs
```diff
@@ -19,28 +19,51 @@
...
-	fn on_state_proof(_state_proof: &RelayChainStateProof) -> (Weight, UnincludedSegmentCapacity) {
+	fn on_state_proof(state_proof: &RelayChainStateProof) -> (Weight, UnincludedSegmentCapacity) {
 		// Ensure velocity is non-zero.
 		let velocity = V.max(1);
+		let relay_chain_slot = state_proof.read_slot().expect("failed to read relay chain slot");
 
-		let authored = pallet::Pallet::<T>::slot_info()
-			.map(|(_slot, authored)| authored)
+		let (slot, authored) = pallet::Pallet::<T>::slot_info()
 			.expect("slot info is inserted on block initialization");
```
slumber, line 55: @rphmeier
Now that integration tests are failing due to this panic, I wonder if it's OK to optionally verify the slot and skip it if the value is not present?

pallets/aura-ext/src/consensus_hook.rs
rphmeier: It's reasonable to enforce it in the boundary as long as that aligns with the instantiations that are typically used in runtime declarations. Seems like a good solution to me.

pallets/aura-ext/src/consensus_hook.rs
rphmeier: I would guess that it's most likely the case that blocks built in tests don't contain the pre-runtime digest which sets the slot in `pallet_aura` and that they need to.

rphmeier (approved):

pallets/aura-ext/src/consensus_hook.rs
```diff
@@ -19,28 +19,51 @@
...
+	ConsensusHook for FixedVelocityConsensusHook<T, RELAY_CHAIN_SLOT_DURATION, V, C>
+where
+	<T as pallet_timestamp::Config>::Moment: Into<u64>,
 {
 	// Validates the number of authored blocks within the slot with respect to the `V + 1` limit.
-	fn on_state_proof(_state_proof: &RelayChainStateProof) -> (Weight, UnincludedSegmentCapacity) {
+	fn on_state_proof(state_proof: &RelayChainStateProof) -> (Weight, UnincludedSegmentCapacity) {
 		// Ensure velocity is non-zero.
 		let velocity = V.max(1);
+		let relay_chain_slot = state_proof.read_slot().expect("failed to read relay chain slot");
```
BradleyOlson64, line 52: What happens here if the collator isn't up to date with the latest relay block? I think a slot mismatch would occur.

Right now it seems that collation is still triggered by new incoming relay blocks, so the scenario I described won't occur. But my understanding is that we will want to unteather these.

BradleyOlson64 (approved):
Nice 👍

pallets/aura-ext/src/consensus_hook.rs
slumber: I believe the state proof comes from the relay parent the para is building on top of.

rphmeier: That's right - the parachain has no information about what the latest relay-parent is, other than what the collator gives it. But validators will only accept blocks with recent relay-parents.

PR #2681, opened 2023-06-02
Glutton update script

NachoPal:
Script to update Gluttons Parachains. It allows to update `Compute` (percentage of the left over `ref_time` to consume during `on_idle`) and `Storage` (percentage of left over `proof_size` to consume during `on_idle`) storage items for a range of selected parachain ids.

Necessary to learn what are the limits of the Relay chain and subsequently the max number of Parachains it can handle.


scripts/glutton/update/src/main.rs
```diff
@@ -0,0 +1,173 @@
...
+		println!("- Connecting to client...");
+		let client: OnlineClient<GluttonConfig>;
+
+		match args.relay {
+			RelayChain::Local => {
+				client = OnlineClient::<GluttonConfig>::from_url("ws://127.0.0.1:9810").await?;
+			},
+			RelayChain::Versi => {
+				client = OnlineClient::<GluttonConfig>::from_url(format!(
+					"wss://versi-glutton-collator-{}-node-1.parity-versi.parity.io:443",
```
sandreim, line 135: I would make this a CLI argument.

scripts/glutton/update/src/main.rs
```diff
@@ -0,0 +1,173 @@
...
+
+	/// Update Glutton Parachains from 'para_id = from' (inclusive)
+	#[arg(short, long)]
+	from: u32,
+
+	/// Update Glutton Parachains to 'para_id = to' (inclusive)
+	#[arg(short, long)]
+	to: u32,
+
+	/// Optional new 'storage' value to update
```
sandreim, line 37: Needs to mention what the unit of measurement is.

scripts/glutton/update/src/main.rs
```diff
@@ -0,0 +1,173 @@
...
+	#[arg(short, long)]
+	storage: Option<u32>,
+
+	/// Optional new 'compute' value to update
```
sandreim, line 41: same here

scripts/glutton/update/src/main.rs
sandreim: The entire match would not be needed if the format string is passed as an argument.

sandreim (approved):
LGTM, only a few suggestions.

scripts/glutton/update/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[workspace]
```
ggwpez, line 1: Why should this not be in the main workspace?  
This will make it harder for the monorepo and when updating dependencies.

scripts/glutton/update/src/main.rs
NachoPal: The problem I see there is that the endpoint depends on the `para_id`. Endpoints format is going to be stable anyway.

scripts/glutton/update/Cargo.toml
NachoPal: I thought about that, but on the other hand I didn't want to increase Cumulus compile time, which is already high. After Gluttons Parachains fulfil their mission, I do not think we'll need this script anymore. What do you think?

ggwpez: It does not increase compile time, since this is not part of the cumulus binary. I hope most devs use something like `cargo c -p polkadot-parachain`.

rphmeier:
What does it mean to "update" a glutton parachain? Needs more description.

bkchr:
I have the strong feeling that this could be a 10 line polkadot js script?

NachoPal:
> What does it mean to "update" a glutton parachain? Needs more description.

Improved description

scripts/glutton/update/Cargo.toml
NachoPal: Ok, I moved it to the main workspace

gilescope (approved):
Like the use of subxt!

gilescope:
My only concern is the Cargo.lock changes. At the moment subxt is bringing in another version of substrate. Can/should we persuade it to use the same version of substrate as cumulus does?

scripts/glutton/update/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+
+[[bin]]
+name = "glutton-update"
+path = "src/main.rs"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.4.0", default-features = false }
+subxt = { version = "0.29.0", features = ["unstable-metadata"] }
+sp-keyring = "23.0.0"
```
gilescope, line 16: I think we can nix this dep, don't think it's needed. Seems to compile on my machine without it.
```suggestion
```

scripts/glutton/update/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+description = "Script to update Glutton parachains storage values"
+
+
+[[bin]]
+name = "glutton-update"
+path = "src/main.rs"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.4.0", default-features = false }
+subxt = { version = "0.29.0", features = ["unstable-metadata"] }
```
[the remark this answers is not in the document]
NachoPal: The alternative I initially used was to query the `nonce` from `frame_system` storage, but with the `["unstable-metadata"]` feature I could directly use the runtime API to query it. All this is coming from the fact that the Shell client does not implement the `system_accountNextIndex` method.

In case it breaks in the future I can always bring back the former approach. I'll keep it in mind, thanks for the heads up.

scripts/glutton/update/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+		"AccountNonceApi",
+		"account_nonce",
+		vec![Value::from_bytes(account)],
+	);
+
+	let nonce = client
+		.runtime_api()
+		.at_latest()
+		.await?
+		.call(runtime_api_call)
```
jsdw, line 73: If you want to avoid the unstable-metadata thing you could also use (assuming nonce is a u32, which is true in substrate/polkadot and I'd expect most other chains):

```rust
let next_nonce: u32 = client
    .runtime_api()
    .at_latest()
    .await?
    .call_raw("AccountNonceApi_account_nonce", Some(&account.encode()))
    .await?;
```

I'm actually going to move Subxt to do this too, and will likely expose a way to get this in the next version

scripts/glutton/update/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+	let tx = glutton_para::tx().sudo().sudo(call);
+
+	// Build the signed tx
+	let signed_tx = client
+		.tx()
+		.create_signed_with_nonce(&tx, signer, nonce, Default::default())
+		.unwrap();
+
+	// Submit and watch the tx
+	let in_block = signed_tx.submit_and_watch().await.unwrap().wait_for_in_block().await.unwrap();
```
jsdw, line 101: Just to note that `.wait_for_in_block()` doesn't guarantee that the block will be finalized; use `.wait_for_finalized()` for that, or even `.wait_for_finalized_success()` to also verify that the tx was successfully executed (ie an ExtrinsicSuccess event was produced from its inclusion)

jsdw:
> My only concern is the Cargo.lock changes. At the moment subxt is bringing in another version of substrate. Can/should we persuade it to use the same version of substrate as cumulus does?

Just to reiterate I said in DM; subxt won't bring in any substrate crates (except `sp-core-hashing` which is tiny) unless used with the "substrate-compat" feature flag so this hsouldn't matter :)

scripts/glutton/update/Cargo.lock
```diff
@@ -0,0 +1,4403 @@
+# This file is automatically @generated by Cargo.
```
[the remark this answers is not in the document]
NachoPal: If we do not add it to the main workspace, what is the issue for having other `Cargo.lock` files on their own package?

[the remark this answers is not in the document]
ggwpez: The dependabot does not update it (so the deps will be out dated) and our CI tooling like fmt, clippy will not work.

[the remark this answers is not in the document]
NachoPal: I do not mind at all to move it to its own repo. It is just a tool to help to achieve a higher goal which is Glutton's Parachains management. It is not mandatory to include it in Cumulus.

But if it is true that adding `subxt` to the main workspace could potentially lead to multiple versions of the same crate, and at the same time, we do not want to have multiple `Cargo.lock` files, then it means we are banning `subxt` from Cumulus/monorepo.

jsdw:
> > My only concern is the Cargo.lock changes. At the moment subxt is bringing in another version of substrate. Can/should we persuade it to use the same version of substrate as cumulus does?
> 
> Just to reiterate I said in DM; subxt won't bring in any substrate crates (except `sp-core-hashing` which is tiny) unless used with the "substrate-compat" feature flag so this hsouldn't matter :)

Actually this isn't true; the `substrate-compat` feature is enabled by default, and you need the substrate crates currently to sign transactions, and Subxt needs to pull these from crates.io, so there's not currently a way to avoid the dupe deps at the mo!

I am working on adding some native signing support to Subxt so that you can do basic transaction signing like here without pulling in a load of substrate crates.

Alternately I'm hoping for a future where substrate crates are published regularly again, and polkadot/cumulus depend on the crates and not github master; it'll be easier to keep things like Subxt in line with the latest deps then :)

scripts/glutton/update/Cargo.lock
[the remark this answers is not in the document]
bkchr: > I do not mind at all to move it to its own repo

I would personally favour this.

NachoPal:
Closing in favour of: https://github.com/paritytech/parachains-utils/pull/1

PR #2711, opened 2023-06-08
Xcm Emulator: prepare XCMP on init

muharem:
The XCM messages are not passing between parachains right now, without manually calling `init` and `reset` before a test.
In the PR we preparing/opening the XCMP channels on a network initialization which happens once when `execute_with` invoke on any chain.
Calling just `reset` like in the examples [here](https://github.com/shaunxw/xcm-simulator/blob/d011e5ca62b93e8f688c2042c1f92cdbafc5d1d0/xcm-emulator/example/src/lib.rs#L195) does not help, since static params like PARA_IDS are not initialized. 

The PR also includes missing license headers for it tests.


gilescope:
The other tests seem to work without needing to call `prepare_for_xcmp`. Does this fix just mean better emulation?

NachoPal (approved):

PR #2714, opened 2023-06-08
Add account nonce api to Glutton Parachain

NachoPal:
Now that there is `sudo` in the Glutton's runtime, runtime API is needed to check sudo txs `nonce`.
Needed for https://github.com/paritytech/parachains-utils/pull/1 to work

### DIFF
--- a/parachains/runtimes/glutton/glutton-kusama/src/lib.rs
+++ b/parachains/runtimes/glutton/glutton-kusama/src/lib.rs
@@ -352,6 +352,12 @@ impl_runtime_apis! {
 		}
 	}
 
+  impl frame_system_rpc_runtime_api::AccountNonceApi<Block, AccountId, Index> for Runtime {
+		fn account_nonce(account: AccountId) -> Index {
+			System::account_nonce(account)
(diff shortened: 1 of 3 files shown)


parachains/runtimes/glutton/glutton-kusama/src/lib.rs
```diff
@@ -352,6 +352,12 @@ impl_runtime_apis! {
 		}
 	}
 
+  impl frame_system_rpc_runtime_api::AccountNonceApi<Block, AccountId, Index> for Runtime {
+		fn account_nonce(account: AccountId) -> Index {
+			System::account_nonce(account)
+		}
+	}
+
```
gilescope, line 360: cargo fmt?

gilescope (approved):

bkontur (approved):

parachains/runtimes/glutton/glutton-kusama/src/lib.rs
NachoPal: the CI passed

Issue #2703, opened 2023-06-07
runtime integration test don't initialize pallets


slumber:
Tests in question are `parachains/integration-tests`. @rphmeier suggested the issue is related to
https://github.com/paritytech/cumulus/blob/31b1469d830aaa7b0441d4d60f0c0c17be0cf114/xcm/xcm-emulator/src/lib.rs#L193-L195

For instance, explicit panic in `aura-ext` pallet on initialization isn't triggered.
Related to #2658

slumber:
cc @muharem

rphmeier:
The issue we were observing was that with the new logic from https://github.com/paritytech/cumulus/pull/2658/, https://github.com/paritytech/cumulus/pull/2501, https://github.com/paritytech/cumulus/pull/2551 , that integration tests for all runtimes began to fail when integrating the new Aura hook.

The change in behavior is that `ParachainSystem::set_validation_data` can now require some other pallets to initialized properly (i.e. when calling the configured `ConsensusHook`).

I see that the logic in `decl_test_parachains` is directly invoking `on_initialize` and `set_validation_data`, so I see two solutions going forward:
  * Add `AuraExt` to the `pallets_extra` field for the macro for each runtime and ensure it is initialized
  * Alter the macro to initialize all pallets

bkchr:
>   * Alter the macro to initialize all pallets

If that isn't done currently, it sounds like the tests are not being setup correctly.

PR #2722, opened 2023-06-09
Make `xcm-emulator` test pass for new consensus hook

NachoPal:
This PR tries to solve: https://github.com/paritytech/cumulus/issues/2703

Tests are not panicking anymore with `slot info is inserted on block initialization`, but now they are running into another issue. Tests are failing with `included head not present in relay storage proof`.

Should I add the rest of the hooks (`on_finalize`, etc)?


slumber:
We need a full parent header to be included into the storage proof. For example, this is how it's done for parachain system tests
https://github.com/paritytech/cumulus/blob/6141995fa6bcee862d2d0ffaefb641b52a8e3b6c/pallets/parachain-system/src/tests.rs#L403-L404

https://github.com/paritytech/cumulus/blob/6141995fa6bcee862d2d0ffaefb641b52a8e3b6c/pallets/parachain-system/src/tests.rs#L347-L351

To my understanding, headers are not accessible from the runtime, so we need system pallet to be finalized.

muharem:
I would say that the emulator should behave as close as possible to the real environment, while only mocking the network layer. `on_finalize` should also run for all pallets, in a right order.

parachains/integration-tests/emulated/common/src/lib.rs
```diff
@@ -27,6 +27,7 @@ decl_test_relay_chains! {
 			MessageQueue: polkadot_runtime::MessageQueue,
 			XcmConfig: polkadot_runtime::xcm_config::XcmConfig,
 			SovereignAccountOf: polkadot_runtime::xcm_config::SovereignAccountOf,
+			AllPallets: polkadot_runtime::AllPalletsWithSystem,
 			System: polkadot_runtime::System,
 			Balances: polkadot_runtime::Balances,
```
muharem, line 32: not sure if we even need to list the pallets, and pallets_extra. 
may be optionally, if there is some use case for it.

xcm/xcm-emulator/src/lib.rs
```diff
@@ -553,16 +563,26 @@ macro_rules! __impl_test_ext_for_parachain {
...
+					})
+				});
 
 				let para_id = <$name>::para_id().into();
 
 				$ext_name.with(|v| {
 					v.borrow_mut().execute_with(|| {
 						// Make sure it has been recorded properly
-						let relay_block_number = <$name>::relay_block_number();
+						// let relay_block_number = <$name>::relay_block_number();
```
muharem, line 585:
```suggestion
```

NachoPal:
bot fmt

NachoPal:
I dropped the idea of running all pallets `on_initialized` and `on_finalized` with `AllPalletsWithSystem`. The reason is I was running into more errors because of multiple reasons. One of them was the slot verification when using defaults pre-runtime digest. It forced me to manually build them + having to set inherent for timestamp etc...

I realised I was building something very specific for our runtimes using `Aura` and `AuraExt`. One of the goals of `xcm-emulator` is to be as generic as possible, so other parachains (that are maybe using another consensus system) can use it too.

I created this new issue https://github.com/paritytech/cumulus/issues/2734 to add `on_initialize` and `on_finalize` optional hooks to `xcm-emulator`.

The changes in this PR should be good enough to make the test pass, and therefore not being a blocker anymore.

Please merge if you are ok with it.

gilescope (approved):
It's a closer emulation than what we've currently got so it gets my vote.

slumber (approved):

slumber:
Thanks @NachoPal!

PR #2731, opened 2023-06-13
Add set_current_relay_chain_state method for benchmarks

tmpolaczyk:
This is very similar to the already existing `set_block_number`, but it also allows setting the `relay_parent_storage_root`.

We need this to be able to benchmark extrinsics that use StorageProofs in our project.

### DIFF
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -1166,6 +1166,13 @@ pub trait RelaychainStateProvider {
 	///
 	/// **NOTE**: This is not guaranteed to return monotonically increasing relay parents.
 	fn current_relay_chain_state() -> RelayChainState;
+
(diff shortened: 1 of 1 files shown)


ggwpez:
Not sure if you can also just make `ValidationData` public and write to it directly?

girazoki:
yes this would probably also work. Is that preferred?

tmpolaczyk:
Is there a way to make it public only if the `runtime-benchmarks` feature is enabled?

girazoki:
@ggwpez Any update on the appraoch to follow? do you prefer this or making the storage item public?

ggwpez (approved):

ggwpez:
Did not see, sorry.

ggwpez:
bot merge

PR #2013, opened 2022-12-21
Transfer asset via bridge

bkontur:
PR introduces:
- new pallet to support transfer over bridge-hubs and (xcm) configuration
- asset-hub-kusama configuration for sending/receiving assets (only KSM/DOT) over bridge-hub to asset-hub-polkadot
- asset-hub-polkadot configuration for sending/receiving assets (only KSM/DOT) over bridge-hub to asset-hub-kusama
- asset-hub-westend configuration for sending/receiving in assets over bridge-hub (kusama-local - for local testing, rococo local+live setup)

(even if pallet is configured in runtimes and deployed live, there is no way to initiate transfer unless HRMPs to bridge-hubs are opened)

---
- [x] fix Weight for `fn universal_origin(_: &Junction) -> Weight {`
- [x] UniversalAliases/TrustedBridgedNetworks - create some general pallet for on-chain cfg (something like bridge-asset-transfer has, maybe split)
- [x] move all usefull origin_conversion/barriers stuff to the polkadot/xcm-builder/xcm-executor - https://github.com/paritytech/polkadot/pull/7016
- [x] investigate/check/add RefundSurplus + ClearOrigin -> add to test where it is trapped (now, BuyExecution pays SA of remote statemine/t but assets are trapped to local bridge hub account, probably because of ClearOrigin)
- [ ] verify and prepare test for ETH scenario - https://github.com/paritytech/cumulus/pull/2013#discussion_r1094909290 
- [x] setup both ways Statemine <-> Westmint
- [ ] add integaration e2e/emulated tests for moving assets (native, trusted, foreign, foreign from different consensus, eth, ...)
- [x] fix SA - local-bridge-hub vs remote target location
- [x] fix benchmark
- [ ] fix local run
- [ ] open follow up PR with on-chain configuration by governance
- [x] allow set_storage for fee
---

**Note: do not merge until https://github.com/paritytech/cumulus/pull/2642 is merged or remove TrailingSetTopicAsId from here and fix test**

PR #2754, opened 2023-06-19
BHs readme nits

bkontur:


parachains/runtimes/bridge-hubs/README.md
```diff
@@ -191,65 +195,10 @@ RUST_LOG=runtime=trace,rpc=trace,bridge=trace \
 	- Pallet: **bridgeRococoParachain**
 	- Keys: **bestParaHeads()**
 
-### Send messages
-
-#### Local zombienet run
```
acatangiu, line 196: These instructions are useful for setting up local testing, why remove them (I don't see them in #2013 README.md so looks like they are lost if we remove here)?

bkontur:
bot rebase

parachains/runtimes/bridge-hubs/README.md
bkontur: those are not actual and I will add them for ro/wo in new branch `origin/bko-transfer-asset-via-bridge-pallet-xcm-ro-wo` with `pallet_xcm` support for transfer

PR #2763, opened 2023-06-20
Contracts-rococo: Update migration

pgherveou:


agryaznov (approved):

ggwpez (commented):
We added some structure to this in Polkadot:  
https://github.com/paritytech/polkadot/blob/0e614b09b842583d5e08d430e2c56254018b92d8/runtime/kusama/src/lib.rs#L1505

That makes it a bit easier to remove old ones.

pgherveou:
> We added some structure to this in Polkadot: https://github.com/paritytech/polkadot/blob/0e614b09b842583d5e08d430e2c56254018b92d8/runtime/kusama/src/lib.rs#L1505
> 
> That makes it a bit easier to remove old ones.

@ggwpez  Are you suggesting that we should set the migration here to something like ? 
```rust
pub type Migrations = (rococo_runtime::migrations::V0943, ...);
```
because we passed a migration in between 42 and 43 (my bad for testing the multiblock migration system), I think we should merge this PR as is (as it's passing try_runtime tests) and then for future release only uses the wasm that are published in the release page

athei (approved):

pgherveou:
bot merge

PR #2778, opened 2023-06-25
Allow functions to work over both parachains and relay chains

gilescope:
Moving common types into a common `Chain` trait so that generic functions can be written against them.

(Should Balances also move? I don't think Balances pallet should be essential to a parachain so have not pulled that type up.)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3067751

NachoPal (approved):
Curious about what common function you need.

NachoPal (changes requested):
Please do not merge yet

PR #2784, opened 2023-06-26
Runtime: Success value and reachable location for Polkadot Collectives benchmarks

muharem:
This Pull Request addresses two issues encountered during the benchmark run:

1. Ranked collectives benchmarks:
An error occurs when the benchmarks attempt to promote a member above rank 9.
To resolve this, we have set the success value for the Root as the maximum value of u16.
We have employed a similar approach in the Polkadot runtime.

2. Salary pallet benchmarks:
In the benchmark environment, sending a message to a Parachain fails due to the absence of a channel between the Parachains.
To address this, we wrap the Pay implementation to open a channel with ensure_successful.

### DIFF
--- a/parachains/runtimes/collectives/collectives-polkadot/src/fellowship/mod.rs
+++ b/parachains/runtimes/collectives/collectives-polkadot/src/fellowship/mod.rs
@@ -105,9 +105,14 @@ pub type FellowshipCollectiveInstance = pallet_ranked_collective::Instance1;
 impl pallet_ranked_collective::Config<FellowshipCollectiveInstance> for Runtime {
 	type WeightInfo = weights::pallet_ranked_collective::WeightInfo<Runtime>;
 	type RuntimeEvent = RuntimeEvent;
+
+	#[cfg(not(feature = "runtime-benchmarks"))]
 	// Promotions and the induction of new members are serviced by `FellowshipCore` pallet instance.
+	type PromoteOrigin = frame_system::EnsureNever<pallet_ranked_collective::Rank>;
+	#[cfg(feature = "runtime-benchmarks")]
 	// The maximum value of `u16` set as a success value for the root to ensure the benchmarks will pass.
 	type PromoteOrigin = EnsureRootWithSuccess<Self::AccountId, ConstU16<65535>>;
+
 	// Demotion is by any of:
 	// - Root can demote arbitrarily.
 	// - the FellowshipAdmin origin (i.e. token holder referendum);
--- a/parachains/runtimes/collectives/collectives-polkadot/src/fellowship/mod.rs
+++ b/parachains/runtimes/collectives/collectives-polkadot/src/fellowship/mod.rs
@@ -106,12 +106,15 @@ impl pallet_ranked_collective::Config<FellowshipCollectiveInstance> for Runtime 
 	type WeightInfo = weights::pallet_ranked_collective::WeightInfo<Runtime>;
 	type RuntimeEvent = RuntimeEvent;
 	// Promotions and the induction of new members are serviced by `FellowshipCore` pallet instance.
-	type PromoteOrigin = EnsureRootWithSuccess<Self::AccountId, ConstU16<{ ranks::DAN_9 }>>;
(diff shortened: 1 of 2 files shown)


parachains/runtimes/collectives/collectives-polkadot/src/fellowship/mod.rs
```diff
@@ -106,12 +106,15 @@ impl pallet_ranked_collective::Config<FellowshipCollectiveInstance> for Runtime
 	type WeightInfo = weights::pallet_ranked_collective::WeightInfo<Runtime>;
 	type RuntimeEvent = RuntimeEvent;
 	// Promotions and the induction of new members are serviced by `FellowshipCore` pallet instance.
-	type PromoteOrigin = EnsureRootWithSuccess<Self::AccountId, ConstU16<{ ranks::DAN_9 }>>;
+	// The maximum value of `u16` set as a success value for the root to ensure the benchmarks will pass.
+	type PromoteOrigin = EnsureRootWithSuccess<Self::AccountId, ConstU16<65535>>;
```
gilescope, line 110: should we have this behind `#[cfg(feature = "runtime-benchmarks")]` and have DAN_9 if the feature is not enabled?

gilescope: If this was pulled out into a type def then that type could also be referenced in the DemoteOrigin

muharem: I liked this way for simplicity, both ways looks ok to me.

muharem: i added a separate promote origin for benches, please approve if looks good

parachains/runtimes/collectives/collectives-polkadot/src/fellowship/mod.rs
```diff
@@ -182,8 +190,18 @@ pub type FellowshipSalaryInstance = pallet_salary::Instance1;
 
 use xcm::prelude::*;
 
+#[cfg(not(feature = "runtime-benchmarks"))]
 parameter_types! {
 	pub AssetHub: MultiLocation = (Parent, Parachain(1000)).into();
+}
+
+#[cfg(feature = "runtime-benchmarks")]
```
bkontur, line 198: if benchmark uses XcmRouter at the end, then in this case it would add different weights for `ParentAsUmp` and not for `XcmpQueue` as expected for `AssetHub` 

there is feature to open HRMP channel for benchmarks `open_outbound_hrmp_channel_for_benchmarks` e.g. here: https://github.com/paritytech/cumulus/blob/master/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L1054

I dont know which benchmark/pallet here needs this, but probably the clean way, would be to extend pallet's Config for benchmark and benchmark itself with some `pre_initialization`, e.g.: here https://github.com/paritytech/substrate/blob/master/frame/system/benchmarking/src/lib.rs#L36-L45

bkontur:
to my comment above,

>        To address this, we have utilized the Parent location as a destination, following the same approach used in the pallet_xcm benchmarks.

for pallet_xcm_benchmarks is it probably ok, because pallet_xcm can really use Parent in calls,
the question is if `ParentAsUmp` is worst case or `XcmpQueue`

muharem:
@bkontur thanks! makes total sense to me, i am checking how to make it work more accurate

parachains/runtimes/collectives/collectives-polkadot/src/fellowship/mod.rs
muharem: please check now

bkontur (approved):
@muharem 
I cannot say if `ranks::DAN_9` -> `65535` is ok or not, but besides that the rest lgtm

gavofyork (approved):

gavofyork:
@bkchr needs a locks review...

bkchr (approved):

bkontur:
bot merge

PR #2794, opened 2023-06-28
introduce a method for fetching relay chain header to `RelayChainInterface`

seunlanlege:
This is particularly useful for fetching arbitrary relay chain headers within inherents.

### DIFF
--- a/client/relay-chain-interface/src/lib.rs
+++ b/client/relay-chain-interface/src/lib.rs
@@ -112,7 +112,7 @@ pub trait RelayChainInterface: Send + Sync {
 	async fn best_block_hash(&self) -> RelayChainResult<PHash>;
 
 	/// Fetch the block header of a given height
-	async fn header(&self, block_id: BlockId) -> RelayChainResult<PHeader>;
+	async fn header(&self, block_id: BlockId) -> RelayChainResult<Option<PHeader>>;
 
 	/// Get the hash of the finalized block.
 	async fn finalized_block_hash(&self) -> RelayChainResult<PHash>;
@@ -298,7 +298,7 @@ where
 		(**self).new_best_notification_stream().await
 	}
 
-	async fn header(&self, block_id: BlockId) -> RelayChainResult<PHeader> {
+	async fn header(&self, block_id: BlockId) -> RelayChainResult<Option<PHeader>> {
 		(**self).header(block_id).await
 	}
 }
--- a/client/relay-chain-interface/src/lib.rs
+++ b/client/relay-chain-interface/src/lib.rs
@@ -26,6 +26,7 @@ use jsonrpsee_core::Error as JsonRpcError;
 use parity_scale_codec::Error as CodecError;
 use sp_api::ApiError;
 
+use cumulus_primitives_core::relay_chain::BlockId;
 pub use cumulus_primitives_core::{
 	relay_chain::{
 		CommittedCandidateReceipt, Hash as PHash, Header as PHeader, InboundHrmpMessage,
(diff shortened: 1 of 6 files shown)


BradleyOlson64 (approved):

BradleyOlson64:
Looks like @rphmeier introduced a method with the same intent and a slightly different signature [here](https://github.dev/paritytech/cumulus/blob/e00621fef26bbaf9ebdd6646936101a21e696ac2/client/relay-chain-interface/src/lib.rs#L117).

Perhaps your version could be used instead.

seunlanlege:
Yeah my use case requires block numbers

client/relay-chain-interface/src/lib.rs
```diff
@@ -110,6 +111,9 @@ pub trait RelayChainInterface: Send + Sync {
 	/// Get the hash of the current best block.
 	async fn best_block_hash(&self) -> RelayChainResult<PHash>;
 
+	/// Fetch the block header of a given height
+	async fn header(&self, block_id: BlockId) -> RelayChainResult<PHeader>;
```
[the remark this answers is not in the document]
seunlanlege: fixed

client/relay-chain-interface/src/lib.rs
```diff
@@ -110,6 +111,9 @@ pub trait RelayChainInterface: Send + Sync {
 	/// Get the hash of the current best block.
 	async fn best_block_hash(&self) -> RelayChainResult<PHash>;
 
+	/// Fetch the block header of a given height
+	async fn header(&self, block_id: BlockId) -> RelayChainResult<Option<PHeader>>;
```
bkchr, line 115: Please use `PHash`. No new `BlockId` in public interfaces anymore.

seunlanlege: But i specifically need to query using `BlockNumber`, any reason for this? 

> No new BlockId in public interfaces anymore

bkchr: Block numbers should only be used in places where there is no other way around them. Block numbers are not ambiguous. Some interfaces (like RPC/CLI) still require block numbers, but most "internal code" can be rewritten in a way to use hashes. E.g. get the header of block X and then go back using `parent_hash()`.

seunlanlege: > But i specifically need to query using BlockNumber, any reason for this?

The use case here is on-chain code which wishes to read the relay chain, they signal a request given a specific relay chain height (parachains don't have access to the relay chain blockhashes). So hence the `BlockId` here. 

Doesn't seem like there's any alternatives here

bkchr: And how does the parachain validates the relay chain headers?

bkchr: > The use case here is on-chain code which wishes to read the relay chain, they signal a request given a specific relay chain height (parachains don't have access to the relay chain blockhashes).

You say it signals which relay chain header it wants to read. So you are just wanting to get the header to get the storage root?

seunlanlege: yes

bkchr: Okay, when we got: https://github.com/paritytech/cumulus/issues/303

This can be changed.

bkchr (approved):

rphmeier:
This shouldn't have been merged as-is - the interface was changed to `Option<Header>` but the implementations still wrapped it in opaque errors for `BlockNumber`

bkchr:
By "opaque error" you mean that it is a string error?

seunlanlege:
Will do a follow up

seunlanlege:
follow up: https://github.com/paritytech/cumulus/pull/2830

PR #2800, opened 2023-06-30
Fix for `xcm-emulator` - Wrong Parachain processing message

NachoPal:
I found out that the condition to process messages by a Parachain is incomplete.

We did not notice because of the coincidence that for all our tests, the destination Parachain that is supposed to receive/process the message, had been declared always before the other Parachains.

Example of failing case:
We send a message to `PenpalPolkadot`. The message will never reach its destination because `AssetHubPolkadot` will try to process it first as it is a message to the same `Network` `AssetHubPolkadot` and `PenpalPolkadot` live.
```rust
		parachains = vec![
			AssetHubPolkadot,
			PenpalPolkadot,
		],
```
Same test would pass just switching Parachains order declarations.
```rust
		parachains = vec![
			PenpalPolkadot,
			AssetHubPolkadot,
		],
```

Solution: filter also by `para_id` and not only by `Network`.

### DIFF
--- a/xcm/xcm-emulator/src/lib.rs
+++ b/xcm/xcm-emulator/src/lib.rs
@@ -780,6 +780,8 @@ macro_rules! decl_test_networks {
 					while let Some((to_para_id, messages))
 						= $crate::DOWNWARD_MESSAGES.with(|b| b.borrow_mut().get_mut(stringify!($name)).unwrap().pop_front()) {
 						$(
+							let para_id: u32 = <$parachain>::para_id().into();
+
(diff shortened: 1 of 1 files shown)


gilescope (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3097485

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3097741

muharem (approved):

NachoPal:
bot merge

PR #2812, opened 2023-07-03
Add Bridges to `xcm-emulator`

NachoPal:
Adding bridging capabilities to `xcm-emulator`.

The `bridge` that a `Network` is expecting is any type that implements the `BridgeMessageHandler` trait. In this way we decouple `xcm-emulator` from the particular Bridge implementation for `BridgeHub`, avoiding also the need of importing all related packages (`bp_messages`, `pallet_bridge_messages`, etc). In addition, this approach would allow other teams to add their own Bridge implementations to the `Network`.

The `BridgeHub` transport layer has been mocked, abstracting out all the unnecessary complexity related to Relayers and proofs.

In the PR have been added other improvements:
- More logs (`dmp`, `ump` and `hrmp`)
- `Network` trait refactoring
- `decl_test_sender_receiver_accounts_parameter_types` macro


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3108012

NachoPal:
bot fmt

NachoPal:
bot fmt

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3108864

bridges/primitives/messages/src/lib.rs
```diff
@@ -102,6 +102,18 @@ impl TypeId for LaneId {
 	const TYPE_ID: [u8; 4] = *b"blan";
 }
 
+impl From<LaneId> for u32 {
```
bkontur, line 105: @NachoPal @svyatonik 
I am not sure about this, can we move this out of `bridges` substree?
Because then we need to backport it back to bridges repo and also in new version LaneId has a different type: https://github.com/paritytech/parity-bridges-common/pull/2221/files#diff-836b204328629e6af4b33c2a34ae41191cab06cf89c066e31bdb7085314fcda1L185-R203

xcm/xcm-emulator/src/lib.rs
```diff
@@ -190,6 +193,63 @@ pub trait Parachain: XcmpMessageHandler + DmpMessageHandler {
...
+
+impl BridgeMessageHandler for () {
+	fn get_source_outbound_messages() -> Vec<BridgeMessage> {
+		Default::default()
+	}
+
+	fn dispatch_target_inbound_message(
+		_message: BridgeMessage,
+	) -> Result<(), BridgeMessageDispatchError> {
+		Err(BridgeMessageDispatchError(Box::new("Not a brige")))
```
gilescope, line 236:
```suggestion
		Err(BridgeMessageDispatchError(Box::new("Not a bridge")))
```

bridges/primitives/messages/src/lib.rs
gilescope: Agreed there's probably a better way to do this.

gilescope (approved):

bridges/primitives/messages/src/lib.rs
NachoPal: The problem is implementing `From` for `LaneId` or the fact that `LaneId` is going to change? I do not think is harmful to implement `From` for a new type.

I could try to locally (in `xcm-emulator`) wrap `LaneId` in a new type and implement the `From` trait for it.

NachoPal:
bot fmt

bridges/primitives/messages/src/lib.rs
NachoPal: > I could try to locally (as part of the BridgeHubMessageHandler impl) wrap LaneId in a new type and implement the From trait for it.

@bkontur @gilescope Check it out

NachoPal:
bot merge

PR #2826, opened 2023-07-05
Allow polkadot-parachain bin to be a dep

gilescope:
(for example integration tests currently duplicate genesis information because it can't pull this info in as a dep)

### DIFF
--- a/polkadot-parachain/src/lib.rs
+++ b/polkadot-parachain/src/lib.rs
@@ -26,6 +26,7 @@ mod cli;
 mod command;
 mod rpc;
 
-fn main() -> sc_cli::Result<()> {
+/// Process command line arguments and execute them.
+pub fn run() -> sc_cli::Result<()> {
 	command::run()
 }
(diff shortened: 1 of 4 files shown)


polkadot-parachain/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+
+// Cumulus is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Cumulus.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Cumulus test parachain collator
```
joepetrowski, line 17: Why test?

gilescope, line 17:
```suggestion
//! Cumulus parachain node (which could be run as a collator)
```

gilescope:
There's a lot of logic in that crate that others may wish to make use of rather than having to replicate. It's rarely helpful to have a lot of logic in a rust binary, best to hand off to a library asap because at some point someone wants to reuse that logic.

PR #1156, opened 2022-04-07
Co #11164: Sub-commands for `benchmark`

ggwpez:
Changes in the `parachain-template` and `collator`:  
- :point_right: **Breaking**: Move the pallet benchmarking to `benchmark pallet`
- Add `benchmark storage` and `benchmark block`
- Add some tests for these

This does not include `benchmark overhead`, as it would take me longer to figure out the inherents and signed extension stuff.  
The "breaks api" tag is here since it changes the client commands.

Companion for https://github.com/paritytech/substrate/pull/11164 and https://github.com/paritytech/polkadot/pull/5247

PR #2858, opened 2023-07-12
Add overhead benchmark to parachain-template

kalaninja:
Adds `benchmark overhead` that was missing in the previous PR #1156


parachain-template/node/src/benchmarking.rs
```diff
@@ -0,0 +1,133 @@
+//! Setup code for [`super::command`] which would otherwise bloat that module.
```
ggwpez, line 1: License?

ggwpez (commented):
Is there a test to check that the new command works?  
Fot the `storage` benchmark there seems to be one for the asset hubs: https://github.com/paritytech/cumulus/blob/80aaed3022b35540b39a673ce5beef94abbb1961/polkadot-parachain/tests/benchmark_storage_works.rs#L34

parachain-template/node/src/benchmarking.rs
bkchr: The template has no license.

kalaninja:
> Fot the `storage` benchmark there seems to be one for the asset hubs:

This test is from `polkadot-parachain`. I decided to split this into 2 PRs. So the overhead benchmark and the test for collateral will come in a different PR.

> Is there a test to check that the new command works?

I tested this one manually.

kalaninja:
@ggwpez the test was added in #2900

bkontur:
@kalaninja could we run benchbot here to see that it works? maybe for parachain-template it is not configured

ggwpez:
> @kalaninja could we run benchbot here to see that it works? maybe for parachain-template it is not configured

We can run it for the asset hubs over here once its merged and the bench bot is updated: https://github.com/paritytech/cumulus/pull/2900. 
The Bot does not care about templates, since they dont need correct weights.

parachain-template/node/src/command.rs
```diff
@@ -181,25 +184,20 @@ pub fn run() -> Result<()> {
...
 			match cmd {
-				BenchmarkCmd::Pallet(cmd) =>
-					if cfg!(feature = "runtime-benchmarks") {
-						runner.sync_run(|config| cmd.run::<Block, ()>(config))
-					} else {
-						Err("Benchmarking wasn't enabled when building the node. \
+				#[cfg(not(feature = "runtime-benchmarks"))]
+				BenchmarkCmd::Pallet(_) => Err("Benchmarking wasn't enabled when building the node. \
 					You can enable it with `--features runtime-benchmarks`."
-							.into())
```
ggwpez, line 190: Also here the Question; why not use the `cfg!` to make it look more compact instead of the `cfg(not`...?

ggwpez (approved):
LGTM (tested it locally).

We could still add an integration test as good measure to ensure that we dont break it in the future, but its also fine like this 😄

parachain-template/node/src/command.rs
kalaninja: the only reason is that it is done with attributes in all of the other places.  So, just didn't want to mix multiple approaches.

PR #2875, opened 2023-07-15
Rename squatted crates

Morganamilo:
This commit adds the staging- prefix to squatted crates so we can go forward and publish them to crates.io.

Using the staging- prefix is a temp fix until we decide on replacement names. https://forum.parity.io/t/renaming-squated-crates-in-substrate-polkadot-cumulus/1964/6


bkchr (changes requested):
Imo we should just remove the crate. It doesn't bring any benefit

Morganamilo:
A bunch of other crates depend on it. So removing it causes a bunch of compile issues. I'd rather keep this PR to just a simple rename and if some one want's to remove the crate later on they can.

gilescope:
Shouldn't it be called `pallet-parachain-info` in line with all the other pallets in substrate? That would remove the squatting problem.

ggwpez:
> Shouldn't it be called `pallet-parachain-info` in line with all the other pallets in substrate? That would remove the squatting problem.

Many of them are also prefixed with `cumulus-pallet-`. Going to reserve both (just in case).

ggwpez (approved):

sam0x17 (approved):

Morganamilo:
@bkchr can you approve. This PR is to just get things working. Any cleanup or actual changes like removing a crate can be handled after.

bkchr (approved):

PR #2893, opened 2023-07-18
Xcm-Emulator: use `ext_wrapper` to execute with externalities

muharem:
The processing of an XCM message should not trigger the network to process all the XCM channels, which happens if execute with `execute_with`.
The `ext_wrapper` will wrap the execution only with test externalities.

### DIFF
--- a/xcm/xcm-emulator/src/lib.rs
+++ b/xcm/xcm-emulator/src/lib.rs
@@ -319,7 +319,7 @@ macro_rules! decl_test_relay_chains {
 					use $mq as message_queue;
 					use $runtime_event as runtime_event;
 
-					Self::execute_with(|| {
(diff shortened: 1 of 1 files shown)


NachoPal (approved):

bkontur (approved):
@NachoPal @muharem 
can you guys, please, add some docs/description to the:
```
        fn execute_with<R>(execute: impl FnOnce() -> R) -> R;
	fn ext_wrapper<R>(func: impl FnOnce() -> R) -> R;
```

they have the same signature but would be good to know that is the difference

muharem:
@bkontur there is a branch with that from @NachoPal, coming soon

muharem:
bot merge

PR #2900, opened 2023-07-19
Add overhead benchmark to polkadot-parachain

kalaninja:
- add `benchmark overhead` that was missing in the previous PR #1156 
- add `benchmark_overhead_works` test
- add "collectives-polkadot" runtime to `benchmark_storage_works`


polkadot-parachain/tests/benchmark_storage_works.rs
```diff
@@ -8,7 +8,8 @@ use std::{
 use tempfile::tempdir;
 
 /// The runtimes that this command supports.
-static RUNTIMES: [&str; 3] = ["asset-hub-westend", "asset-hub-kusama", "asset-hub-polkadot"];
+static RUNTIMES: [&str; 4] =
+	["asset-hub-westend", "asset-hub-kusama", "asset-hub-polkadot", "collectives-polkadot"];
```
bkontur, line 12: @kalaninja 
why `bridge-hub-*` runtimes are excluded? Because of different `let extra: runtime::SignedExtra` extentions?

what is the purpose or usage of `overhead`?
I just want to understand if we need to add support also for BridgeHub runtimes

kalaninja: > what is the purpose or usage of overhead?
they generate base weights for a block and extrinsic. 

> why bridge-hub-* runtimes are excluded?
The reason why I excluded these is that they are excluded in [construct_benchmark_partials](https://github.com/paritytech/cumulus/blob/b894b3d224c64d46bca93f61f34e1daf9e61738d/polkadot-parachain/src/command.rs#L38) macro.

> Because of different let extra: runtime::SignedExtra extentions?
Included runtimes have different SignedExtra too. So, this shouldn't be the reason.

bkontur: @kalaninja 
aha, ok, thank you, 
so let me add them to the `construct_benchmark_partials`

bkontur: @kalaninja 
ok, I added BridgeHubs [here](https://github.com/paritytech/cumulus/pull/2901),
it is merged in master, please, rebase this

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3218984

polkadot-parachain/tests/benchmark_storage_works.rs
kalaninja: Thanks @bkontur
I added benchmarks for BridgeHubs. Please, take a look.

polkadot-parachain/src/benchmarking.rs
```diff
@@ -0,0 +1,525 @@
...
+			parachains_common::Signature::Sr25519(signature),
+			extra,
+		);
+
+		Ok(extrinsic.into())
+	}
+}
+
+impl frame_benchmarking_cli::ExtrinsicBuilder
+	for RemarkBuilder<bridge_hub_kusama_runtime::RuntimeApi>
```
ggwpez, line 368: Its quite a lot of copy&paste...

Is it only the name or also the signed extensions?  
Maybe you can write a macro to expand it for all runtimes, otherwise it will be impossible to change later on.

polkadot-parachain/src/command.rs
```diff
@@ -695,24 +696,19 @@ pub fn run() -> Result<()> {
 
 			// Switch on the concrete benchmark sub-command-
 			match cmd {
+				#[cfg(not(feature = "runtime-benchmarks"))]
+				BenchmarkCmd::Pallet(_) => Err("Benchmarking wasn't enabled when building the node. \
+					You can enable it with `--features runtime-benchmarks`."
+					.into()),
+				#[cfg(feature = "runtime-benchmarks")]
 				BenchmarkCmd::Pallet(cmd) =>
-					if cfg!(feature = "runtime-benchmarks") {
```
ggwpez, line 699: I think its a bit easier to read with the `cfg!` macro. Why the change?

polkadot-parachain/tests/benchmark_overhead_works.rs
```diff
@@ -0,0 +1,42 @@
...
+		let runtime = format!("{}-dev", runtime);
+
+		let status = Command::new(cargo_bin("polkadot-parachain"))
+			.args(["benchmark", "overhead", "--chain", &runtime, "-d"])
+			.arg(base_path)
+			.arg("--weight-path")
+			.arg(base_path)
+			.args(["--warmup", "10", "--repeat", "10"])
+			.args(["--add", "100", "--mul", "1.2", "--metric", "p75"])
+			.args(["--max-ext-per-block", "10"])
```
ggwpez, line 33: Depending on CI timing we can reduce these numbers a bit here and for `warmup` and `repeat` but it should be rather fast.  
The slowest part is actually building the block.

polkadot-parachain/src/command.rs
kalaninja: I don't see any big difference. The only reason why I changed this is that it is done with attributes in all other places. So, just didn't want to mix multiple approaches.

polkadot-parachain/src/benchmarking.rs
kalaninja: I decoupled `ExtrinsicBuilder` implementation from signed extensions for each runtime with `SignedExtraProvider` trait. Now `RemarkBuilder` has a single generic implementation and a corresponding `SignedExtraProvider` is injected into its type constructor.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3240827

polkadot-parachain/src/command.rs
ggwpez: I dont have a strong opinion about it - just wanted to comment.

polkadot-parachain/src/benchmarking.rs
ggwpez: Very nice. I did not think about this before - should probably be upstreamed eventually. I will create an issue.

ggwpez (approved):
LGTM besides the failing test with "Bad input data provided to apply_extrinsic: Codec error" so something with the signed extensions is maybe wrong? (Good that you added the tests 😆).

kalaninja:
@ggwpez Fixed. Good that among two of at least one knows that tests should be run after refactoring. :laughing:

PR #2911, opened 2023-07-19
Feature Flagging Consensus Hook Type Parameter

BradleyOlson64:
This is a first try for me using feature flags. Hopefully somewhat correct :)

### DIFF
--- a/pallets/parachain-system/Cargo.toml
+++ b/pallets/parachain-system/Cargo.toml
@@ -77,3 +77,5 @@ runtime-benchmarks = [
 ]
 
 try-runtime = ["frame-support/try-runtime"]
+
+parameterized-consensus-hook = []
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -212,6 +212,8 @@ pub mod pallet {
 		/// [`consensus_hook::ExpectParentIncluded`] here. This is only necessary in the case
 		/// that collators aren't expected to have node versions that supply the included block
 		/// in the relay-chain state proof.
+        ///
+        /// This config type is only available when the `parameterized-consensus-hook` crate feature is activated.
 		#[cfg(feature = "parameterized-consensus-hook")]
 		type ConsensusHook: ConsensusHook;
 	}
--- a/pallets/parachain-system/src/lib.rs
+++ b/pallets/parachain-system/src/lib.rs
@@ -492,11 +492,9 @@ pub mod pallet {
 
 			// Update the desired maximum capacity according to the consensus hook.
 			#[cfg(feature = "parameterized-consensus-hook")]
-			let (consensus_hook_weight, capacity) =
-				T::ConsensusHook::on_state_proof(&relay_state_proof);
+			let (consensus_hook_weight, capacity) = T::ConsensusHook::on_state_proof(&relay_state_proof);
 			#[cfg(not(feature = "parameterized-consensus-hook"))]
-			let (consensus_hook_weight, capacity) =
-				ExpectParentIncluded::on_state_proof(&relay_state_proof);
+			let (consensus_hook_weight, capacity) = ExpectParentIncluded::on_state_proof(&relay_state_proof);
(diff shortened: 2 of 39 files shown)


parachain-template/runtime/src/lib.rs
```diff
@@ -225,11 +225,14 @@ const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
 
 /// Maximum number of blocks simultaneously accepted by the Runtime, not yet included
 /// into the relay chain.
+#[cfg(feature = "parameterized-consensus-hook")]
```
rphmeier, line 228: we don't actually need feature-flags in the parachain-template, as it's meant to be example code. So it should use the most recently available logic.

parachains/runtimes/assets/asset-hub-kusama/src/lib.rs
```diff
@@ -47,7 +47,9 @@ use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 
 use codec::{Decode, Encode, MaxEncodedLen};
-use constants::{consensus::*, currency::*, fee::WeightToFee};
+#[cfg(feature = "parameterized-consensus-hook")]
```
rphmeier, line 50: we don't need the feature flag in system-parachain runtimes either, as they're binaries, not libraries.

test/runtime/src/lib.rs
```diff
@@ -286,6 +286,7 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
 	type XcmpMessageHandler = ();
 	type ReservedXcmpWeight = ();
 	type CheckAssociatedRelayNumber = cumulus_pallet_parachain_system::AnyRelayNumber;
+	#[cfg(feature = "parameterized-consensus-hook")]
```
rphmeier, line 289: likewise here - it seems fine for the test runtime (which is effectively a binary crate used internally, not a library) to explicitly use the `RequireParentIncluded`

BradleyOlson64: Which way is best to get this to compile without the feature flag here? 

Screenshot 2023-07-19 at 5 02 57 PM

I expect some change is needed around line 215 of `pallets/parachain-system/src/lib.rs` such that it can still accept a `ConsensusHook` type even without the feature flag enabled. Just haven't seen an example of what that would look like.

rphmeier: Can we just have it unconditionally enable the crate feature in its `Cargo.toml`?

BradleyOlson64: That does make sense 👍 Will be back on to try it this evening.

slumber: Robert probably meant doing
```toml
cumulus-pallet-parachain-system = { path = "../../../../pallets/parachain-system", default-features = false, features = ["parameterized-consensus-hook"] }
```
instead of introducing a feature to test-crates.

pallets/parachain-system/src/lib.rs
```diff
@@ -212,6 +212,7 @@ pub mod pallet {
 		/// [`consensus_hook::ExpectParentIncluded`] here. This is only necessary in the case
 		/// that collators aren't expected to have node versions that supply the included block
 		/// in the relay-chain state proof.
+		#[cfg(feature = "parameterized-consensus-hook")]
```
rphmeier, line 215:
```suggestion
        ///
        /// This config type is only available when the `parameterized-consensus-hook` crate feature is activated.
		#[cfg(feature = "parameterized-consensus-hook")]
```

rphmeier (approved):
Looks good! This looks properly backwards-compatible.

pallets/parachain-system/Cargo.toml
```diff
@@ -77,3 +77,5 @@ runtime-benchmarks = [
 ]
 
 try-runtime = ["frame-support/try-runtime"]
+
+parameterized-consensus-hook = []
```
slumber, line 81: missing endlines though

PR #2914, opened 2023-07-21
Remove collators from bootnodes list

BulatSaif:
We are no longer using collators in the bootnode list; instead, we have replaced them with "connect" nodes. This change serves two purposes: first, it removes unnecessary load from collators, and secondly, it enables us to proceed with the planned decommission of collators nodes as outlined in [Referendum 84](https://polkadot.subsquare.io/referenda/referendum/84)

### DIFF
--- a/parachains/chain-specs/bridge-hub-kusama.json
+++ b/parachains/chain-specs/bridge-hub-kusama.json
@@ -3,10 +3,6 @@
   "id": "bridge-hub-kusama",
   "chainType": "Live",
   "bootNodes": [
-    "/dns/kusama-bridge-hub-collator-ew1-0.polkadot.io/tcp/30334/p2p/12D3KooWP2Gngt4tt2sz5BgDaAbMTxasPWk3V2Z99bQTmFcAorqa",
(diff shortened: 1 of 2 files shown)


parachains/chain-specs/bridge-hub-kusama.json
joepetrowski, line 1: This is Kusama, but the referendum is for Polkadot. Did you mean to make this change to `bridge-hub-polkadot` (instead of in addition)?

BulatSaif: Collators from bridge-hub-polkadot were already removed (I believe I did it before the release). This cleanup is for the remaining parachains.

BulatSaif: Note: For testnet parachains, we still use collators as bootnodes since the load is smaller, and it doesn't make financial sense to spin up additional nodes just for bootnodes.

joepetrowski (approved):

bkchr (approved):

bkchr:
bot merge

BulatSaif:
> Error: Statuses failed for [c86fa1d](https://github.com/paritytech/cumulus/commit/c86fa1d4ee7819c5e1212226f5fe760a6da99a82)

Rule "Core developers" needs at least 2 approvals, but 1 were given

bkontur (approved):

BulatSaif:
bot merge

PR #2925, opened 2023-07-24
improve deps for xcm-emulator

xlc:
So that it doesn't need to pulling cumulus-test-service, which have huge dependencies.

I noticed this when updating Acala and noticed I am building pallet-glutton which I shouldn't.

This is also one of the that I am against mono repo as it is too easy for people to pulling deps from other crates without proper thinking the dependency tree.

There are other offenders such as cumulus-pallet-xcmp-queue depends polkadot-runtime-common which itself is huge. But the proper fix requires a larger refactoring.


gilescope (approved):
Totally agree. I'm sure I have the same change on one of my local branches.

gilescope:
bot fmt

ggwpez (approved):
Yea looks like a good trade-off.

gilescope:
bot merge

PR #2949, opened 2023-07-28
Explicitly set `full-node` feature for emulated tests

gilescope:
Was implicitly turned on before by a dep that has been recently removed by #2925 .

(without this if you go to `parachains/integration-tests/emulated/assets/asset-hub-westend` and run `cargo test` it will lead to compile errors.)


PatricioNapoli (approved):

gilescope:
bot merge

PR #2977, opened 2023-08-07
(async backing) restore `CheckInherents` for backwards-compatibility

slumber:
re https://github.com/paritytech/cumulus/pull/2300#discussion_r1285301409

The `CheckInherents` parameter is back to `register_validate_block!`. It's empty in all cumulus runtimes since we moved the timestamp check to aura consensus hook and didn't have any other usages.

Warn users about this trait being deprecated if the consensus hook feature is enabled.


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3341945

pallets/parachain-system/src/lib.rs
```diff
@@ -1497,6 +1497,33 @@ impl<T: Config> UpwardMessageSender for Pallet<T> {
 	}
 }
 
+/// Something that can check the inherents of a block.
+#[deprecated = "use `cumulus-pallet-aura-ext::FixedVelocityConsensusHook` instead"]
```
rphmeier, line 1501:
```suggestion
#[deprecated = "use `cumulus-pallet-aura-ext::FixedVelocityConsensusHook` instead"]
```

A couple nits:
  1. `type ConsensusHook` in parachain-system currently lives under a feature flag - therefore we are deprecating and asking users to enable a feature flag simultaneously. I'd recommend just adding a comment here for now rather than a deprecation notice.
  2. The comment should be general and not point directly to `aura_ext`.

pallets/parachain-system/proc-macro/src/lib.rs
```diff
@@ -101,6 +102,28 @@ pub fn register_validate_block(input: proc_macro::TokenStream) -> proc_macro::To
...
 	};
 
+	let check_inherents = match check_inherents {
+		Some(_check_inherents) => {
+			#[cfg(not(feature = "parameterized-consensus-hook"))]
+			quote::quote! { #_check_inherents }
+			#[cfg(feature = "parameterized-consensus-hook")]
+			return Error::new(
+				Span::call_site(),
+				"CheckInherents is incompatible with consensus hook feature",
```
rphmeier, line 112: I like this approach to giving a compiler error when trying to use both `CheckInherents` and `ConsensusHook`.

pallets/parachain-system/src/lib.rs
rphmeier:
```suggestion
#[cfg_attr(feature = "parameterized-consensus-hook", deprecated = "Perform these checks using a `ConsensusHook` instead"]
```

Something like this?

pallets/parachain-system/proc-macro/src/lib.rs
```diff
@@ -31,7 +31,7 @@ mod keywords {
 struct Input {
 	runtime: Path,
 	block_executor: Path,
-	check_inherents: Option<Path>,
+	check_inherents: Path,
```
rphmeier, line 34: Leaving this as an `Option` seems fine, as it'd be backwards compatible. It's just that if the `Option` is `None` we'd want to drop in an empty `CheckInherents` implementation (to avoid all runtimes having to implement one, especially if it's deprecated)

rphmeier (approved):

PR #2978, opened 2023-08-07
Set test crates to nopublish and organise metadata

Morganamilo:


coderobe (approved):

bridges/bin/runtime-common/Cargo.toml
```diff
@@ -1,10 +1,10 @@
 [package]
 name = "bridge-runtime-common"
 version = "0.1.0"
-authors = ["Parity Technologies <admin@parity.io>"]
-edition = "2021"
-repository = "https://github.com/paritytech/parity-bridges-common/"
-license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
```
ggwpez, line 7: I am not sure if we can change the license like this. The classpath exception can be quite important for downstream projects.

bkchr: Node side with should be GPL with classpath exception and runtime Apache 2.0.

PR #2979, opened 2023-08-07
Symlink chain-specs json files to crate where they are used

Morganamilo:
When publishing crates, each crate becomes it's own tarball that can't access files from other crates. So symlink the files to be crate local and cargo will replace the symlinks with real files at publish time.


bkchr:
We should just move the `chain-specs` directory directly. No need to have it symlinked.

PR #2980, opened 2023-08-07
init para ids when reseting network

samelamin:
I noticed that resetting the network in tests fails because PARA_IDs are not initialized. This is a small fix for that

### DIFF
--- a/parachains/integration-tests/emulated/assets/asset-hub-kusama/src/tests/reserve_transfer.rs
+++ b/parachains/integration-tests/emulated/assets/asset-hub-kusama/src/tests/reserve_transfer.rs
@@ -18,6 +18,7 @@ use crate::*;
 
 #[test]
 fn reserve_transfer_native_asset_from_relay_to_assets() {
+	KusamaMockNet::reset();
 	// Init tests variables
 	let amount = KUSAMA_ED * 1000;
 	let relay_sender_balance_before = Kusama::account_data_of(KusamaSender::get()).free;
--- a/parachains/integration-tests/emulated/assets/asset-hub-kusama/src/tests/reserve_transfer.rs
+++ b/parachains/integration-tests/emulated/assets/asset-hub-kusama/src/tests/reserve_transfer.rs
@@ -18,7 +18,6 @@ use crate::*;
 
 #[test]
 fn reserve_transfer_native_asset_from_relay_to_assets() {
-	KusamaMockNet::reset();
 	// Init tests variables
 	let amount = KUSAMA_ED * 1000;
 	let relay_sender_balance_before = Kusama::account_data_of(KusamaSender::get()).free;
--- a/xcm/xcm-emulator/src/lib.rs
+++ b/xcm/xcm-emulator/src/lib.rs
@@ -809,7 +809,7 @@ macro_rules! decl_test_networks {
 
 			impl $name {
 				pub fn reset() {
-					use $crate::{TestExt, VecDeque};
+					use $crate::{TestExt, VecDeque, Network};
 
 					$crate::INITIALIZED.with(|b| b.borrow_mut().remove(stringify!($name)));
 					$crate::DOWNWARD_MESSAGES.with(|b| b.borrow_mut().remove(stringify!($name)));
@@ -818,6 +818,7 @@ macro_rules! decl_test_networks {
 					$crate::HORIZONTAL_MESSAGES.with(|b| b.borrow_mut().remove(stringify!($name)));
 					$crate::BRIDGED_MESSAGES.with(|b| b.borrow_mut().remove(stringify!($name)));
 					$crate::RELAY_BLOCK_NUMBER.with(|b| b.borrow_mut().remove(stringify!($name)));
+					$crate::PARA_IDS.with(|b| b.borrow_mut().insert(stringify!($name).to_string(), Self::_para_ids()));
 
 					<$relay_chain>::reset_ext();
 					$( <$parachain>::reset_ext(); )*
--- a/xcm/xcm-emulator/src/lib.rs
+++ b/xcm/xcm-emulator/src/lib.rs
@@ -818,7 +818,7 @@ macro_rules! decl_test_networks {
 					$crate::HORIZONTAL_MESSAGES.with(|b| b.borrow_mut().remove(stringify!($name)));
 					$crate::BRIDGED_MESSAGES.with(|b| b.borrow_mut().remove(stringify!($name)));
 					$crate::RELAY_BLOCK_NUMBER.with(|b| b.borrow_mut().remove(stringify!($name)));
-					$crate::PARA_IDS.with(|b| b.borrow_mut().insert(stringify!($name).to_string(), Self::para_ids()));
+					$crate::PARA_IDS.with(|b| b.borrow_mut().remove(stringify!($name)));
 
 					<$relay_chain>::reset_ext();
 					$( <$parachain>::reset_ext(); )*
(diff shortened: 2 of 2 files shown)


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3342910

gilescope (approved):
Good spot.

NachoPal (approved):

NachoPal:
@samelamin Thanks for spotting it, but I think just adding:

`$crate::PARA_IDS.with(|b| b.borrow_mut().remove(Self::name()));` to `reset()` would be enough.

samelamin:
> $crate::PARA_IDS.with(|b| b.borrow_mut().remove(Self::name()));

thanks @NachoPal , perhaps you meant `$crate::PARA_IDS.with(|b| b.borrow_mut().remove(stringify!($name)));`

because with `Self::name()` we get the below error
```
error[E0599]: no function or associated item named `name` found for struct `PolkadotMockNet` in the current scope
   --> parachains/integration-tests/emulated/common/src/lib.rs:403:1
   ```

**Comment by NachoPal** [comment]
> > $crate::PARA_IDS.with(|b| b.borrow_mut().remove(Self::name()));
> 
> thanks @NachoPal , perhaps you meant `$crate::PARA_IDS.with(|b| b.borrow_mut().remove(stringify!($name)));`
> 
> because with `Self::name()` we get the below error
> 
> ```
> error[E0599]: no function or associated item named `name` found for struct `PolkadotMockNet` in the current scope
>    --> parachains/integration-tests/emulated/common/src/lib.rs:403:1
> ```

Yes, sorry, I copied that from another branch that is about to be merged.

As you said, it would be:
```
$crate::PARA_IDS.with(|b| b.borrow_mut().remove(stringify!($name)));
```

File: xcm/xcm-emulator/src/lib.rs [hunk 1]
```diff
@@ -809,7 +809,7 @@ macro_rules! decl_test_networks {
 
 			impl $name {
 				pub fn reset() {
-					use $crate::{TestExt, VecDeque};
+					use $crate::{TestExt, VecDeque, Network};
```
NachoPal, line 812: I think `Network` can be removed from here.

parachains/integration-tests/emulated/assets/asset-hub-kusama/src/tests/reserve_transfer.rs
```diff
@@ -18,6 +18,7 @@ use crate::*;
 
 #[test]
 fn reserve_transfer_native_asset_from_relay_to_assets() {
+	KusamaMockNet::reset();
```
NachoPal, line 21: This to me removed also

NachoPal (changes requested):

parachains/integration-tests/emulated/assets/asset-hub-kusama/src/tests/reserve_transfer.rs
samelamin: I can remove, but wouldn't it make sense to reset the network at the start of the test to ensure the test is encapsulated and not affected by states from other tests?

NachoPal: There is no need of reseting the state in that case. The state is per thread, and each test is run on its own thread. You do not have to worry about sharing state between tests.

`reset()` it is only useful if you want to reset the state inside a test.

I am wondering now what's actually the issue you encountered when trying to `reset()`. It is true that `PARA_IDS` where not reset, but I think that was left like that because it shouldn't change once the Network is initialised the first time. `PARA_IDS` should be constant.

samelamin: it was failing on `0.9.43` because when trying to unwrap
for example

```
 $crate::DOWNWARD_MESSAGES.with(|b| b.borrow_mut().get_mut(stringify!($name)).unwrap().pop_front()) {
```

However, with the latest changes on main I am not sure if it applies. I will close if you feel it isnt required

NachoPal: How `DOWNWARD_MESSAGES` and `PARA_IDS` are related? 

This ` $crate::DOWNWARD_MESSAGES.with(|b| b.borrow_mut().get_mut(stringify!($name)).unwrap().pop_front())` shouldn't fail as it is supposed to be initialized on `init()`.

If it is panicking in the `unwrap()` then it means that either there is a bug or you are doing something wrong. We shouldn't close the issue until this is clarified.

samelamin: > How `DOWNWARD_MESSAGES` and `PARA_IDS` are related?
> 
> ` $crate::DOWNWARD_MESSAGES.with(|b| b.borrow_mut().get_mut(stringify!($name)).unwrap().pop_front())` shouldn't fail as it is supposed to be initialized on `init()`.
> 
> If it is panicking in the `unwrap()` then it means that either there is a bug or you are doing something wrong. We shouldn't close the issue until this is clarified.

Apologies  @NachoPal i referenced the wrong part of the code, its actually [below](https://github.com/paritytech/cumulus/blob/master/xcm/xcm-emulator/src/lib.rs#L972-L975)

```
			for recipient_para_id in $crate::PARA_IDS.with(|b| b.borrow_mut().get_mut(stringify!($name)).unwrap().clone()) {
						let recipient_para_id = $crate::ParaId::from(recipient_para_id);
						if let Err(idx) = e_index.binary_search(&recipient_para_id) {
							e_index.insert(idx, recipient_para_id);
						}
```

When I attempted to reset the network the `unwrap` panicked

I am not able to replicate on main on cumulus though so thats why I am unsure if it still applies. If you switch the `polkadot-v0.9.43` then you will be able to replicate when you run reset on a test

NachoPal: Can you share a link to you failing test?

I think I know what's happening. I thing that the first thing you are doing is calling `YourChain::reset()`. In `polkadot-v0.9.43`, when `reset()`, [this is called](https://github.com/paritytech/cumulus/blob/polkadot-v0.9.43/xcm/xcm-emulator/src/lib.rs#L727) that ends up calling [this](https://github.com/paritytech/cumulus/blob/polkadot-v0.9.43/xcm/xcm-emulator/src/lib.rs#L842).

Because no previous externality of you parachain exists, `init()` has never being called, and there `PARA_IDS` do not exist yet.

In `master`, [it was moved to](https://github.com/paritytech/cumulus/blob/master/xcm/xcm-emulator/src/lib.rs#L842) `init()`. That's why it is not failing.

As I said, it doesn't make sense to call `reset()` in a test if the the chain's externality was never initialised yet, there is not state to reset. You can try to initialise the state first, and then call `reset()` and it should fail as `PARA_IDS` already exist.

NachoPal:
@samelamin Can it be closed?

parachains/integration-tests/emulated/assets/asset-hub-kusama/src/tests/reserve_transfer.rs
samelamin: sorry for the late reply @NachoPal 

this makes sense, thanks for clarifying. 

I didn't have a failing test, I just noticed that adding reset was causing a failure while I was writing a few more tests for our project

This means this issue can be closed. Thank again Nacho

PR #2994, opened 2023-08-11
Girazoki make emulator functions more generic

girazoki:
Makes emulator a bit more generic with respect to the location to account id and account data. Prior to this PR always the accountID32 type was being used, which made the emulator unusable with 20 byte accounts like Moonbeam. This PR makes the macro compatible with these except for one single case: the `account_id_of` of the Parachain impl. The main reason I did not want to touch this function is because:

- 20 byte account generation probably has to be done with ecdsa and it would cause a lot of conflicts
- It is not critical as currently its mostly used to declare senders and receivers, which we can already do through `parameter_types`

### DIFF
--- a/xcm/xcm-emulator/src/lib.rs
+++ b/xcm/xcm-emulator/src/lib.rs
@@ -455,19 +455,19 @@ macro_rules! __impl_relay {
 				(Ancestor(0), Parachain(id.into())).into()
 			}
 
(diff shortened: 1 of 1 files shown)


girazoki:
NVM I see the emulator changed a lot lately, sowill need to adapt this pr to that

PR #2762, opened 2023-06-20
Transfer asset via bridge with pallet xcm

bkontur:

PR #2997, opened 2023-08-11
Transfer asset via bridge pallet xcm with dynamic fees and back-pressure

bkontur:
This PR extends [asset transfer with `pallet_xcm` PR](https://github.com/paritytech/cumulus/pull/2762), adding **dynamic fees** and **back-preassure** mechanism.

You can review `bridges` subdirectory either here or in its dedicated [PR in parity-bridges-common](https://github.com/paritytech/parity-bridges-common/pull/2294) where it was originally developed.

**TODO:**
- [x] set/check static "base fee" in `BridgeTable` (according to ExportMessage calculation from BridgeHub)
- [x] set/verify default value for `XcmBridgeHubRouterByteFee` (actaully is `TransactionByteFee`) or setup governance call?
- [x] add possiblity to change `XcmBridgeHubRouterByteFee` via governance + test
- [x] set up benchmarks for router `WeightInfo`
- [x] add support/barrier for unpaid `report_congestion` + test
- [ ] assert `FeeManager` charged fees in test when merged https://github.com/paritytech/polkadot/pull/7005
  - search code for `// TODO:check-parameter: change and assert in tests when (https://github.com/paritytech/polkadot/pull/7005) merged`


paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3366884

parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -471,7 +480,11 @@ type LocalXcmRouter = (
 
 /// The means for routing XCM messages which are not for local execution into the right message
 /// queues.
-pub type XcmRouter = WithUniqueTopic<(LocalXcmRouter, bridging::BridgingXcmRouter)>;
+pub type XcmRouter = WithUniqueTopic<(
+	LocalXcmRouter,
+	// Router which wraps and sends xcm to BridgeHub to be delivered to the Polkadot GlobalConsensus
+	ToPolkadotXcmRouter,
```
svyatonik, line 486: @bkontur Just some random thought. Can you, please, confirm that noone can use the XCM pallet send (or something like that) to send the program with `ExportMessage` instruction from AH to sibling BH? Otherwise he could avoid paying the proper fee for that.

parachains/runtimes/assets/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -586,7 +605,9 @@ pub mod bridging {
 					.add_equals(AssetHubKusama::get().interior.split_global().expect("invalid configuration for AssetHubKusama").1),
 					// and nothing else
 				BridgeHubPolkadot::get(),
-				None
+				// base delivery fee to local `BridgeHub`
+				// (initially was calculated `51220000` + 10% by test `BridgeHubPolkadot::can_calculate_weight_for_paid_export_message_with_reserve_transfer`)
+				Some((XcmBridgeHubRouterFeeAssetId::get(), 56342000).into())
```
svyatonik, line 610: Could probably also be moved to BHP primitives and tested there?

parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs
bkontur: @svyatonik do you mean something like if any other sibling parachain send `xcm::Transact(pallet_xcm::send(to_bridge_hub, ExportMessage(...)))` ?

all AssetHubs have disabled `pallet_xcm::send`:
```
        // We want to disallow users sending (arbitrary) XCMs from this chain.
	type SendXcmOrigin = EnsureXcmOrigin<RuntimeOrigin, ()>;
...
```

but I think I got your good point,
if above stuff will be allowed, we would charge origin by `XcmpQueue` router fee and not by `ToPolkadotXcmRouter` fee, if XcmpQueue fee will be enough and lower than `ToPolkadotXcmRouter`, then yes we could have a problem



I think also there is a simple solution for this:
`pallet_xcm::send_xcm` [adds `DescendOrigin`](https://github.com/paritytech/polkadot/blob/master/xcm/pallet-xcm/src/lib.rs#L1516-L1528), so if we adjust xcm filter on BridgeHubs e.g. do not allow `DescendOrigin` or allow only `MultiLocation {parents: 1, interior: X1(Parachain(para_id))}` pattern it should be enough

svyatonik: I trust your judgement here :) If we do not allow sending arbitrary messages to sibling BH **now**, then probably it is OK - just wanted to be sure. For v2 we won't need it (I assume), because we will be accepting messages from non-system parachains and (I guess) we don't want to force them to use something that is not strictly necessary (like disallowing programs with `DescendOrigin` instruction). Because in any case, we'll either need to support fishermens, or impl something like https://github.com/paritytech/parity-bridges-common/issues/2324 for activating backpressure

parachains/runtimes/assets/asset-hub-polkadot/src/xcm_config.rs
bkontur: fixed done + sanity check

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3398216

svyatonik (approved):

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3400178

bkontur:
bot bench cumulus-assets --subcommand=pallet --pallet=pallet_xcm_bridge_hub_router --runtime=asset-hub-kusama
bot bench cumulus-assets --subcommand=pallet --pallet=pallet_xcm_bridge_hub_router --runtime=asset-hub-polkadot

bkontur:
bot rebase

bkontur:
bot bench cumulus-assets --subcommand=pallet --pallet=pallet_xcm_bridge_hub_router --runtime=asset-hub-kusama
bot bench cumulus-assets --subcommand=pallet --pallet=pallet_xcm_bridge_hub_router --runtime=asset-hub-polkadot

acatangiu (commented):
looks good except couple of comments

parachains/common/src/xcm_config.rs
```diff
@@ -156,6 +161,59 @@ impl<
...
+					// or else we can simply allow all calls
+					let mut decoded_call = DoubleEncoded::<RuntimeCall>::from(encoded_call.clone());
+					ensure!(
+						AllowedCall::contains(
+							decoded_call
+								.ensure_decoded()
+								.map_err(|_| ProcessMessageError::BadFormat)?
+						),
+						ProcessMessageError::BadFormat,
+					);
```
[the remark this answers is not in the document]
svyatonik: Unless I miss something, the `AllowUnpaidTransactsFrom` is configured with `AllowedCall` set to calls of `pallet-xcm-bridge-hub-router`:
```rust
	impl Contains<RuntimeCall> for ToPolkadotXcmRouter {
		fn contains(call: &RuntimeCall) -> bool {
			matches!(
				call,
				RuntimeCall::ToPolkadotXcmRouter(
					pallet_xcm_bridge_hub_router::Call::report_bridge_status { .. }
				)
			)
		}
	}
...
bridging::AllowUnpaidStatusReportsFromSiblingBridgeHub<ToPolkadotXcmRouter>,
...
mod bridging {
...
	/// Barrier for `pallet_xcm_bridge_hub_router::Pallet` to receive congestion statuses from
	/// sibling BridgeHub.
	pub type AllowUnpaidStatusReportsFromSiblingBridgeHub<XcmBridgeHubRouter> =
		AllowUnpaidTransactsFrom<RuntimeCall, XcmBridgeHubRouter, Equals<BridgeHubKusama>>;
...
}
```

So it'll return `contains` only for the call we're interested in. And for better understanding it is better to read `.contains()` here as `.eq()` - because it doesn't iterate through calls in the `decoded_call` - it just compares the `decoded_call` with the `AllowedCall`s. And as I mentioned, our `AllowedCall` is the `report_bridge_status` - so all `utility.batchAll` and so on will be rejected.

So imo all should be fine. But actually this particular struct is maybe an overkill (I was experimenting with filters and as you see it works with some hacks - e.g. decoding) - I think we may allow unpaid calls from BH on AH for v1. But I'm fine with the current version too

parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -202,6 +205,14 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 			}
 		}
 
+		// Allow to change dedicated storage items (called by governance-like)
+		match call {
+			RuntimeCall::System(frame_system::Call::set_storage { items })
+				if items.iter().any(|(k, _)| k.eq(&bridging::XcmBridgeHubRouterByteFee::key())) =>
+				return true,
+			_ => (),
+		};
```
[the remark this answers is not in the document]
svyatonik: Yep, that has caught my attention too. But maybe that is alleviated by the fact that only root (relay chain governance) may call this fn. So it works more like a safety filter, not a strict filter. But yep - we may change the comparison to do a strict cmp with `vec![XcmBridgeHubRouterByteFee::key()]` instead of `.iter().any()`

We may add a dedicated call to the `pallet-xcm-bridge-hub-router` and store it there. But it conflicts a bit with configuration options - like right now we're getting both the `BaseFee` and `ByteFee` from the pallet configuration. So having a dedicated calls for `ByteFee` and not having for the `BaseFee` seems not ok. But we can revert my suggestion (this: https://github.com/paritytech/parity-bridges-common/pull/2313#discussion_r1281417787) and store + manage both `BaseFee` + `ByteFee` by the pallet. It has some drawbacks, but I guess we'll be changing the configuration in the v2 anyway.

Could do both options

acatangiu (approved):

parachains/runtimes/assets/asset-hub-kusama/src/xcm_config.rs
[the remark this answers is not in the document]
acatangiu: Ok, pushed the quick fix: filter exclusively matches `XcmBridgeHubRouterByteFee::key`.

On `Config::(Base/Byte)Fee` vs `<pallet-storage>::(Base/Byte)Fee`, I guess it depends on whether we expect this to change often-ish. If we expect to set it once and have it forever, then config is better. If we expect periodic changes to it, might be better UX to have `setter` call.

PR #2002, opened 2022-12-19
The Ambassador Program

muharem:
The Ambassador Program / Polkadot Collectives Parachain 

The Polkadot Ambassador Program already exist for a while, [here](https://wiki.polkadot.network/docs/ambassadors) you can read more about it.
In the PR, we are bringing the program on chain.

The first version introduces the next functionality:
- managed set of the program members, where every member has a rank (Candidate, Ambassador, Senior Ambassador, Head Ambassador). A member can be promoted or demoted by a plurality voice of higher ranked members given via a referendum. Via substrate pallet_racked_collective.
- referendum functionality for the program members to propose, vote and execute passed proposals on behalf of the members of a certain rank. Via substrate pallet_referenda.
- managed (by members) content (charter, announcements). Via new cumulus pallet pallet_collective_content.

References: 
- The structure of the Ambassador Program - [link](https://www.notion.so/4-Structure-of-the-Ambassador-Program-a10a189cc7f44726b8ae5e584d868efa). note, the Global Head Ambassador role is not introduced in the PR. 

Next steps:
- Mint ambassador nft badge on statemint when promoting 
- Introduce a Pot to store funds managed an ambassador plurality origin

PR #2998, opened 2023-08-11
benchmark script: remove file name for --output argument

muharem:
RRemove the file name from the --output argument for the command that benchmarks the runtimes.

The benchmark CLI tool will automatically resolve the name if it's not provided. In cases where there are multiple instances of one pallet within a single runtime, the file name needs an instance name to avoid a collision. This is already supported by the CLI tool if the file name is omitted in the --output.

The current change to the script does not affect the name resolutions for the existing weights' files.
This change is required for this PR. - https://github.com/paritytech/cumulus/pull/2002

### DIFF
--- a/scripts/benchmarks-ci.sh
+++ b/scripts/benchmarks-ci.sh
@@ -31,11 +31,11 @@ fi
 
 for pallet in ${pallets[@]}
 do
-	output_file="${pallet//::/_}"
+	output_dir=""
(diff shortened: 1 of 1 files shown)


gilescope (approved):

joepetrowski (approved):

muharem:
bot merge

PR #2948, opened 2023-07-28
parachain-system: update RelevantMessagingState according to the unincluded segment

rphmeier:
Closes #2471  ,#2950

PR #3023, opened 2023-08-16
advance parachain properly in XCM-emulator for tests

rphmeier:
This is cherry-picked from #2300 - we needed to treat the parachain as an actual chain for the unincluded segment logic introduced in #2948 .

This PR has test failures which need to be fixed by the system parachains folks

### DIFF
--- a/xcm/xcm-emulator/src/lib.rs
+++ b/xcm/xcm-emulator/src/lib.rs
@@ -1055,7 +1055,6 @@ macro_rules! decl_test_networks {
 				fn hrmp_channel_parachain_inherent_data(
 					para_id: u32,
 					relay_parent_number: u32,
-					parent_head_data: $crate::HeadData,
 				) -> $crate::ParachainInherentData {
 					use $crate::cumulus_primitives_core::{relay_chain::{HeadData, HrmpChannelId}, AbridgedHrmpChannel};
 
@@ -1069,8 +1068,6 @@ macro_rules! decl_test_networks {
 						if let Err(idx) = e_index.binary_search(&recipient_para_id) {
 							e_index.insert(idx, recipient_para_id);
 						}
-
-						sproof.included_para_head = parent_head_data.clone().into();
 
 						sproof
 							.hrmp_channels
--- a/xcm/xcm-emulator/src/lib.rs
+++ b/xcm/xcm-emulator/src/lib.rs
@@ -39,6 +39,7 @@ pub use frame_support::{
 		ProcessMessageError, ServiceQueues,
 	},
 	weights::{Weight, WeightMeter},
+	StorageHasher,
 };
 pub use frame_system::{AccountInfo, Config as SystemConfig, Pallet as SystemPallet};
 pub use pallet_balances::AccountData;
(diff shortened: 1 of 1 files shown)


xcm/xcm-emulator/src/lib.rs
```diff
@@ -622,14 +626,16 @@ macro_rules! decl_test_parachains {
 						let block_number = <Self as Chain>::System::block_number();
 						let mut relay_block_number = <Self as NetworkComponent>::Network::relay_block_number();
 
-						let _ = <Self as Parachain>::ParachainSystem::set_validation_data(
```
rphmeier, line 625: Calling `set_validation_data` here is definitely illegal so had to be removed. Tests possibly depended on it, but it lead to an impossible sequence of calls.

rphmeier:
related to https://github.com/paritytech/cumulus/issues/2703

xcm/xcm-emulator/src/lib.rs
rphmeier: elaborating a bit more. `init()` is called within `execute_block` the first time it's run.

`set_validation_data` needs to be followed by a `System::finalize` call.
It might be more correct as 
1. initialize
2. set_validation_data
3. finalize
4. initialize

but that is basically just running a full block, which shouldn't be necessary.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3403396

rphmeier:
btw, one of the reasons these tests are so flaky is because the `execute_block` function for every parachain invokes `Network::process_messages()` which side-channel dumps messages into every other parachain as well as itself.

The side-effecting nature of the code here would likely make it near-impossible to write tests which didn't depend on highly specific implementation details and control flow of the test runner, which is why they fail every time we so much as look at this code.

Strong recommend a refactoring here. Executing a block on a parachain should cause messages to be processed only to itself. This may require tests to be rewritten to deliberately execute empty blocks on a receiving parachain in order to observe the events - this is OK and what happens in the real world.

xcm/xcm-emulator/src/lib.rs
```diff
@@ -816,16 +833,26 @@ macro_rules! __impl_test_ext_for_parachain {
...
+							.insert(para_id.into(), $crate::HeadData(created_header.encode()))
+						);
+
 						// clean events
 						<Self as Chain>::System::reset_events();
+
+						// reinitialize before next call.
+						let next_block_number = block_number + 1;
+						<Self as Chain>::System::initialize(&next_block_number, &created_header.hash(), &Default::default());
+						<Self as Parachain>::ParachainSystem::on_initialize(next_block_number);
```
ggwpez, line 855: It would be nice if we can just do `AllPalletsWithSystem::on_initialize` here, but it seems to not be supported currently.  
Some pallets (like the MessageQueue) only advance on initialize, and injecting them manually into the config is also ugly.

ggwpez: Found a better solution without on_initialize needs.

PR #3026, opened 2023-08-17
Revert #2958

gilescope:
#2958 is causing some trouble with a few other PRs. There's some suggestion that we should revert it. This PR does this. We will re-add it as a series of PRs that should be less disruptive to other devs.


bkontur:
> causing some trouble with a few other PRs

Just out of curiosit, what troubles? Could you please link those PRs here?

gilescope:
Sure, https://github.com/paritytech/cumulus/pull/3023 (holding up async backing being merged) and https://github.com/paritytech/cumulus/pull/2157 - unifying the xcm message queues. Both are things we want to land before monorepo.

bkontur:
#2157 ran one week ago with no failure: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3355109
#2157 should be rebased on actual master and we will see if there is any failure,
or am I missing something here?

#3023 ok, has some errors: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3403396, but why revert instead of fixing for async backing?

gilescope:
Rebasing/merge master of #2157 does lead to a lot of conflicts (understandably).

gilescope:
But I think that you're right - https://github.com/paritytech/cumulus/pull/3023 is going to still be a problem even if we revert this changeset.

rphmeier:
> instead of fixing for async backing?

Please do it if possible. I left a comment in #3023 about why this test harness is flaky & unrealistic, so it needs refactoring regardless. Unfortunately, I had to fix some of that unrealistic behavior in order to ensure the invariants of the `ParachainSystem` pallet in async-backing were upheld, and this caused downstream test failures.

gilescope:
Given that people have found workarounds for now we will keep this in, and improve on what's there in upcoming PRs.

PR #3037, opened 2023-08-21
Flexibility in xcm-emulator methods

girazoki:
This PR decouples the `accountId` type from the `accountId32` type and rather makes emulator methods depend on the `accountId` type defined by the chain runtime in question. This is important if we want the emulator to stay compatible with chains that have 20 byte accounts like Moonbeam.

The only method I did not change is `account_id_of`, because it is tied to `sr25519` account generation. In any case this method is not that crucial because 20 byte account senders and receivers can be declared with the `parameter_types!` macro.

Kusama address: HZetxGJ1k9kJdNwHFHUn3i2Xk11jqiX6kRG89BJZXcNAm8z

### DIFF
--- a/xcm/xcm-emulator/src/lib.rs
+++ b/xcm/xcm-emulator/src/lib.rs
@@ -372,7 +372,7 @@ macro_rules! decl_test_relay_chains {
 				type RuntimeEvent = $runtime::RuntimeEvent;
 				type System = $crate::SystemPallet::<Self::Runtime>;
 
-				fn account_data_of(account: AccountId) -> $crate::AccountData<Balance> {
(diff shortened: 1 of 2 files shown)


crystalin (approved):

librelois (approved):

Agusrodri (approved):

xcm/xcm-emulator/src/lib.rs
```diff
@@ -76,6 +76,8 @@ pub use xcm::{
 };
 pub use xcm_executor::traits::ConvertLocation;
 
+pub type RuntimeAccountId<T> = <T as frame_system::Config>::AccountId;
```
gilescope, line 79: Just wondering if we could call it `AccountIdOf`?

PR #3045, opened 2023-08-22
Refactor docker image build

EgorPopelyaev:
This PR refactors github actions which build the docker image for the `polkadot-parachain`binary. It touches both flows automated and manual.
From now on it is going to be only one workflow which triggers automatically, when a new release is published or can be triggered manually on demand to publish an RC or Release image


alvicsam (approved):

chevdor (changes requested):
Some small changes but that loooks good already.
I would be nice to keep only ONE single workflow, that can handle `workflow_dispatch` if the needs occurs. That would also be easier to maintain that 2 worfklows we need to keep in sync.

paritytech-cicd-pr:
**Logs**: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/3426014

.github/workflows/release-50_publish-docker-manual.yml
```diff
@@ -0,0 +1,179 @@
...
+            curl -s \
+              -H "Authorization: Bearer ${GITHUB_TOKEN}" https://api.github.com/repos/$OWNER/$REPO/releases | \
+              jq '.[] | { name: .name, id: .id }'
+        required: true
+        type: string
+      production_image:
+        description: "Set to true for an image for the final release or false for an rc image"
+        required: true
+        type: string
+        default: "false"
```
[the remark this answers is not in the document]
EgorPopelyaev: Addressed :)

.github/workflows/release-50_publish-docker-manual.yml
```diff
@@ -142,7 +145,7 @@ jobs:
                 tag=$(git name-rev --tags --name-only $(git rev-parse HEAD)) && \
                 [ "${tag}" != "undefined" ] && echo "tag=${tag}" >> $GITHUB_OUTPUT || \
                 echo "No tag, doing without"
-            else
+            elif [ "${type}" == "final" ]; then
```
chevdor, line 148: if...elif without else could be risky.
If we ever add a new choice or worst, rename a choice, it would silently ignore it.
It would be safer adding an `else` branch that raises an error. That would force us to cover all cases properly.

chevdor: A swicth/case may be more appropriate here.

chevdor (changes requested):

.github/workflows/release-50_publish-docker.yml
```diff
@@ -0,0 +1,204 @@
...
+          mkdir -p release-artifacts && cd release-artifacts
+
+            for f in $BINARY $BINARY.asc $BINARY.sha256; do
+                URL="https://github.com/${{ github.event.repository.full_name }}/releases/download/${{ github.event.release.tag_name }}/$f"
+                echo " - Fetching $f from $URL"
+                wget "$URL" -O "$f"
+            done
+            chmod a+x $BINARY
+            cp -f ${TMP}/lib.sh .
+            ls -al
```
chevdor, line 92: Some indentation does not look right here :)

chevdor (approved):
LGTM beside the indentation remark

.github/workflows/release-50_publish-docker.yml
EgorPopelyaev: good catch, it's addressed :)

EgorPopelyaev:
bot merge

